PR #7, opened 2024-04-12
5 Add reference and note

ibevers:


ibevers:
Redundant since new commits automatically added to https://github.com/sensein/pipepal/pull/6

PR #48, opened 2024-06-03
Adding utility functions

### LINKED ISSUE #24: Task: Preprocessing
### Description

Preprocess audio data by resampling, transforming stereo to mono (via downmix or channel selection), chunking based on start and end timestamps, extracting audio from video, and performing audio data augmentation.

### Tasks

- [x] resampling, 
- [x] transforming stereo to mono (via downmix or channel selection),
- [x] chunking based on start and end timestamps, 
- [ ] extracting audio from video, 
- [x] #43

### Freeform Notes

_No response_

### LINKED ISSUE #30: Task: Utilities
### Description

Implement utility functions for computing Word Error Rate (WER), cosine similarity between embeddings, Equal Error Rate (EER), Centered Kernel Alignment (CKA), Canonical Correlation Analysis (CCA), dimensionality reduction (PCA, t-SNE, UMAP), and cross-correlation between two signals.

### Tasks

- [x] Computing Word Error Rate (WER), 
- [x] Computing cosine similarity between embeddings, 
- [x] Computing Equal Error Rate (EER), 
- [x] Computing Centered Kernel Alignment (CKA), 
- [x] Computing Canonical Correlation Analysis (CCA), 
- [x] Computing cross-correlation between two signals.

### Freeform Notes

_No response_

fabiocat93:
I have added some utility functions for computing cosine similarity, eer, wer, cca, cka, cross correlation

### DIFF
--- a/src/senselab/audio/tasks/preprocessing.py
+++ b/src/senselab/audio/tasks/preprocessing.py
@@ -46,12 +46,14 @@ def downmix_audios_to_mono(audios: List[Audio]) -> List[Audio]:
     """
     down_mixed_audios = []
     for audio in audios:
-        if audio.waveform.dim() != 2 or audio.waveform.size(0) < 1:
-            raise ValueError("waveform should have shape (num_channels, num_samples)")
+        down_mixed_audios.append(
+            Audio(
+                waveform=audio.waveform.mean(dim=0, keepdim=True),
+                sampling_rate=audio.sampling_rate,
+                metadata=audio.metadata.copy(),
+            )
+        )
 
-        down_mixed_audio = audio.copy()
-        down_mixed_audio.waveform = audio.waveform.mean(dim=0)
-        down_mixed_audios.append(down_mixed_audio)
     return down_mixed_audios
 
 
@@ -64,23 +66,19 @@ def select_channel_from_audios(audios: List[Audio], channel_index: int) -> List[
         channel_index (int): The index of the channel to select.
 
     Returns:
-        List[Audio]: The list of audio objects with the selected channel. Shape: (num_samples).
+        List[Audio]: The list of audio objects with the selected channel. Shape: (num_channels, num_samples).
     """
     mono_channel_audios = []
     for audio in audios:
-        print("audio.waveform.shape")
-        print(audio.waveform.shape)
-        if audio.waveform.dim() != 2:
-            raise ValueError("waveform should have shape (num_channels, num_samples)")
         if audio.waveform.size(0) <= channel_index:
             raise ValueError("channel_index should be valid")
-
-        mono_channel_audio = audio.copy()
-        mono_waveform = audio.waveform[channel_index, :]
-        if mono_waveform.dim() != 2:
-            mono_waveform = mono_waveform.unsqueeze(0)
-        mono_channel_audio.waveform = mono_waveform
-        mono_channel_audios.append(mono_channel_audio)
+        mono_channel_audios.append(
+            Audio(
+                waveform=audio.waveform[channel_index, :],
+                sampling_rate=audio.sampling_rate,
+                metadata=audio.metadata.copy(),
+            )
+        )
     return mono_channel_audios
 
 
--- a/src/senselab/audio/tasks/preprocessing.py
+++ b/src/senselab/audio/tasks/preprocessing.py
@@ -1,19 +1,16 @@
 """This module implements some utilities for the preprocessing task."""
 
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Tuple
 
 import torch
 import torchaudio.functional as F
 from datasets import Dataset
 
 from senselab.utils.data_structures.audio import Audio
-from senselab.utils.tasks.input_output import (
-    _from_dict_to_hf_dataset,
-    _from_hf_dataset_to_dict,
-)
+from senselab.utils.tasks.input_output import _from_dict_to_hf_dataset, _from_hf_dataset_to_dict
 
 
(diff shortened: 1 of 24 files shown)


wilke0818 (approved):
Looks mostly good to me. I have two suggestions that might be helpful in the future that can be minor tasks for others later:
- In speech_to_text_evaluation I think it would be good to give more concrete definitions on what some of these mean or reference say the jiwer docs since I haven't heard of/am not sure the definitions of say WIP or WIL.
- In cca_cka we should create an enum for the types of kernels that are allowed as a better software practice

fabiocat93:
thank you, @wilke0818, for your feedback. For now, I have created an enum for the kernel types. I will address documentation issues more calmly if that sounds reasonable. 
Also, I have added some more functionalities for audio pre-processing. Please feel free to review the code whenever you have time

src/senselab/audio/tasks/preprocessing.py
```diff
@@ -30,13 +27,93 @@ def resample_audio_dataset(audios: List[Audio], resample_rate: int, rolloff: flo
...
+    Args:
+        audios (List[Audio]): A list of Audio objects with a tensor representing the audio waveform.
+                                 Shape: (num_channels, num_samples).
+
+    Returns:
+        List[Audio]: The list of audio objects with a mono waveform averaged from all channels. Shape: (num_samples).
+    """
+    down_mixed_audios = []
+    for audio in audios:
+        if audio.waveform.dim() != 2 or audio.waveform.size(0) < 1:
```
wilke0818: Not sure if this is needed since Pydantic should be handling this?

src/senselab/audio/tasks/preprocessing.py
```diff
@@ -30,13 +27,93 @@ def resample_audio_dataset(audios: List[Audio], resample_rate: int, rolloff: flo
...
+            raise ValueError("waveform should have shape (num_channels, num_samples)")
+
+        down_mixed_audio = audio.copy()
```
wilke0818: think we need to change this to `audio.model_copy(deep=True)` but curious if what's better is this or just creating a new Audio instance?

src/senselab/audio/tasks/preprocessing.py
```diff
@@ -30,13 +27,93 @@ def resample_audio_dataset(audios: List[Audio], resample_rate: int, rolloff: flo
...
+        down_mixed_audio.waveform = audio.waveform.mean(dim=0)
```
wilke0818: I worry slightly about this but maybe your test cases handle this, but does this guarantee that waveform is shape (1, num_samples)? I'm not 100% sure that the field_validator runs in this case but if it does then this is fine. If it doesn't, then use `audio.waveform.mean(dim=0, keepdim=True)`

src/senselab/audio/tasks/preprocessing.py
```diff
@@ -30,13 +27,93 @@ def resample_audio_dataset(audios: List[Audio], resample_rate: int, rolloff: flo
...
+
+def downmix_audios_to_mono(audios: List[Audio]) -> List[Audio]:
+    """Downmixes a list of Audio objects to mono by averaging all channels.
+
+    Args:
+        audios (List[Audio]): A list of Audio objects with a tensor representing the audio waveform.
+                                 Shape: (num_channels, num_samples).
+
+    Returns:
+        List[Audio]: The list of audio objects with a mono waveform averaged from all channels. Shape: (num_samples).
```
wilke0818, line 45: I guess this gets into a later comment I made, but at least in the Audio class as of right now, we standardize the waveform field to be (num_channels, num_samples) and perhaps to keep consistency we keep it as (1, num_samples)?

src/senselab/audio/tasks/preprocessing.py
```diff
@@ -30,13 +27,93 @@ def resample_audio_dataset(audios: List[Audio], resample_rate: int, rolloff: flo
...
+        if audio.waveform.dim() != 2 or audio.waveform.size(0) < 1:
+            raise ValueError("waveform should have shape (num_channels, num_samples)")
+
+        down_mixed_audio = audio.copy()
+        down_mixed_audio.waveform = audio.waveform.mean(dim=0)
+        down_mixed_audios.append(down_mixed_audio)
+    return down_mixed_audios
+
+
+def select_channel_from_audios(audios: List[Audio], channel_index: int) -> List[Audio]:
```
wilke0818, line 60: basically see all of my comments for the above function

src/senselab/utils/tasks/eer.py
wilke0818, line 1: we've broken up a lot of the tasks into individual files (e.g. individual modules) and I'm wondering if we should consolidate a bit to make it easier to use. Not super opinionated on this though.

src/senselab/audio/tasks/preprocessing.py
fabiocat93: fair. i have removed this

src/senselab/audio/tasks/preprocessing.py
fabiocat93: i am not sure there is a big difference here. But can easily switch to creating a new Audio object so that we are consistent in coding style plus it becomes clearer how we create a new entity/item

src/senselab/audio/tasks/preprocessing.py
fabiocat93: it was managed anyway, but i have now added `keepdim=True`  to make it clearer

src/senselab/utils/tasks/eer.py
fabiocat93: this is a good point. I think we can proceed like this for this first release and next week we do some restructure

PR #50, opened 2024-06-04
adding some more API classes (participant, session, dataset)

### LINKED ISSUE #42: Task: Internal architecture (data structures)
### Description

Developing custom Audio, Dataset (this one will have the method `.from_bids`), Participant, Session, Metadata, HuggingFaceModel classes for the package so we don't rely directly on HF Datasets. 

1. In the future, we will also need Image and Video classes.

### Tasks

- [x] Audio
- [x] Dataset
- [x] Participant
- [x] Session
- [x] HuggingFaceModel

### Freeform Notes

_No response_

fabiocat93:


scripts/experiment7.py
wilke0818, line 1: this pretty easily could've been a test file

src/senselab/utils/data_structures/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+
+
+class Participant(BaseModel):
+    """Data structure for a participant in a dataset."""
+
+    id: str = Field(default_factory=lambda: str(uuid.uuid4()))
+    metadata: Dict = Field(default={})
+
+    @field_validator("id", mode="before")
+    def set_id(cls, v: str) -> str:
```
wilke0818, line 16: I don't know if this will ever get run because you have a default_factory if the id is not provided so probably not needed.

src/senselab/utils/data_structures/dataset.py
```diff
@@ -0,0 +1,90 @@
...
+        return v or str(uuid.uuid4())
+
+    def __eq__(self, other: object) -> bool:
+        """Overloads the default BaseModel equality to correctly check that ids are equivalent."""
+        if isinstance(other, Session):
+            return self.id == other.id
+        return False
+
+
+class SenselabDataset(BaseModel):
```
wilke0818, line 45: seems like we made complimentary datasets (mine handles Audio/Video, yours has some base functions for the participants and sessions).

wilke0818 (commented):
Let's connect on making sure we can merge our Datasets together correctly but seems good generally. Do you know that the HuggingFace functionality works? Do we have tests for it?

scripts/experiment7.py
wilke0818: Pretty sure that this in fact did become a test file but probably should limit use of these experiment files to keep the repo clean

fabiocat93: agree. we can remove that `scripts` folder

src/senselab/utils/data_structures/dataset.py
fabiocat93: But the default_factory is if you use it in a Dataset. Otherwise, you need some checks (and this is why we have the validator)

src/senselab/utils/data_structures/dataset.py
fabiocat93: sounds amazing. If you want, we can make a coding session together tomorrow after lunch and marge all what we have

PR #52, opened 2024-06-07
release 0.2.0 including the new custom interface, pre-processing functions, utility functions

fabiocat93:


satra:
@fabiocat93 - if you want to trigger releases using auto, i think it the `auto` workflow needs a `major`, `minor`, or `patch` label in addition to `release`. it may be that this will require another PR merge to actually make a release.

PR #60, opened 2024-06-11
Adding voice cloning API

fabiocat93:


src/senselab/audio/tasks/voice_cloning/api.py
```diff
@@ -0,0 +1,43 @@
...
+
+from typing import Any, Dict, List, Optional
+
+import pydra
+
+from senselab.audio.data_structures.audio import Audio
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import SenselabModel, TorchModel
+
+from .knnvc import KNNVC
```
wilke0818: not sure if it would be better to use the full import path like above

src/senselab/utils/data_structures/model.py
```diff
@@ -35,10 +37,71 @@ def validate_path_or_uri(cls, value: Union[str, Path]) -> Union[str, Path]:
 
         return value
 
+
+class HFModel(SenselabModel):
```
wilke0818, line 41: I am not sure I see the necessity for these model classes

src/tests/audio/tasks/voice_cloning_test.py
wilke0818, line 1: do you have any tests where voice cloning works?

wilke0818 (approved):

src/senselab/utils/data_structures/model.py
fabiocat93: this way, we do check if the model actually exists on the hub

src/senselab/audio/tasks/voice_cloning/api.py
fabiocat93: fixed

src/tests/audio/tasks/voice_cloning_test.py
fabiocat93: done

PR #62, opened 2024-06-13
ref: adjust poetry to install opensmile dependency 

satra:
poetry will install iso-639 before installing packages. on macos m1 this appears to resolve the dependency issue.


fabiocat93 (approved):
in person

PR #67, opened 2024-06-17
Add diarization tests

### LINKED ISSUE #72: Task: Add diarization tests
### Description

Add at least three tests to speaker diarization.

### Tasks

- [ ] test _annotation_to_dict empty input 
- [ ] test _annotation_to_dict single input 
- [ ] test _annotation_to_dict longer input 
- [ ] validate correct tensor transformation and shape adjustments _pyannote_diarize_batch
- [ ] correct output with each model type
- [ ] check diarizations same with batch true and batch false

### Freeform Notes

_No response_

ibevers:


ibevers:
@fabiocat93 Fyi, it seems that numpy >=2.0.0 is a dependency of one of the more recent packages that was added. pyannote-audio  uses `np.NaN`  which was deprecated in numpy 2.0.0. I submitted a PR to pyannote-audio replacing this with `np.nan` (the numpy >=2.0.0 version) and rolled back the numpy version. 

I then ran `poetry install --with dev,docs` again to see if there are any dependency conflicts. I didn't get any issues, so hopefully this means that the roll-back won't affect any other tasks.

PR #78, opened 2024-06-24
adding speech enhancement task

fabiocat93:


src/senselab/audio/tasks/speaker_embeddings/api.py
```diff
@@ -37,7 +37,7 @@ def extract_speaker_embeddings_from_audios(
         >>> print(embeddings[0].shape)
         torch.Size([192])
     """
-    if isinstance(model, HFModel):  # TODO: check that this is a speechbrain model!!!
```
wilke0818, line 40: seems you were too excited

src/senselab/audio/tasks/speech_enhancement/api.py
```diff
@@ -0,0 +1,36 @@
...
+
+from senselab.audio.data_structures.audio import Audio
+from senselab.audio.tasks.speech_enhancement.speechbrain import SpeechBrainEnhancer
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import HFModel, SenselabModel
+
+
+def enhance_audios(
+    audios: List[Audio], 
+    model: SenselabModel = HFModel(path_or_uri="speechbrain/sepformer-wham16k-enhancement", revision="main"),
```
wilke0818, line 14: should this be SenselabModel or just HFModel? Do we think there'll be a point where other models will be able to work?

src/senselab/audio/tasks/speech_enhancement/speechbrain.py
```diff
@@ -0,0 +1,92 @@
...
+                source=model.path_or_uri, 
+                run_opts={"device": device.value}
+            )
+        return cls._models[key]
+
+    @classmethod
+    def enhance_audios_with_speechbrain(
+        cls,
+        audios: List[Audio],
+        model: HFModel = HFModel(path_or_uri="speechbrain/sepformer-wham16k-enhancement", revision="main"),
```
wilke0818, line 48: I guess similar to above. We should just be consistent on what we use.

src/senselab/audio/tasks/speech_enhancement/speechbrain.py
```diff
@@ -0,0 +1,92 @@
...
+                    f"Audio waveform must be mono (1 channel), but got {audio.waveform.shape[0]} channels"
+                )
+            if audio.sampling_rate != expected_sample_rate:
+                raise ValueError(
+                    "Audio sampling rate " + str(audio.sampling_rate) + 
+                    " does not match expected " + str(expected_sample_rate)
+                )
+
+        # Stack audio waveforms for batch processing
+        waveforms = torch.stack([audio.waveform.squeeze() for audio in audios])
```
wilke0818, line 81: I've been thinking whether our Audio implementation makes sense when many functions require us to drop it down to a single channel and so we're forced to always squeeze them together. Also, should they all be stacked together, or batched at a certain size? Also with device, do you need to place them on the device or is it assumed they are already on there?

src/tests/audio/tasks/speech_enhancement_test.py
```diff
@@ -0,0 +1,74 @@
...
+from senselab.audio.data_structures.audio import Audio
+from senselab.audio.tasks.preprocessing.preprocessing import resample_audios
+from senselab.audio.tasks.speech_enhancement.api import enhance_audios
+from senselab.audio.tasks.speech_enhancement.speechbrain import SpeechBrainEnhancer
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import HFModel
+
+
+@pytest.fixture
+def sample_audio() -> Audio:
```
wilke0818: might be better to make a utility function for the tests that does this so we don't have to keep repeating it

src/tests/audio/tasks/speech_enhancement_test.py
```diff
@@ -0,0 +1,74 @@
...
+    return HFModel(path_or_uri="speechbrain/sepformer-wham16k-enhancement")
+
+@pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") == "true", reason="Don't want to hit the GitHub API limit")
+def test_enhance_audios_stereo_audio(stereo_audio: Audio, hf_model: HFModel) -> None:
+    """Test that enhancing stereo audios raises a ValueError."""
+    with pytest.raises(ValueError, match="Audio waveform must be mono"):
+        SpeechBrainEnhancer.enhance_audios_with_speechbrain(audios=[stereo_audio], model=hf_model)
+
+@pytest.mark.skipif(os.getenv("GITHUB_ACTIONS") == "true", reason="Don't want to hit the GitHub API limit")
+def test_enhance_audios(sample_audio: Audio, hf_model: HFModel) -> None:
```
wilke0818: So this is also something I have done within my own code and I question whether these tests are good practices. Should we have any expectations on what the results of enhanced_audios should be? Right now we're testing it at a very basic level that doesn't really guarantee functionality but instead guarantees runability and compilation.

wilke0818 (approved):
Left some comments. Most are more general thoughts than changes that you need to make so approving it.

src/senselab/audio/tasks/speaker_embeddings/api.py
fabiocat93: haha do you have any idea on how we can check if a local model is a `speechbrain` model?

src/senselab/audio/tasks/speech_enhancement/api.py
fabiocat93: we want this to be a senselab model, but for now the default is the child class HFModel. Hopefully, in the close future we will add more tools beyond the hf ones

src/senselab/audio/tasks/speech_enhancement/speechbrain.py
fabiocat93: since speechbrain models are HF models, there is no need to initialize model as SenselabModel. We want to initialize it as HFModel to better validate the input indeed

src/senselab/audio/tasks/speech_enhancement/speechbrain.py
fabiocat93: good points. 
- i think we are good with our Audio class, since it enables working with stereo audio. The fact that many models only work with mono audio is not under our control. 
- If we want to optimize it, we should pass a batch
- moving to the specified device is done under the hood.

src/tests/audio/tasks/speech_enhancement_test.py
fabiocat93: Done

src/tests/audio/tasks/speech_enhancement_test.py
fabiocat93: You are so right. For now, we are just testing the basics. Later on we will work on making senselab robuster.

src/senselab/audio/tasks/speaker_embeddings/api.py
wilke0818: I think we can trust the SpeechBrain functionality to check that for us. This is the function that ends of getting called when you give the source which handles if it's HuggingFace or a local file
https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/utils/fetching.py
As well as this which seems to imply that they'll handle non-speech brain files
https://github.com/speechbrain/speechbrain/blob/develop/speechbrain/inference/classifiers.py#L30

fabiocat93: will think about that, thanks!

PR #80, opened 2024-06-24
adding the text to speech task

fabiocat93:


src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -0,0 +1,38 @@
...
+from senselab.audio.data_structures.audio import Audio
+from senselab.audio.tasks.text_to_speech.huggingface import HuggingFaceTTS
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.language import Language
+from senselab.utils.data_structures.model import HFModel, SenselabModel
+
+
+def synthesize_texts(
+    texts: List[str], 
+    model: SenselabModel = HFModel(path_or_uri="suno/bark", revision="main"),
```
wilke0818, line 13: not too important but why this model? I know StyleTTS2.0 has been mentioned before though I don't know how their performances compare

src/senselab/audio/tasks/text_to_speech/huggingface.py
```diff
@@ -0,0 +1,80 @@
...
+            https://huggingface.co/docs/transformers/tasks/text-to-speech
+        """
+        pipe = HuggingFaceTTS._get_hf_tts_pipeline(model=model, device=device)
+
+        synthesis_kwargs: Dict[str, Any] = {}
+        synthesized_audios = pipe(texts, **synthesis_kwargs)
+        
+        audios = []
+        for synth in synthesized_audios:
+            audios.append(Audio(waveform=synth["audio"],
```
wilke0818, line 77: do you think it would make sense to create some piece of metadata along with it? So far, that's been pretty under-utilized.

src/senselab/audio/tasks/text_to_speech/huggingface.py
```diff
@@ -0,0 +1,80 @@
...
+                "text-to-speech",
+                model=model.path_or_uri,
+                revision=model.revision,
+                device=device.value,
+                torch_dtype=torch_dtype
+            )
+        return cls._pipelines[key]
+
+    @classmethod
+    def synthesize_texts_with_transformers(
```
wilke0818, line 47: Under the assumption you want people interacting with these through the API, then we should probably have this as _synthesize...?

wilke0818 (approved):

src/senselab/audio/tasks/text_to_speech/api.py
fabiocat93: Consider it as a placeholder for now. We will need to add more models and experiment with them. I like MARS-TTS the most, but I am waiting for them to confirm that I can integrate their technology under an MIT/Apache license. StyleTTS is of course another good option

src/senselab/audio/tasks/text_to_speech/huggingface.py
fabiocat93: My perception is that metadata is more of a field that we have prepared for the user. How do you see it used here?

src/senselab/audio/tasks/text_to_speech/huggingface.py
fabiocat93: good point

src/senselab/audio/tasks/text_to_speech/huggingface.py
wilke0818: There are a few ways you could probably use it, including putting the text transcript in or including some details of the model/process that was run. Can also be user-defined and leave it for them to input since they'll have all of those details theoretically.

PR #81, opened 2024-06-25
Add dimensionality reduction task

### LINKED ISSUE #44: Task: Dimensionality reduction
### Description

Implementing an interface for dimensionality reduction

### Tasks

- [x] PCA
- [x] tSNA
- [x] UMAP

### Freeform Notes

_No response_

900miles:
Implements the dimensionality reduction task for umap, tsne, and pca.

### DIFF
--- a/src/senselab/utils/tasks/dimensionality_reduction.py
+++ b/src/senselab/utils/tasks/dimensionality_reduction.py
@@ -1,0 +1,71 @@
+"""This module provides the implementation of dimensionality reduction."""
+
+import pydra
+import torch
+import umap
+from sklearn.decomposition import PCA as sklearn_PCA
+from sklearn.manifold import TSNE as sklearn_TSNE
+
+
+def compute_dimensionality_reduction(
+    data: torch.Tensor, model: str = "pca", n_components: int = 2, **kwargs: object
+) -> torch.Tensor:
+    """Reduce the dimensionality of the given data using the specified model.
+
+    This function provides three methods: PCA, t-SNE, or UMAP. The reduced data is returned as a PyTorch tensor.
+
+    Args:
+        data (torch.Tensor): Input data tensor of shape (n_samples, n_features).
+        model (str, optional): The dimensionality reduction model to use.
+            Choices are:
+                * "pca" for Principal Component Analysis,
+                * "tsne" for t-Distributed Stochastic Neighbor Embedding
+                * "umap" for Uniform Manifold Approximation and Projection
+        n_components (int, optional): Number of dimensions in the output.
+            Must be less than or equal to the number of features in the input data.
+            Defaults to 2.
+        **kwargs: Additional keyword arguments to pass to the chosen model.
+
+    Returns:
+        torch.Tensor: The reduced data tensor of shape (n_samples, n_components).
+
+    Raises:
+        ValueError: If an invalid model choice is provided or if n_components is
+            greater than the number of features in the input data or less than 0.
+
+    Examples:
+        >>> data = torch.randn(100, 10)  # 100 samples, 10 features
+        >>> reduced_data = compute_dimensionality_reduction(data, model="pca", n_components=2)
+        >>> print(reduced_data.shape)
+        torch.Size([100, 2])
+    """
+    if n_components > data.shape[1]:
+        raise ValueError("n_components must be less than or equal to the number of features in the input data")
+
+    if n_components < 1:
+        raise ValueError("n_components must be greater than 0")
+
+    if data.numel() == 0:
+        raise ValueError("Input data must not be empty")
+
+    if model == "pca":
+        # Perform PCA
+        reducer = sklearn_PCA(n_components=n_components, **kwargs)
+        reduced_data = reducer.fit_transform(data)
+
+    elif model == "tsne":
+        # Perform t-SNE
+        reducer = sklearn_TSNE(n_components=n_components, **kwargs)
+        reduced_data = reducer.fit_transform(data)
+
+    elif model == "umap":
+        reducer = umap.UMAP(n_components=n_components, **kwargs)
+        reduced_data = reducer.fit_transform(data)
+
+    else:
+        raise ValueError(f"Invalid model choice: {model}")
+
+    return torch.from_numpy(reduced_data).float()
+
+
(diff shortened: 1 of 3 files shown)


src/senselab/utils/tasks/dimensionality_reduction.py
```diff
@@ -0,0 +1,71 @@
...
+
+import pydra
+import torch
+import umap
+from sklearn.decomposition import PCA as sklearn_PCA
+from sklearn.manifold import TSNE as sklearn_TSNE
+
+
+def compute_dimensionality_reduction(
+    data: torch.Tensor, model: str = "pca", n_components: int = 2, **kwargs: object
```
fabiocat93: @900miles can you make the model more robust as an Enum (see https://github.com/sensein/senselab/blob/main/src/senselab/utils/data_structures/device.py)

fabiocat93, line 11: how if I want to change some parameters (e.g., perplexity, random_state, max_iter, learning_rate, initial_state, ...)?

fabiocat93: alternatively, you can make them classes and include the parameters they want

src/senselab/utils/tasks/dimensionality_reduction.py
```diff
@@ -0,0 +1,71 @@
...
+) -> torch.Tensor:
+    """Reduce the dimensionality of the given data using the specified model.
+
+    This function provides three methods: PCA, t-SNE, or UMAP. The reduced data is returned as a PyTorch tensor.
+
+    Args:
+        data (torch.Tensor): Input data tensor of shape (n_samples, n_features).
+        model (str, optional): The dimensionality reduction model to use.
+            Choices are:
+                * "pca" for Principal Component Analysis,
```
fabiocat93: can you add a reference to the documentation of these tools?

src/senselab/utils/tasks/dimensionality_reduction.py
```diff
@@ -0,0 +1,71 @@
...
+    Returns:
+        torch.Tensor: The reduced data tensor of shape (n_samples, n_components).
+
+    Raises:
+        ValueError: If an invalid model choice is provided or if n_components is
+            greater than the number of features in the input data or less than 0.
+
+    Examples:
+        >>> data = torch.randn(100, 10)  # 100 samples, 10 features
+        >>> reduced_data = compute_dimensionality_reduction(data, model="pca", n_components=2)
```
fabiocat93: Can you add some more examples including the important arguments for each model?

fabiocat93 (changes requested):
thank you @900miles !i have left some comments

src/senselab/utils/tasks/dimensionality_reduction.py
fabiocat93: the point of the entire senselab tool is that we want to give people an easier access to methods. ideally, we want them to find everything here, without the need of jumping from one documentation to another

900miles: I'm not exactly sure what you mean here. The scipy standard seems to be to use typing.Literal for choosing models (see `svd_solver` in https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html). Is there a reason this should use an enum or classes instead? Also because we directly use scipy/umap to implement this, wouldn't it be better to link directly to the documentation of those methods, so we don't run into problems with outdated or incomplete documentation?

fabiocat93: The main idea behind proposing an Enum or separate classes for each model type (PCA, t-SNE, UMAP) is to encapsulate the specific configurations and parameters each model might need, in a more structured and self-documenting way. This approach can help ensure that users of the `senselab` package do not need to remember specific parameter names or configurations for each model, which can improve usability and reduce errors. But I like the idea of pointing to their external documentation for reference

900miles: Ok, makes sense. But that would mean we would need to update the classes every time a parameter is deprecated or a new one is added in the underlying models, right?

900miles: Actually maybe I still don't fully understand. I pushed some commits linking to documentation, adding examples, and making the model type a string Literal instead of a generic string. Does that cover what you're thinking of?

fabiocat93: Thank you @900miles . Given that the release is today, i will accept the PR. Still, I would prefer to see classes and good docstrings with all params for each tool (PCA, tSNE, UMAP), especially since I don't expect the math behind these will change in the next short future. This would make `senselab` a much better interface for the user.

fabiocat93 (approved):

src/senselab/utils/tasks/dimensionality_reduction.py
900miles: Ok, do you mean something like this?

```
class DimensionalityReduction():

    ...
    
    def pca(n_components=None, *, copy=True, whiten=False, svd_solver='auto', tol=0.0, iterated_power='auto', ...):
        """
        n_components: <int, float or ‘mle’, default=None>
            Number of components to keep
        copy: <bool, default=True>
            If False, data passed to fit are overwritten and running fit(X).transform(X) will not yield the expected results, use fit_transform(X) instead.
        ...
        
        See https://scikit-learn.org/stable/modules/generated/sklearn.decomposition.PCA.html for more details.
        """
        ...
        
    def tsne(n_components=2, *, perplexity=30.0, early_exaggeration=12.0, learning_rate='auto', max_iter=None, ...):
        ...
    
    def umap(...):
        ...
```

I worry that if we explicitly write out the parameters like this in our code, we would have to update it every time the underlying implementation changes in scipy or umap. For example, the most recent scipy (1.5.0) changed the names of one of the parameters in tsne and deprecated another, and added a new option to the `svd_solver` parameter in PCA, among other things. Although I suppose that might be worth the tradeoff of making the interface better in senselab itself. We would just need to be careful every time we update scipy or umap that it doesn't break our class models here.

fabiocat93: That's something we can fairly easily check with unit tests I think. I definitely see your point, and partially agree. But dream of a package that with a good user experience. we can leave as you say for now and see how it goes.

PR #87, opened 2024-07-05
Add b2ai speaker verification functions

### LINKED ISSUE #86: Task: add b2ai speaker verification functions
### Description

Move speaker verification functions from b2aiprep

### Tasks

- [ ] verify speaker
- [ ] verify_speaker_from_files

### Freeform Notes

_No response_

ibevers:


ibevers:
- [x] update to not use big models in remote test
- [x] update tests to use the audio mono audio fixture

ibevers:
@fabiocat93 would you mind reviewing this? It's the last component of b2aiprep's process module that needs to be incorporated into senselab as far as I can tell. It would also be great if we could do a minor release when we merge it to facilitate integration into b2aiprep

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+
+import torch
+from scipy import signal
+from speechbrain.augment.time_domain import Resample
+from speechbrain.inference.speaker import SpeakerRecognition
+
+from senselab.audio.data_structures.audio import Audio
+
+
+def _resample_iir(audio: Audio, lowcut: float, new_sample_rate: int, order: int = 4) -> Audio:
```
fabiocat93: we do have resampling already in `senselab`. if you think some changes/improvements are necessary, please edit the existing method and use that instead of creating a new one

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+    sos = signal.butter(order, lowcut, btype="low", output="sos", fs=new_sample_rate)
+    filtered = torch.from_numpy(signal.sosfiltfilt(sos, audio.waveform.squeeze().numpy()).copy()).float()
+    resampler = Resample(orig_freq=audio.sampling_rate, new_freq=new_sample_rate)
+    return Audio(waveform=resampler(filtered.unsqueeze(0)).squeeze(0), sampling_rate=new_sample_rate)
+
+
+def verify_speaker(
+    audio1: Audio,
+    audio2: Audio,
+    model: str = "speechbrain/spkrec-ecapa-voxceleb",
```
fabiocat93: model should be of type SpeechbrainModel, including id and revision.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+    model_rate: int = 16000,
+    device: ty.Optional[str] = None,
```
fabiocat93: device should be of type DeviceType

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+        audio1 (Audio): The first audio sample.
+        audio2 (Audio): The second audio sample.
+        model (str): The path to the speaker verification model.
+        model_rate (int): The sample rate expected by the model.
+        device (str, optional): The device to run the model on. Defaults to None.
+
+    Returns:
+        Tuple[float, bool]: The verification score and prediction.
+    """
+    if model_rate != audio1.sampling_rate:
```
fabiocat93: in all the other tasks, if the sampling rate is not ok, we just raise an error. i would keep that policy unless you have strong opinions on that

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+        audio1 = _resample_iir(audio1, model_rate / 2 - 100, model_rate)
+    if model_rate != audio2.sampling_rate:
+        audio2 = _resample_iir(audio2, model_rate / 2 - 100, model_rate)
+
+    verification = SpeakerRecognition.from_hparams(source=model, run_opts={"device": device})
```
fabiocat93: one question: is there an advantage in using `verification.verify_batch` and not our module for extracting speaker_embeddings PLUS applying the threshold?

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+        Audio: The resampled audio signal.
+    """
+    sos = signal.butter(order, lowcut, btype="low", output="sos", fs=new_sample_rate)
+    filtered = torch.from_numpy(signal.sosfiltfilt(sos, audio.waveform.squeeze().numpy()).copy()).float()
+    resampler = Resample(orig_freq=audio.sampling_rate, new_freq=new_sample_rate)
+    return Audio(waveform=resampler(filtered.unsqueeze(0)).squeeze(0), sampling_rate=new_sample_rate)
+
+
+def verify_speaker(
+    audio1: Audio,
```
fabiocat93: all out tasks accept multiple audio clips as input. why don't you adapt this method so that it accepts something like List[tuple(Audio, Audio)]?

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,81 @@
...
+        audio1 = _resample_iir(audio1, model_rate / 2 - 100, model_rate)
+    if model_rate != audio2.sampling_rate:
+        audio2 = _resample_iir(audio2, model_rate / 2 - 100, model_rate)
+
+    verification = SpeakerRecognition.from_hparams(source=model, run_opts={"device": device})
+    score, prediction = verification.verify_batch(audio1.waveform, audio2.waveform)
+    return float(score), bool(prediction)
+
+
+def verify_speaker_from_files(
```
fabiocat93: do we need this method? all the tasks we have do work accepting as an input Audio objects and Audio.from_filepath() is our only interface to read from the file system

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
fabiocat93: and would make the threshold a parameter in our function

src/tests/audio/tasks/speaker_verification_test.py
```diff
@@ -0,0 +1,66 @@
...
+from senselab.audio.tasks.speaker_verification.speaker_verification import (
+    _resample_iir,
+    verify_speaker,
+    verify_speaker_from_files,
+)
+
+MONO_AUDIO_PATH = "src/tests/data_for_testing/audio_48khz_mono_16bits.wav"
+
+
+@pytest.fixture
```
fabiocat93: please, use the global fixture as in the other tests

fabiocat93:
hi @ibevers , i finally got the chance to review your code. i have left some comments

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Updated resample_audios.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Updated.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Updated.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Sticking to that policy--updated.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Removed.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Done.

src/tests/audio/tasks/speaker_verification_test.py
ibevers: Ohhh, I don't have to import it--couldn't import it, so I just copied it. I see that I can just use it now

ibevers: Done.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Done.

ibevers:
@fabiocat93 I incorporated your feedback. Hopefully, this is mergable now:)

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,59 @@
...
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import SpeechBrainModel
+
+
+def verify_speaker(
+    audios: List[Tuple[Audio, Audio]],
+    model: SpeechBrainModel = SpeechBrainModel(path_or_uri="speechbrain/spkrec-ecapa-voxceleb", revision="main"),
+    model_training_sample_rate: int = 16000,  # spkrec-ecapa-voxceleb trained on 16kHz audio
+    device: DeviceType = DeviceType.CPU,
+    threshold: float = 0.5,
```
fabiocat93: why `0.5`? is this what they say in the specification of the model?

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,59 @@
...
+from senselab.audio.data_structures.audio import Audio
+from senselab.audio.tasks.speaker_embeddings.speechbrain import SpeechBrainEmbeddings
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import SpeechBrainModel
+
+
+def verify_speaker(
+    audios: List[Tuple[Audio, Audio]],
+    model: SpeechBrainModel = SpeechBrainModel(path_or_uri="speechbrain/spkrec-ecapa-voxceleb", revision="main"),
+    model_training_sample_rate: int = 16000,  # spkrec-ecapa-voxceleb trained on 16kHz audio
```
fabiocat93: this is not a user parameter, but a configuration variable you should get from the model configuration. please, see https://github.com/sensein/senselab/blob/main/src/senselab/audio/tasks/speech_enhancement/speechbrain.py

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,59 @@
...
+    device: DeviceType = DeviceType.CPU,
```
fabiocat93: can you please make device's default always as None and use our _select_device_and_dtype method? this way, if the user doesn't have any preference and a GPU is available, it will be used. in your code, you force the user to use a CPU no matter what

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
fabiocat93: please check https://speechbrain.readthedocs.io/en/latest/_modules/speechbrain/inference/speaker.html#SpeakerRecognition.verify_files

src/senselab/audio/tasks/preprocessing/preprocessing.py
```diff
@@ -1,34 +1,54 @@
...
 
 from senselab.audio.data_structures.audio import Audio
 
 
-def resample_audios(audios: List[Audio], resample_rate: int, rolloff: float = 0.99) -> List[Audio]:
-    """Resamples all Audios to a given sampling rate.
-
-    Takes a list of audios and resamples each into the new sampling rate. Notably does not assume any
-    specific structure of the audios (can vary in stereo vs. mono as well as their original sampling rate)
+def resample_audios(
```
fabiocat93, line 13: thanks you @ibevers . this is helpful. i did some further research and i think we should go with an alternative implementation that is not yours or mine, but that from transforms.Resample. transforms.resample is not that different from functionals.resample, but it precomputes and reuses the resampling kernel, so using it will result in more efficient computation if resampling multiple waveforms with the same resampling parameters. they both internally do the butterworth filtering for anti-aliasing - which is why your method and my method are redundant in the same wrapping function - and then resample the signal. we can pass order and lowcut as param and compute roll off by ourselves. I would appreciate if you could refactor the code.
reference: https://pytorch.org/audio/main/generated/torchaudio.transforms.Resample.html

fabiocat93:
> @fabiocat93 I incorporated your feedback. Hopefully, this is mergable now:)

Almost! but we are very close... i feel so sorry!!

src/senselab/audio/tasks/preprocessing/preprocessing.py
satra: just an fyi that transforms.resample does not use butterworth filtering. it uses sinc interpolation with a hamming or kaiser window. in at least my initial antiliasing tests with fixed sinuisoids it was not great at creating a good filter. it still passed some amount of the signal through. i can't find the notebook right this minute, but the general idea of the test is:

create sinusoid at 14 KHz sampled at 48K, then filter down to samping rate of 16K. you should not see on an FFT any signal peak at 2K (the aliased signal). if you do, that means the anti-aliasing filter is not doing a good job. anything that far from nyquist (8K) should be completely filtered out. 

hence it may make sense to have multiple resamplers still.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Updated to match.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: I'm pretty sure the training sample rate is not available from the model configuration for `speechbrain/spkrec-ecapa-voxceleb`.

I checked the output of the instance methods of `model = SpeechBrainModel(path_or_uri="speechbrain/spkrec-ecapa-voxceleb", revision="main")` (e.g. `model.get_model_info()`).

I also get an error when I run:
```
def get_model_sample_rate(model, device=DeviceType.CPU):
    enhancer = SpeechBrainEnhancer._get_speechbrain_model(model=model, device=device)
    return enhancer.hparams.sample_rate

# Usage example
model = SpeechBrainModel(path_or_uri="speechbrain/spkrec-ecapa-voxceleb", revision="main")
sample_rate = get_model_sample_rate(model)
print(f"The model's sample rate is: {sample_rate} Hz")
```
Error output:
```
'types.SimpleNamespace' object has no attribute 'sample_rate'
```

However, the above code works when I run it with `speechbrain/sepformer-wham16k-enhancement`.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: With MPS, was getting:
```
ValueError: The requested DeviceType is either not available or                             compatible with this functionality.

src/senselab/utils/data_structures/device.py:60: ValueError
```
I excluded it from compatible_devices.

src/senselab/audio/tasks/preprocessing/preprocessing.py
ibevers: Can we hash this out in #90 and leave the resampling the way it is for this PR? @fabiocat93 @satra

ibevers:
@fabiocat93 I made some changes and responded to some of your feedback

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
fabiocat93: you are so right. in the other use case when we use ecapa-tdnn we fixed `expected_sampling_rate` to 16khz in the code (https://github.com/sensein/senselab/blob/main/src/senselab/audio/tasks/speaker_embeddings/speechbrain.py), since all speechbrain models for speaker embeddings work at 16khz. it's not ideal, but a good workaround. I think you can hardcode 16khz in the code, too and remove it from the params. we want to remove any chance for the user to make silly mistakes. thanks!! and nice catch!!

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
fabiocat93: yup. speechbrain models don't support mps yet. excluding it from compatible_devices is the way to go. idk why i cannot see your last changes though and keep seeing `device: DeviceType = DeviceType.CPU`. am i missing something?

src/senselab/audio/tasks/preprocessing/preprocessing.py
fabiocat93: i see. i think that at this point we could just remove the torchaudio implementation and use yours from b2aiprep @ibevers . this will solve 2 issues at once

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
```diff
@@ -0,0 +1,62 @@
...
+
+    Returns:
+        List[Tuple[float, bool]]: A list of tuples containing the verification score and
+                                  the prediction for each pair of audio samples. The
+                                  verification score is a float indicating the similarity
+                                  between the two samples, and the prediction is a boolean
+                                  indicating if the two samples are from the same speaker.
+    """
+    if device is None:
+        device = _select_device_and_dtype(compatible_devices=[DeviceType.CPU, DeviceType.CUDA])[0]
```
fabiocat93: do you mind calling the `_select_device_and_dtype` method even when device is not None? the way it is now, the code will brake if the user picks a device that is not supported....

fabiocat93 (commented):
99% done 👍

src/senselab/audio/tasks/preprocessing/preprocessing.py
ibevers: Done.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Updated.

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Can you see it now?

src/senselab/audio/tasks/speaker_verification/speaker_verification.py
ibevers: Thanks! Done.

ibevers:
@fabiocat93 I made another round of updates. Is it mergable now?

PR #89, opened 2024-07-09
14 add forced alignment

### LINKED ISSUE #14: Add forced alignment
### Description

Create a forced alignment task in audio folder

### Tasks

- [x] create forced alignment task
- [x] write tests

### Freeform Notes

_No response_

ibevers:


ibevers:
@fabiocat93 I integrated all your feedback from the other PR, except for using HFModel because I wasn't sure how to integrate it cleanly with `Wav2Vec2ForCTC` and `Wav2Vec2Processor` in the following usage:
```
processor = Wav2Vec2Processor.from_pretrained(model_name)
model = Wav2Vec2ForCTC.from_pretrained(model_name).to(device.value)
```

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+    SingleSegment,
+    SingleWordSegment,
+)
+from senselab.audio.tasks.preprocessing.preprocessing import resample_audios
+from senselab.utils.data_structures.device import DeviceType, _select_device_and_dtype
+from senselab.utils.data_structures.language import Language
+from senselab.utils.data_structures.script_line import ScriptLine
+
+
+def _prepare_audio(audio: Audio) -> Audio:
```
fabiocat93: why this method? our Audio objects always contain waveform which is a tensor of shape (channels, timestamps)

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+        bool: True if the segment can be aligned, False otherwise.
+    """
+    if segment["clean_char"] is None or len(segment["clean_char"]) == 0:
+        return False
+    if t1 >= max_duration:
+        return False
+    return True
+
+
+def _prepare_waveform_segment(
```
fabiocat93: we do have a method doing exactly the same in our preprocessing module...

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+        language (Language): The language of the audio (default is "en").
+
+    Returns:
+        List[List[ScriptLine]]: A list of lists, where each inner list contains the aligned script lines for each audio.
+    """
+    aligned_script_lines = []
+
+    # Define the language code and load model
+    device = _select_device_and_dtype()[0]  # DeviceType object
+    model_name = DEFAULT_ALIGN_MODELS_HF.get(language.language_code, "facebook/wav2vec2-base-960h")
```
fabiocat93: can you make the default model (facebook/wav2vec2-base-960h) a constant in your utility file constrants.py?

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+
+    processor = Wav2Vec2Processor.from_pretrained(model_name)
```
fabiocat93: please, add the revision

src/senselab/audio/tasks/forced_alignment/constants.py
```diff
@@ -0,0 +1,43 @@
...
+DEFAULT_ALIGN_MODELS_TORCH = {
+    "en": "WAV2VEC2_ASR_BASE_960H",
+    "fr": "VOXPOPULI_ASR_BASE_10K_FR",
+    "de": "VOXPOPULI_ASR_BASE_10K_DE",
+    "es": "VOXPOPULI_ASR_BASE_10K_ES",
+    "it": "VOXPOPULI_ASR_BASE_10K_IT",
+}
+
+DEFAULT_ALIGN_MODELS_HF = {
+    "ja": "jonatasgrosman/wav2vec2-large-xlsr-53-japanese",
```
fabiocat93: please, add the revision here

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+    model = Wav2Vec2ForCTC.from_pretrained(model_name).to(device.value)
+
+    for audio, transcription in audios_and_transcriptions:
+        audio = resample_audios([audio], SAMPLE_RATE)[0]  # Resample to fit expectations of Wav2Vec2
```
fabiocat93: - what if the used model is not wav2vec2?
- can you please, just raise an error if the sampling rate is not the expected rate instead of doing resampling?

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+
+        # Ensure start and end are not None
+        start = transcription.start if transcription.start is not None else 0.0
```
fabiocat93, line 656: are you sure this is possible? Scriptline should already take care of all this. and if not, those checks should go there

src/senselab/audio/tasks/forced_alignment/constants.py
fabiocat93: consider making those TorchModel and HFModel objects as in the rest of our package

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -0,0 +1,747 @@
...
+        ]
+
+        with threadpool_limits(limits=1, user_api="blas"):
+            alignment = _align_transcription(
+                transcript=segments,
+                model=model,
+                align_model_metadata={
+                    "dictionary": processor.tokenizer.get_vocab(),
+                    "language": language.language_code,
+                    "type": "huggingface",
```
fabiocat93: is this always "huggingface"? if you use the HFModel and TorchModel classes, you don't need this param anymore and your code looks prettier and is more readable

fabiocat93: ofc these changes would produce some cascade changes in the rest of the code

fabiocat93:
> @fabiocat93 I integrated all your feedback from the other PR, except for using HFModel because I wasn't sure how to integrate it cleanly with `Wav2Vec2ForCTC` and `Wav2Vec2Processor` in the following usage:
> 
> ```
> processor = Wav2Vec2Processor.from_pretrained(model_name)
> model = Wav2Vec2ForCTC.from_pretrained(model_name).to(device.value)
> ```

thank you @ibevers . you did a great job. i think you only need to integrate it better with the rest of senselab, using the functions that we do already have and the general logics that has been applied in the rest of the package

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: Removed.

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: This one is a little different because it pads the audio segment, but I updated the preprocessing module and replaced this function.

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: Added

src/senselab/audio/tasks/forced_alignment/constants.py
ibevers: Done

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: Done.

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: For now, it will only support wav2vec2-based architectures. There is no model parameter to align_transcriptions, so a user passing the wrong model is not a possible problem. Updated to raise an error instead of doing resampling.

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: Ah, I see. Removed.

ibevers: Sometimes we want might want a scriptline that has a None start value though (e.g. just a transcript). Also, we can't update the end time without an audio file, and we can't assume someone has an audio file when they make a scriptline object.

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
ibevers: Updated to use torchaudio when appropriate. Going to leave it the way it is for the sake of expediency...

ibevers:
@fabiocat93 Is there anything else that needs to be done, or can we merge?

src/senselab/audio/tasks/preprocessing/preprocessing.py
```diff
@@ -103,40 +103,111 @@ def select_channel_from_audios(audios: List[Audio], channel_index: int) -> List[
...
+                    sampling_rate=audio.sampling_rate,
+                    metadata=audio.metadata.copy(),
+                    orig_path_or_id=audio.orig_path_or_id,
+                )
             )
+        extracted_segments.append(single_audio_segments)
+    return extracted_segments
+
+
+def pad_audios(audios: List[Audio], desired_samples: int) -> List[Audio]:
```
fabiocat93, line 156: @ibevers does this work with stereo audio recordings?

src/senselab/audio/tasks/preprocessing/preprocessing.py
```diff
@@ -103,40 +103,111 @@ def select_channel_from_audios(audios: List[Audio], channel_index: int) -> List[
...
+            sampling_rate=audio.sampling_rate,
+            metadata=audio.metadata.copy(),
+            orig_path_or_id=audio.orig_path_or_id,
         )
-    return chunked_audios
+        padded_audios.append(padded_audio)
+    return padded_audios
+
+
+def evenly_segment_audios(
```
fabiocat93, line 185: same as above. not sure if people will actually need this functionality, but ok

src/senselab/utils/data_structures/script_line.py
```diff
@@ -15,15 +16,15 @@ class ScriptLine(BaseModel):
 
     text: Optional[str] = None
     speaker: Optional[str] = None
-    start: Optional[float] = None
-    end: Optional[float] = None
+    start: Optional[float] = 0
```
fabiocat93: why this?

fabiocat93: conceptually, if there is no start or end, I do prefer to have None instead of 0.

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+)
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.language import Language
+from senselab.utils.data_structures.script_line import ScriptLine
+
+MONO_AUDIO_PATH = "src/tests/data_for_testing/audio_48khz_mono_16bits.wav"
+
+
+@pytest.fixture
+def mono_audio_sample() -> Audio:
```
fabiocat93: please, use the global fixture as we do in the other tests

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+    """Fixture for sample mono audio."""
+    return Audio.from_filepath(MONO_AUDIO_PATH)
+
+
+class SingleCharSegment(TypedDict):
```
fabiocat93: how is this useful?

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+    assert interpolated_series.isnull().sum() == 0
+
+
+if os.getenv("GITHUB_ACTIONS") != "true":
+
+    def test_get_prediction_matrix(dummy_model: tuple) -> None:
+        """Test generation of prediction matrix."""
+        model, _ = dummy_model
+        waveform_segment = torch.randn(1, 16000)
+        prediction_matrix = _get_prediction_matrix(model, waveform_segment, None, "huggingface", DeviceType.CPU)
```
fabiocat93, line 111: why do you need a model type? the type is an intrinsic characteristic of the model

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+        """Test alignment of segments."""
+        model, processor = dummy_model
+        model_dictionary = processor.tokenizer.get_vocab()
+
+        # Create a sample transcript
+        transcript = [SingleSegment(start=0.0, end=1.0, text="test")]
+
+        # Preprocess the transcript segments
+        preprocessed_transcript = _preprocess_segments(
+            transcript, model_dictionary, model_lang="en", print_progress=False, combined_progress=False
```
fabiocat93: why lang is a string?

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+        # Ensure the model dictionary has the necessary keys
+        for char in "test":
+            if char not in model_dictionary:
+                model_dictionary[char] = len(model_dictionary)
+
+        aligned_segments, word_segments = _align_segments(
+            transcript=preprocessed_transcript,
+            model=model,
+            model_dictionary=model_dictionary,
+            model_lang="en",
```
fabiocat93: same as above

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+            model_type="huggingface",
```
fabiocat93, line 141: same as above

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+        aligned_transcriptions = align_transcriptions(audios_and_transcriptions)
+        assert len(aligned_transcriptions) == 1
+        assert len(aligned_transcriptions[0]) == 1
+        assert aligned_transcriptions[0][0].text == "test"
+
+    def test_align_transcriptions_multilingual(
+        resampled_mono_audio_sample: Audio, script_line_fixture: ScriptLine
+    ) -> None:
+        """Test alignment of transcriptions."""
+        languages = ["de", "es"]
```
fabiocat93, line 194: why is language a string?

src/tests/audio/tasks/forced_alignment_test.py
```diff
@@ -0,0 +1,215 @@
...
+        expected_text = "test"  # Replace with the appropriate expected text for your fixtures
+
+        for lang in languages:
+            audios_and_transcriptions = [(resampled_mono_audio_sample, script_line_fixture)]
+            aligned_transcriptions = align_transcriptions(audios_and_transcriptions, Language(language_code=lang))
```
fabiocat93: why isn't language part of the tuple with audio and transcript?

src/senselab/audio/tasks/preprocessing/preprocessing.py
ibevers: Yes, `padded_waveform = torch.nn.functional.pad(audio.waveform, (0, padding_needed))` pads along the time dimension, so each channel of the audio is padded. I updated the test for this function to test for this.

src/senselab/audio/tasks/preprocessing/preprocessing.py
ibevers: Yes, it works for stereo. I updated the test to check for this. I included this functionality because Summer School students were doing this with the B2AI audio files.

src/senselab/utils/data_structures/script_line.py
ibevers: I agree. Undone. I think I did this because of a mypy issue or for testing something temporarily

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Fixed

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Haha, great question--it's not. Removed.

src/senselab/audio/tasks/preprocessing/preprocessing.py
```diff
@@ -103,37 +103,105 @@ def select_channel_from_audios(audios: List[Audio], channel_index: int) -> List[
     return mono_channel_audios
 
 
-def chunk_audios(data: List[Tuple[Audio, Tuple[float, float]]]) -> List[Audio]:
```
fabiocat93: @ibevers why did you replaced `chunk_audios` with `extract_segments`? they don't do they same thing. `extract_segments` may use `chunk_audios` under the hood.

fabiocat93:
@ibevers what's the progress with this? did you find any blockers?

ibevers:
> @ibevers what's the progress with this? did you find any blockers?

I've been slowly addressing your feedback. I haven't found any blockers. I just deprioritized this to focus on the b2aiprep integration and then the reviews, but I aim to address the rest of your feedback today.

src/tests/audio/tasks/forced_alignment_test.py
ibevers: @fabiocat93 As we discussed before, since HFModel and TorchAudioModel don't have an instance variable for the model itself (Out of curiosity, why they don't they have this?), it makes more sense to pass the model as a `torch.nn.Module`. As far as I know `torch.nn.Module` doesn't have a direct way to tell whether a model is a torch audio model or a hugging face model, so for the sake of making the code more interpretable and reliable and because it required fewer changes, I decided to pass a string with the model type.

Which of the following approaches do you think is best?

Possible:
- Do nothing. Adds bulk to the function signatures, and makes the code a little rigid.

- Load the model in  `_get_prediction_matrix`. That would slow down performance since it would be loaded multiple times. 

- Add a factory. This might cause issues with Pydra that we discussed.

- Update the HFModel and TorchAudioModel types to store a reference to the loaded model and then pass those. This could be an optional instance variable to avoid causing problems with existing usage.

- Use intrinsic features of `torch.nn.Module` such as the solution below. This would be less transparent and reliable.
```
def _get_prediction_matrix(
    model: torch.nn.Module,
    waveform_segment: torch.Tensor,
    lengths: Optional[torch.Tensor],
    device: DeviceType,
) -> torch.Tensor:
    """Generate prediction matrix from the alignment model.

    Args:
        model (torch.nn.Module): The alignment model.
        waveform_segment (torch.Tensor): The audio segment to be processed.
        lengths (Optional[torch.Tensor]): Lengths of the audio segments.
        device (DeviceType): The device to run the model on.

    Returns:
        torch.Tensor: The prediction matrix.
    """
    with torch.inference_mode():
        if hasattr(model, 'forward') and 'torchaudio' in str(type(model)):
            emissions, _ = model(waveform_segment.to(device.value), lengths=lengths)
        elif hasattr(model, 'logits') or hasattr(model, 'forward') and 'transformers' in str(type(model)):
            emissions = model(waveform_segment.to(device.value)).logits
        else:
            raise NotImplementedError(f"Align model of type {type(model)} not supported.")

        emissions = torch.log_softmax(emissions, dim=-1)

    return emissions
```

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Because the main point of using the Language object was to validate the language code, which is done in the in interface function `align_transcriptions`, and I didn't have time to update it everywhere else. I just changed it though.

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Changed

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Changed

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Now, these are used to generate the Language object which is passed to the function.

src/tests/audio/tasks/forced_alignment_test.py
ibevers: Good question--now it is

src/senselab/audio/tasks/preprocessing/preprocessing.py
ibevers: With chunk audios, I don't immediately know whether two extracted chunks are from the same audio file, whereas I do with `extract_segments`. I updated `extract_segments` to use `chunk_audios`.

ibevers:
@fabiocat93 I finished incorporating your latest feedback

fabiocat93:
> @fabiocat93 I finished incorporating your latest feedback

thank you @ibevers !! hopefully, people with start use this task soon! We still miss a tutorial with this. you may integrate it as part of the speech to text tutorial. I am going to merge this meanwhile

fabiocat93 (approved):
I accept this since it's already in a good status, but still needs tutorial and peer-review.

ibevers:
Thanks! I will add making the tutorial to my to-do list.

PR #93, opened 2024-07-11
adjusting dependencies and github actions plus adding pooling methods as a task plus fixing speechbrain issues plus adding some documentation

fabiocat93:
- adjusting dependencies and github actions 
- adding pooling methods as a task
- fixing speechbrain issues
- adding some documentation


fabiocat93:
this also includes `pooling` as a new utility task

PR #96, opened 2024-07-12
:sparkles: add pooling methods

GasserElbanna:
Adding mean, max, and min pooling methods for torch tensors.


fabiocat93:
thank you @GasserElbanna , it looks great. do you mind also adding some unit tests, please?

PR #114, opened 2024-07-23
Update pyproject.toml

satra:
Test changing caret to tilde for package specification


satra:
we can add ubuntu/macos arm 3.11 and 3.12 tests as required to the branch protection requirements as they are likely to be the most common use cases. i didn't enable windows as i haven't used a windows machine in a while.

fabiocat93 (approved):

PR #115, opened 2024-07-23
Pydra

wilke0818:
- Imports pydra serialization for tensors
- Currently has the serializer in 2 places, the \_\_init__ and in a helper function module, should choose one and use that
- Has some miscelanrous bug fixes for audio classification

### DIFF
--- a/src/senselab/__init__.py
+++ b/src/senselab/__init__.py
@@ -1,22 +1,7 @@
 """.. include:: ../../README.md"""  # noqa: D415
 
-from typing import Iterator
-
 import nest_asyncio
-import numpy as np
-import torch
-from pydra.utils.hash import Cache, bytes_repr_sequence_contents, register_serializer
+
+from senselab.utils.data_structures.pydra_helpers import *  # NOQA
 
 nest_asyncio.apply()
-
-
-@register_serializer(torch.Tensor)
-def bytes_repr_arraylike(obj: torch.Tensor, cache: Cache) -> Iterator[bytes]:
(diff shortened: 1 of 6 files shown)


src/senselab/__init__.py
```diff
@@ -1 +1,22 @@
...
+
+import nest_asyncio
+import numpy as np
+import torch
+from pydra.utils.hash import Cache, bytes_repr_sequence_contents, register_serializer
+
+nest_asyncio.apply()
+
+
+@register_serializer(torch.Tensor)
```
fabiocat93: hey @wilke0818 why is the `bytes_repr_arraylike` both here and in the `pydra_helpers.py`?

wilke0818: well that's part of the question I had: where do we want this to live? If we leave it in `pydra_helpers.py` then we need to import it in the `init`. I wasn't sure which made more sense especially as I am not sure if we will have more pydra related utilities? You can kind of see the notes I was making for what a Pydra class could look like.

fabiocat93: i like the idea of having it in pydra_helpers and import it in the init. we will probably need something similar for setting up global settings for huggingface and other libraries at some point. just make a decision and let's see how it works for now. we can always change if we realize that there are better options

wilke0818: done. we should look at using \_\_all__ though I don't fully understand it/would require a bit more restructuring than I was originally intending with this PR

fabiocat93 (approved):

PR #119, opened 2024-07-26
109 review and test speech to text task

### LINKED ISSUE #109: Review and test speech-to-text task
### Description

This issue is to track feedback for the senselab speech-to-text task.

### Tasks

- [x] Review tutorial
- [x] Review documentation
- [x] Review code
- [x] Test edge cases

### Freeform Notes

_No response_

fabiocat93:
@900miles i move the discussion here from https://github.com/sensein/senselab/issues/109. let me know if you see any more changes to be done


src/senselab/audio/tasks/speech_to_text/api.py
```diff
@@ -13,7 +16,11 @@
 
 
 def transcribe_audios(
-    audios: List[Audio], model: SenselabModel, language: Optional[Language] = None, device: Optional[DeviceType] = None
+    audios: List[Audio],
+    model: SenselabModel,
```
900miles, line 20: This asks for `SenselabModel`, but the code requires it to be of type `HFModel`

src/senselab/audio/tasks/speech_to_text/api.py
```diff
@@ -22,16 +29,20 @@ def transcribe_audios(
         model (SenselabModel): The model used for transcription.
         language (Optional[Language]): The language of the audio (default is None).
         device (Optional[DeviceType]): The device to run the model on (default is None).
```
900miles, line 31: Saying "default is None" is highly misleading, as behind the scenes it chooses the fastest device to run on if device is not specified here. Perhaps it should clarify that it defaults to fastest available device.

src/senselab/audio/tasks/speech_to_text/api.py
```diff
@@ -22,16 +29,20 @@ def transcribe_audios(
...
-
 
-transcribe_audios_pt = pydra.mark.task(transcribe_audios)
+    Todo:
+        - Include more models (e.g., speechbrain, nvidia)
+    """
+    try:
+        if isinstance(model, HFModel):
+            return HuggingFaceASR.transcribe_audios_with_transformers(
+                audios=audios, model=model, language=language, device=device, **kwargs
```
900miles, line 43: You pass `**kwargs` to `transcribe_audios_with_transformers` but that function does not take `**kwargs` as an input. As the parameters for the function are limited, and we have control over that function, I wonder if we should write them out specifically as parameters for this function as well? Or otherwise we should point to that function in the documentation of this one so users know what can go into `**kwargs` here.

src/senselab/audio/tasks/speech_to_text_evaluation/doc.md
```diff
@@ -0,0 +1,131 @@
...
+- \( C \) = Number of correct words
+- \( N \) = Number of words in the reference
+- \( P \) = Number of words in the prediction
+
+The higher the value, the better the performance of the ASR system, with 1 being a perfect score.
+
+## Examples
+For a given reference "hello world" and hypothesis "hello duck":
+
+- **WER:** \( \frac{0 + 1 + 0}{2} = 0.5 \)
```
900miles, line 127: These LaTeX commands don't render properly (\frac and \approx)

tutorials/speech_to_text.ipynb
900miles, line 1: `Let's preprocess the audio data to make it suitable with the ASR model characteristics, that we can find in the model card in the huggingface Hub.` should not have a comma in it.

900miles, line 1: That's a really cool way to visualize the transcript!

src/senselab/audio/tasks/speech_to_text/huggingface.py
```diff
@@ -128,9 +133,42 @@ def _rename_key_recursive(obj: Dict[str, Any], old_key: str, new_key: str) -> Di
             device=device,
         )
 
+        # Take the end time of the pipeline initialization
+        end_time_pipeline = time.time()
+        # Print the time taken for initialize the hugging face ASR pipeline
+        elapsed_time_pipeline = end_time_pipeline - start_time_pipeline
+        logger.info(f"Time taken for initialize the hugging face ASR pipeline: {elapsed_time_pipeline:.2f} seconds")
```
900miles, line 140: This should say "Time taken **to** initialize the hugging face ..."

900miles (commented):
Hi @fabiocat93. Looks good! Please take a look at the comments I made, mostly minor things.

900miles:
I'm not sure if it's possible to re-open a merged pull request. Should this go on a new PR? Not sure of the procedure here.

src/senselab/audio/tasks/speech_to_text/api.py
fabiocat93: which is correct, because a HFModel is a SenselabModel. We want to keep it open to other types of models that we will hopefully integrate soon

src/senselab/audio/tasks/speech_to_text/api.py
fabiocat93: device means "preferred device". Which is why we accept None in case the user doesn't have any preference and we can pick (based on performance). After some brainstorming we decided for device because it's shorter than preferred_device, but if that's so misleading we can change it everywhere. I have no strong preference. Feel free to open a PR changing this all over the package. I write a not about this

src/senselab/audio/tasks/speech_to_text/api.py
fabiocat93: this is on purpose. `**kwargs` in the general api helps accepting all kind of params for the different models, and then the internal function (`transcribe_audios_with_transformers`) does the validation of the params, which is why it doesn't include `**kwargs`. the accepted params are documented in `transcribe_audios_with_transformers`.

src/senselab/audio/tasks/speech_to_text_evaluation/doc.md
fabiocat93: nice catch. this has been fixed and released already

fabiocat93:
> I'm not sure if it's possible to re-open a merged pull request. Should this go on a new PR? Not sure of the procedure here.

Thank you @900miles. I think that by the time you commented, I had already merged the edits (at least to have something to show during tomorrow's meeting). i have replied to some of your comments and fixed most of them as part of a different PR. you can find all implemented changes in the main branch already

src/senselab/audio/tasks/speech_to_text/api.py
900miles: Got it. Perhaps we should put that in the docstring then. Right now a `SenselabModel` that is not an `HFModel` will pass type checking and there is nothing in the docstring that says the function only accepts an `HFModel`

src/senselab/audio/tasks/speech_to_text/api.py
900miles: What is the main entry point that should be used for speech to text? The `api` documentation does not mention `transcribe_audios_with_transformers`, so if somebody uses that function as their interface for speech to text, they won't know what options are available for `**kwargs`.

src/senselab/audio/tasks/speech_to_text/api.py
900miles: Got it. I don't have a strong opinion on this either but I was confused by it. Is this behavior documented somewhere in senselab?

PR #129, opened 2024-08-07
Enhancing text to speech

### LINKED ISSUE #108: Review and test text-to-speech
### Description

This issue is to track feedback for the senselab text-to-speech task.

### Tasks

- [x] Review tutorial
- [x] Review documentation
- [x] Review code
- [x] Test edge cases

### Freeform Notes
Fabio's documentation guidelines:
- Documentation - Considering the tasks that you have implemented, work on the documentation
- Explain what the task is about, give an overview of the metrics generally used to evaluate the task, report some existing benchmarks in the literature, if any
- Make docstrings in Google style and consistent with with the CONTRIBUTING.md document

Notebook tutorial

- Prepare a jupyter notebook tutorial for each of your tasks. This should show how to use the senselab interface that you have implemented, both with and without using pydra

fabiocat93:
- added mars5tts
- added tutorial
- improved documentation
- addressed @900miles's review


fabiocat93:
> There are no tutorials for this task. There is no explanation for what the task is about, what the metrics are for evaluating the tasks, nor any benchmarks in the literature. There are no examples in the docstring of how the function should be used. The doctring says that model should be of type SenselabModel but the code requires it to be of type HFModel.

@900miles documentation was added and docstrings adjusted

fabiocat93:
> When running synthesize_texts, I get the following warning:  
"The attention mask and the pad token id were not set. As a consequence, you may observe unexpected behavior. Please pass your input's `attention_mask` to obtain reliable results.
Setting `pad_token_id` to `eos_token_id`:10000 for open-end generation." 
This seems to happen once per text synthesized.


@900miles this is a warning that is specific to the bark model. It is nothing we should be worried about for now

fabiocat93:
> The supported models seems to be incorrect. The documentation is vague, but seems to imply that any huggingface tts model is supported. However, I have tried microsoft/speecht5_tts, microsoft/speecht5_tts, fishaudio/fish-speech-1.2-sft, and facebook/mms-tts-eng and all have failed with various errors. suno/bark is the only one I have tried so far that works.


@900miles I have now specified some of the models that we recommend. I have also clarified that not every model is supported, despite they wrongly labeled it with the text-to-speech label. microsoft/speecht5_tts, facebook/mms-tts-eng, suno/bark are in the tests/tutorial and work fine. fishaudio/fish-speech-1.2-sft was never able to make it work

fabiocat93:
> The tests seem to be missing two cases: 1. testing with different models other than the default and 2. testing the case of synthesizing more than one text at one time.

@900miles fixed

fabiocat93:
@900miles this should be ready for a second review whenever you have time

fabiocat93:
@wilke0818 any progress update with styletts2?

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -1,37 +1,109 @@
...
+    **kwargs: Any,  # noqa: ANN401
 ) -> List[Audio]:
     """Synthesizes speech from all texts using the given model.
 
     Args:
         texts (List[str]): The list of text strings to be synthesized.
-        model (SenselabModel): The model used for synthesis
-            (Default is "suno/bark").
+        target (Optional[List[Tuple[Audio, Optional[str]]]]):
+            A list where each element is a tuple of target audio and optional transcript.
```
900miles: A bit confusing but it looks like the code also accepts a list of Audios, not just a list of tuples.

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -1,37 +1,109 @@
...
+        model (HFModel): The model used for synthesis (Default is "suno/bark").
```
900miles: The documentation says model needs to be of type `HFModel` but the code allows it to be of type `HFModel` OR `TorchModel` and the function header allows any `SenselabModel`. These should be reconciled.

src/senselab/audio/tasks/text_to_speech/api.py
900miles: The code also allows target to be None. A couple words on what's going on might be helpful. Maybe something saying that how you specify target depends on which model is being used.

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -1,37 +1,109 @@
...
+                else:
+                    raise TypeError("Expected the second element of the tuple to be a string.")
+            elif isinstance(t, Audio):
+                target_audios.append(t)
+                target_transcripts.append(None)
+            else:
+                raise TypeError("Expected elements of target to be either Audio objects or tuples of (Audio, str).")
+
+        if len(texts) != len(target_audios):
+            raise ValueError("The lengths of texts and target audios must be the same.")
```
900miles: "The lists of texts and target audios must have the same length." or something along those lines

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -1,37 +1,109 @@
...
+        if any(t is not None for t in target_transcripts) and len(texts) != len(target_transcripts):
+            raise ValueError("The lengths of texts and target transcripts must be the same.")
```
900miles: "The lists of texts and target transcripts must have the same length." or something along those lines

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -1,37 +1,109 @@
...
+
     if isinstance(model, HFModel):
-        return HuggingFaceTTS.synthesize_texts_with_transformers(texts=texts, model=model, device=device)
+        return HuggingFaceTTS.synthesize_texts_with_transformers(texts=texts, model=model, device=device, **kwargs)
+    elif isinstance(model, TorchModel):
+        if model.path_or_uri == "Camb-ai/mars5-tts":
+            # Converting target to the required format for Mars5TTS
+            target_for_mars5tts = _get_audio_and_transcript_targets(target_audios, target_transcripts)
```
900miles: Why do you unzip `target` and then rezip `target` instead of just using `target` directly here?

src/senselab/audio/tasks/text_to_speech/doc.md
```diff
@@ -0,0 +1,60 @@
...
+
+## Models
+
+A variety of models are supported by `senselab` for text-to-speech.
+Each model varies in performance, size, license, language support, and more. Performance may also vary depending, among other reasons, on the length of the text or the target speaker (differences in terms of age, dialects, disfluencies). It is recommended to review the model card for each model before use and refer to the most recent literature for an informed decision.
+
+Several text-to-speech models are currently available through `🤗 Transformers`. These models can be explored on the [Hugging Face Hub](https://huggingface.co/models?library=transformers&pipeline_tag=text-to-speech&sort=downloads).
+**Note**: Some `Hugging Face` models, despite having the `text-to-speech` label on their model cards, may not work with the text-to-speech pipeline. These models are not supported in `senselab`, and identifying them often requires trial and error.
+
+In addition to the models from `🤗 Transformers`, senselab also supports `Mars5-TTS`, which provides text-to-speech capabilities with a target voice (and its transcript) in English.
```
900miles: It's a little unclear what "[it] provides text-to-speech capabilities with a target voice (and its transcript) in English" means exactly.

src/senselab/audio/tasks/text_to_speech/marstts.py
```diff
@@ -0,0 +1,150 @@
...
+                The only supported language is "en" for now.
+            device (DeviceType): The device to run the model on (default is None). Supported devices are CPU and CUDA.
+
+        Returns:
+            model: The Torch-based Mars5TTS model.
+            config_class: The configuration class used by the model.
+        """
+        if model.path_or_uri != "Camb-ai/mars5-tts" or model.revision != "master":
+            raise NotImplementedError("Only the 'Camb-ai/mars5-tts' model is supported for now.")
+        if language == Language(language_code="en"):
```
900miles, line 41: This requires `language="en"`, which will make the function throw an error with the default setting (`language=None`)

src/senselab/audio/tasks/text_to_speech/marstts.py
```diff
@@ -0,0 +1,150 @@
...
+        # Check that the target audios are mono and have the correct sampling rate
+        expected_sampling_rate = my_model.sr
+        for item in target:
+            target_audio, _ = item
+            if target_audio.waveform.shape[0] != 1:
+                raise ValueError(f"Stereo audio is not supported. Got {target_audio.waveform.shape[0]} channels")
+            if target_audio.sampling_rate != expected_sampling_rate:
+                raise ValueError(
+                    f"Incorrect sampling rate. Expected {expected_sampling_rate}, " f"got {target_audio.sampling_rate}"
+                )
```
900miles, line 130: These errors should specify which audio is the offending audio.

900miles (commented):
Looks good so far, I've left some comments. I've still got a few files and the tutorial to review.

src/senselab/audio/tasks/text_to_speech/api.py
fabiocat93: fair. i unzip it to make some checks that are not specific to the model. and then, if pass, I structure the target in the format accepted by the mars5tts function. i see how this is not really optimized, any suggestion is welcome

fabiocat93:
thank you @900miles , i think i have addressed your comments. 
@wilke0818 after yesterday's meeting, what do we want to do with `styletts2`? i honestly think that, if we go with a docker-based solution, we will create some challenge for entry-level people who want to start learning about audio signal processing and this is something we don't want of course

wilke0818:
@fabiocat93 I am not sure how hard it would be for new users as theoretically the Docker would be under the hood. However, as to not delay these other tools and in order to give us more time to think about this, I would recommend reverting my commit, I don't know if it is easily revertible since you've committed more, but we could honestly remove the StyleTTS code (though perhaps make sure to push it into a seperate branch so it doesn't have to be rewritten). Do you want to do this? Or I can but it might be near the end of the day.

fabiocat93:
> @fabiocat93 I am not sure how hard it would be for new users as theoretically the Docker would be under the hood. However, as to not delay these other tools and in order to give us more time to think about this, I would recommend reverting my commit, I don't know if it is easily revertible since you've committed more, but we could honestly remove the StyleTTS code (though perhaps make sure to push it into a seperate branch so it doesn't have to be rewritten). Do you want to do this? Or I can but it might be near the end of the day.

That makes sense. I don't have a buffer for doing it today. I would appreciate it if you could do it. Otherwise, I can find some time tomorrow night.
That said, let's make sure we continue the docker-related discussion in the new PR thread

pyproject.toml
```diff
@@ -56,6 +56,7 @@ numpy = "~=1.25"
 umap-learn = "~=0.5"
 scikit-learn = "~=1.5"
 nltk = "~=3.8"
+vocos = "~=0.1"
```
wilke0818, line 59: was this for Mars or StyleTTS?

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
```diff
@@ -300,10 +300,12 @@ def _align_single_segment(
     tokens = [model_dictionary[c] for c in text_clean]
 
     extracted_segment = extract_segments([(audio, [(t1, t2)])])[0][0]
-    lengths = extracted_segment.waveform.shape[-1]
+    lengths = torch.tensor([extracted_segment.waveform.shape[-1]])
     waveform_segment = pad_audios([extracted_segment], MINIMUM_SEGMENT_SIZE)[0].waveform
 
-    emissions = _get_prediction_matrix(model, waveform_segment, lengths, model_type, device)
+    emissions = _get_prediction_matrix(
```
wilke0818, line 306: why is this in this PR?

src/senselab/audio/tasks/speaker_embeddings/speechbrain.py
```diff
@@ -92,7 +92,7 @@ def extract_speechbrain_speaker_embeddings_from_audios(
 
         # Stack audio waveforms for batch processing, zero-padding to ensure equal length
         waveforms = torch.stack(
-            [F.pad(audio.waveform, (0, max_len - audio.waveform.shape[1])) for audio in audios]
+            [F.pad(audio.waveform, (0, int(max_len - audio.waveform.shape[1]))) for audio in audios]
```
wilke0818, line 95: why is this in this PR?

src/senselab/audio/tasks/text_to_speech/__init__.py
```diff
@@ -1 +1,3 @@
-"""This module provides the API for the senselab text to speech task."""
+""".. include:: ./doc.md"""  # noqa: D415
```
wilke0818, line 1: If it becomes common for us to use the doc.md method then let's add D415 to the ignore list.

src/senselab/audio/tasks/text_to_speech/huggingface.py
```diff
@@ -49,6 +49,7 @@ def synthesize_texts_with_transformers(
         texts: List[str],
         model: HFModel = HFModel(path_or_uri="suno/bark", revision="main"),
         device: Optional[DeviceType] = None,
+        forward_params: Optional[Dict[str, Any]] = None,
```
wilke0818, line 52: how do these differ from simple **kwargs?

src/senselab/audio/tasks/text_to_speech/styletts2.py
wilke0818, line 1: this file is generally unused but I think keeping it as an additional example of how to go about integrating new models is good. Also eventually it would be nice to get this integrated if possible.

src/senselab/utils/tasks/cross_correlation.py
```diff
@@ -31,19 +31,19 @@ def compute_normalized_cross_correlation(signal1: torch.Tensor, signal2: torch.T
         raise ValueError("Input signals must be one-dimensional")
 
     # Convert PyTorch tensors to NumPy arrays
-    signal1 = signal1.numpy()
-    signal2 = signal2.numpy()
+    signal1_np: np.ndarray = signal1.numpy()
```
wilke0818, line 34: again, is this supposed to be in this PR or is it just a weird Github thing happening?

src/tests/audio/tasks/text_to_speech_test.py
```diff
@@ -3,31 +3,100 @@
...
+    #         language=language,
+    #         force_reload=True,
+    #     )
+
+    # assert len(audios) == 2
+    # assert isinstance(audios[0], Audio)
+    # assert audios[0].waveform is not None
+    # assert audios[0].sampling_rate == terget_audio_resampling_rate
+
+    def test_synthesize_texts_with_mars5_model(mars5_model: TorchModel, mono_audio_sample: Audio) -> None:
```
wilke0818, line 71: this test takes a really long time but not sure if there is anything we can do about that.

wilke0818 (approved):
Assuming @900miles reviews have been addressed, I think this is good to go

pyproject.toml
fabiocat93: mars5

src/senselab/audio/tasks/forced_alignment/forced_alignment.py
fabiocat93: good question. that's my bad. probably somehow there was a style issue in the main branch and I fixed it since couldn't commit otherwise

src/senselab/audio/tasks/speaker_embeddings/speechbrain.py
fabiocat93: same as above (mypy issue)

900miles:
Sorry, haven't had a chance to finish my review yet. I'll get it done by end-of-day today

src/senselab/audio/tasks/text_to_speech/__init__.py
fabiocat93: I would recommend keeping track of how often this method is used and evaluating if it’s truly beneficial before making the change permanent

src/senselab/audio/tasks/text_to_speech/huggingface.py
fabiocat93: not really different. if you want, i can make it with `**kwargs`

src/senselab/audio/tasks/text_to_speech/styletts2.py
fabiocat93: maybe let's add a comment specifying that this is in progress somewhere?

src/senselab/utils/tasks/cross_correlation.py
fabiocat93: same as above. sorry

src/tests/audio/tasks/text_to_speech_test.py
fabiocat93: i just think that the problem is with it running very inefficiently on the CPU....

fabiocat93:
> Sorry, haven't had a chance to finish my review yet. I'll get it done by end-of-day today

no prob

tutorials/text_to_speech.ipynb
```diff
@@ -0,0 +1,283 @@
...
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "## Quick start\n",
+                "We will start with some ```HuggingFace``` models. \n",
+                "\n",
+                "The very first example shows how to use ```facebook/mms-tts-eng``` which does just require as input the list of pieces of text that you want to synthetize."
```
900miles: "which just requires as input ..."

tutorials/text_to_speech.ipynb
```diff
@@ -0,0 +1,283 @@
...
+                "\n",
+                "# Play the synthesized audio\n",
+                "play_audio(audios[0])"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
+                "Let's proceed with ```microsoft/speecht5_tts``` (https://huggingface.co/microsoft/speecht5_tts). This model requires the list of pieces of text to synthetize plus the speaker embedding of the voice we want to clone (btw, speaker embeddings are some values describing the characteristics of someone's voice). Details about ```microsoft/speecht5_tts```can be found in the model card. In our example, we use some speaker embeddings from the dataset called ```Matthijs/cmu-arctic-xvectors```."
```
900miles: We should add a few words on how the user can extract their own speaker embeddings and link to our docs on speaker embeddings

tutorials/text_to_speech.ipynb
```diff
@@ -0,0 +1,283 @@
...
+            "source": [
+                "model = TorchModel(path_or_uri=\"Camb-ai/mars5-tts\", revision=\"master\")\n",
+                "language = Language(language_code=\"en\")\n",
+                "device = DeviceType.CPU"
+            ]
+        },
+        {
+            "cell_type": "markdown",
+            "metadata": {},
+            "source": [
```
wilke0818, line 171: targets is supposed to be a tuple, not a list. code errors.

900miles: I get the same error. Line 239

tutorials/text_to_speech.ipynb
```diff
@@ -0,0 +1,283 @@
...
+                "Note: If you want to specify more params and customize the process, you can do it. For more details, see the [**dedicated documentation**](https://sensein.group/senselab/senselab/audio/tasks/text_to_speech.html)."
+            ]
+        },
+        {
+            "cell_type": "code",
+            "execution_count": null,
+            "metadata": {},
+            "outputs": [],
+            "source": [
+                "res = synthesize_texts(texts=[\"Hello, world. It's nice to meet you.\"], \n",
```
900miles, line 238: This line took 14 minutes to generate, is that expected?

src/senselab/audio/tasks/text_to_speech/api.py
900miles: Could you just do the checks without creating new variables / functions to unzip and rezip? Like have the loop in the beginning just do the checks, and if they pass, send `target` into `mars5tts`?

900miles (changes requested):
Looks great just a couple more comments

tutorials/text_to_speech.ipynb
wilke0818: i think that's about what I had when running on CPU. don't think there is much that can be done about that.

src/senselab/audio/tasks/text_to_speech/api.py
fabiocat93: @900miles the problem with the approach you suggest is that you cannot guarantee data type consistency and mypy complains about that:
`Argument "target" to "synthesize_texts_with_mars5tts" of "Mars5TTS" has incompatible type "list[tuple[Audio, str | None]] | None"; expected "list[tuple[Audio, str]]"  [arg-type]`
zipping `target` again is a workaround to make the code work and `mypy` happy

tutorials/text_to_speech.ipynb
fabiocat93: fyi, it took 20 seconds on my laptop

src/senselab/audio/tasks/text_to_speech/api.py
wilke0818: I can take a look at this later. there are definitely approaches that will make mypy happy that don't require zipping/unzipping

tutorials/text_to_speech.ipynb
wilke0818: do you have a GPU? On GPU/running a large batch of varied audios, I averaged about 25 seconds per audio. This isn't bad but will be hard to scale with, especially since we haven't been able to show performance improvements by splitting across multiple GPUs

fabiocat93: I have no GPU on my macbook and I request CPU as a preferred device

wilke0818: then to be honest I have no idea how it ran so fast since @900miles and I have both used the code and found it can take up to 15 minutes.

fabiocat93: @wilke0818 sorry, 20 seconds was bark-small. mars5tts takes 8 seconds for the initialization and 4 minutes and 36 seconds for synthesizing "Hello, world. It's nice to meet you." on the CPU

900miles: Hmm.. Perhaps a warning that it might take a while? So users don't think something is broken after waiting 5–15 minutes.

900miles: I'm not sure why yours is so much faster. I ran it on openmind, I don't know what kind of cpus they have

wilke0818: I wouldn't be surprised if newer Apple's (what generation is yours @fabiocat93) have some dedicated GPU(like) hardware that speeds it up

fabiocat93: M2

wilke0818: yeah I would say the M2 simply just is a higher performance chip compared to most Intel chips. It would be interesting to configure actual tests/benchmarks for this and other tasks.

wilke0818:
The TypeGuard should help address the issue that @fabiocat93's zip strategy was working around. Basically MyPy has no real ability to allow for checking Iterables, especially when we allow them to have mixed types, so this is a simple workaround for it. Some other notes:
- I changed `target` to `targets` because it was bothering me
- I made `targets` into `Optional[List[Audio | Tuple[Audio, str]]]` since it seems to me like we have a few different cases, a) no targets provided so Optional, b) targets are all just Audios, c) targets need to be strictly Audios and their transcripts like with Mars5-TTS, and d) if we ever get StyleTTS integrated, the Audio is required with optional transcripts so rather than forcing that choice on the user it seems better to just have internal logic choose which to use
- Technically `targets` needs to be the near last parameter to `synthesize_texts` or else it isn't backwards compatible for users who might've been using purely positional arguments instead of keywords.

fabiocat93:
> The TypeGuard should help address the issue that @fabiocat93's zip strategy was working around. Basically MyPy has no real ability to allow for checking Iterables, especially when we allow them to have mixed types, so this is a simple workaround for it. Some other notes:
> 
> * I changed `target` to `targets` because it was bothering me
> * I made `targets` into `Optional[List[Audio | Tuple[Audio, str]]]` since it seems to me like we have a few different cases, a) no targets provided so Optional, b) targets are all just Audios, c) targets need to be strictly Audios and their transcripts like with Mars5-TTS, and d) if we ever get StyleTTS integrated, the Audio is required with optional transcripts so rather than forcing that choice on the user it seems better to just have internal logic choose which to use
> * Technically `targets` needs to be the near last parameter to `synthesize_texts` or else it isn't backwards compatible for users who might've been using purely positional arguments instead of keywords.

thank you @wilke0818. looks good to me. let's wait for @900miles (final) review

900miles (approved):
Looks great! Thanks @fabiocat93 and @wilke0818

PR #144, opened 2024-08-14
Style tts docker

wilke0818:
- Adds StyleTTS support to senselab
- Consider implementing through a Docker container which can then be applied more generally (run through Pydra): #143 
- Code currently exists in this branch for StyleTTS2 through TorchHub but might be deleted when #129 is pulled (StyleTTS2 support will be removed from that PR for the time being)


wilke0818:
Note these issues with Python 3.12 might be a bigger issue regarding Poetry that is theoretically solved with newer Poetry versions but those newer versions re-create the issue of not being able to use iso-639 (that might be the wrong number but you hopefully get the idea).

PR #149, opened 2024-08-16
Dataset update

wilke0818:
- Updates the Dataset documentation to be more up to our current standards
- Adds functionality for maintaining metadata from HuggingFace datasets, the utility of which will be clear as we write more tutorials that involve real data from HuggingFace
- Adds an additional test for checking that this metadata is extracted correctly


fabiocat93 (approved):

fabiocat93:
looks good to me

PR #152, opened 2024-09-06
Implement diaraization error rate

wilke0818:
- Implements basic Diarization error rate as an evaluation for speaker diarization
- TODO: Consider implementing Jacard Error Rate (https://arxiv.org/pdf/1906.07839)
- TODO: currently relies on an existing mapping between the inferred speaker and the ground truth speakers whereas more commonly we would want the system to decide for us on which speaker is which (commonly uses the Hungarian algorithm)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,7 +57,6 @@ umap-learn = "~=0.5"
 scikit-learn = "~=1.5"
 nltk = "~=3.8"
 vocos = "~=0.1"
-pyinterval = "^1.2.0"
 
 [tool.poetry.group.dev]
 optional = true
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -57,6 +57,7 @@ umap-learn = "~=0.5"
 scikit-learn = "~=1.5"
 nltk = "~=3.8"
 vocos = "~=0.1"
+pyinterval = "^1.2.0"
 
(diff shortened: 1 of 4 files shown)


fabiocat93:
hey @wilke0818 , thank you for implementing this. it's very needed. have you consider using [pyannote.metrics](https://pyannote.github.io/pyannote-metrics/tutorial.html) instead of implementing it all on your own? they give you nice functionalities like [the detailed report](https://pyannote.github.io/pyannote-metrics/basics.html#components) and you can also create your own metric (don't think it's needed for now, but it's a good nice to have)

wilke0818:
I think pyannote.metrics can work for us (and stops us from having to add additional dependencies), however I noticed in the pyannote diarization function, we take an Annotation and convert it to `ScriptLines` and so then converting it back to do the evaluation seemed a bit inefficient, but likely no more inefficient than the current implementation.

Looking deeper into their documentation it also seems like they take care of the Hungarian algorithm under the hood and offer functionality to give the mappings back so it would resolve all of the related TODOs (including the Jacard Error Rate)

fabiocat93 (approved):

fabiocat93:
@wilke0818 looks good to me and is very ready to me merged, unless you want to add more metrics

PR #153, opened 2024-09-11
Text: Sentiment and Emotional Analysis

adi611:


pyproject.toml
```diff
@@ -57,6 +56,7 @@ umap-learn = "~=0.5"
 scikit-learn = "~=1.5"
 nltk = "~=3.8"
 vocos = "~=0.1"
+huggingface-hub = {extras = ["cli"], version = "^0.24.6"}
```
fabiocat93: why is `cli` needed?

src/senselab/text/tasks/emotional_analysis/constants.py
```diff
@@ -0,0 +1,33 @@
+"""This module defines the Emotion enum and related functions for emotional analysis."""
+
+from enum import Enum
+
+
+class Emotion(Enum):
```
fabiocat93: I am not sure that pre-defining the emotional classes is a good idea. What if someone wants to use a model that can capture an emotion that is not in your pool?

src/senselab/text/tasks/emotional_analysis/emotional_analysis.py
```diff
@@ -0,0 +1,79 @@
...
+                - scores (Dict[str, float]): Probabilities for each emotion.
+                - dominant_emotion (str): The most likely emotion in the text.
+
+        Raises:
+            ValueError: If the input list is empty or None.
+        """
+        if not pieces_of_text:
+            raise ValueError("Input list is empty or None.")
+
+        if model is None:
```
fabiocat93: why don't you specify the default in the param definition space (L21)?

src/senselab/text/tasks/emotional_analysis/api.py
```diff
@@ -0,0 +1,33 @@
...
+
+from senselab.text.tasks.emotional_analysis.emotional_analysis import EmotionalAnalysis
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.data_structures.model import HFModel
+
+
+def analyze_emotion(
+    pieces_of_text: List[str],
+    device: Optional[DeviceType] = None,
+    model: Optional[HFModel] = None,
```
fabiocat93: I think model should be of type `SenselabModel`. this way, we can accept models that are not necessarily HF models

src/senselab/text/tasks/emotional_analysis/api.py
```diff
@@ -0,0 +1,33 @@
...
+    max_length: int = 512,
+    overlap: int = 128,
```
fabiocat93: `max_length` and `overlap`  look like model-specific params and shouldn't be in the general emotion recognition API. you can obtain them as part `** kwargs`

src/senselab/text/tasks/emotional_analysis/emotional_analysis.py
```diff
@@ -0,0 +1,79 @@
...
+        )
+
+        tokenizer = cls._get_tokenizer(model, "emotion-analysis")
+        model_instance = cls._load_model(model, "emotion-analysis", device, torch_dtype)
+
+        # Get emotion labels from the model's config
+        emotion_labels = model_instance.config.id2label
+
+        results: List[Dict[str, Any]] = []
+        for text in pieces_of_text:
```
fabiocat93: what is the advantage of using the tokenizer and the model instead of working with the huggingface text classification pipeline? https://huggingface.co/docs/transformers/main_classes/pipelines#transformers.TextClassificationPipeline

adi611: I was mainly using the model directly for processing long texts, but can do the same with a pipeline:

Text → text chunks → pass chunks to the pipeline → average the scores for each label.  
Text → tokenized chunks → pass chunks to the model → softmax for probabilities → average the scores for each label.

I'll switch to using the pipeline for simplicity, will still need the tokenizer for chunking.

pyproject.toml
adi611: not needed, planned to remove this in the final PR

src/senselab/text/tasks/emotional_analysis/constants.py
adi611: you're right, will make changes to the test cases as well.

adi611:
Hi @fabiocat93 - please review. I've made the requested changes and tried to make the code as extensible as possible. If a new model type needs to be added in the future, we won’t need to modify the existing code - just create utils for it and add it to the `MODEL_TYPE_TO_UTILS` dict. These utility classes could also be useful elsewhere in the code.

src/senselab/text/tasks/emotion_analysis/emotion_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+
+        tokenizer = model_utils.get_tokenizer(task="text-classification")
+        pipe = model_utils.get_pipeline(task="text-classification", device=device, torch_dtype=torch_dtype, top_k=None)
+
+        results: List[Dict[str, Any]] = []
+
+        for text in input_data:
+            cls.validate_input(text)
+
+            chunks = chunk_text(text=text, tokenizer=tokenizer, max_length=max_length, overlap=overlap)
```
fabiocat93, line 63: From what i undertsand, pipeline internally implements chunk batching (https://huggingface.co/docs/transformers/main_classes/pipelines#pipeline-chunk-batching). can you please clarify to me why we can't use that directly?

src/senselab/text/tasks/emotion_analysis/api.py
```diff
@@ -0,0 +1,36 @@
...
+        pieces_of_text (List[str]): List of text strings to analyze.
+        model (SenselabModel): The model to use for emotional analysis.
+        device (Optional[DeviceType]): The device to use for computation.
+        **kwargs (Any): Additional keyword arguments.
+
+    Returns:
+        List[Dict[str, Any]]: A list of dictionaries containing emotional analysis results.
+    """
+    model_type = type(model)
+    model_utils = MODEL_TYPE_TO_UTILS.get(model_type)
```
fabiocat93, line 29: If you use a pipeline, you can simply pass the model's name and the revision. What's the advantage here of using the model + tokenizer?

src/senselab/utils/interfaces/analyses.py
```diff
@@ -0,0 +1,68 @@
...
+
+import string
+from abc import ABC, abstractmethod
+from typing import Any, Dict, List, Optional
+
+from senselab.utils.data_structures.device import DeviceType
+from senselab.utils.model_utils import BaseModelSourceUtils
+
+
+class BaseAnalysis(ABC):
```
fabiocat93, line 11: this is an interesting interface. some thoughts:
- I am curious to see how it generalizes on other tasks
- it may be good to include a validate_output method as well

src/senselab/text/tasks/sentiment_analysis/sentiment_analysis.py
```diff
@@ -0,0 +1,91 @@
...
+        max_length = int(kwargs.get("max_length", 512))
+        overlap = int(kwargs.get("overlap", 128))
+        neutral_threshold = float(kwargs.get("neutral_threshold", 0.05))
+
+        device, torch_dtype = _select_device_and_dtype(
+            user_preference=device, compatible_devices=[DeviceType.CUDA, DeviceType.CPU]
+        )
+
+        tokenizer = model_utils.get_tokenizer(task="sentiment-analysis")
+        pipe = model_utils.get_pipeline(task="sentiment-analysis", device=device, torch_dtype=torch_dtype, top_k=None)
```
fabiocat93, line 55: is there an advantage in handling SentimentAnalysis with a different pipeline than "text-classification"?

src/senselab/text/tasks/emotion_analysis/constants.py
```diff
@@ -0,0 +1,43 @@
+"""This module defines the Emotion enum and related constants for emotional analysis."""
+
+from enum import Enum
+from typing import Any, Dict
+
+from senselab.utils.data_structures.model import HFModel
+from senselab.utils.model_utils import HFUtils
+
+
+class Emotion(Enum):
```
fabiocat93, line 10: do we really need this enum?

src/senselab/text/tasks/emotion_analysis/emotion_analysis.py
```diff
@@ -0,0 +1,88 @@
...
+from collections import defaultdict
+from typing import Any, Dict, List, Optional
+
+from senselab.utils.data_structures.device import DeviceType, _select_device_and_dtype
+from senselab.utils.interfaces.analyses import BaseTextAnalysis
+from senselab.utils.model_utils import BaseModelSourceUtils
+from senselab.utils.tasks.chunking import chunk_text
+
+
+class EmotionAnalysis(BaseTextAnalysis):
```
fabiocat93, line 12: clarify that this class is for emotion analysis with HF

src/senselab/utils/model_utils.py
```diff
@@ -0,0 +1,160 @@
+"""This module provides utility classes for handling common utilities based on model type."""
```
fabiocat93, line 1: this is HF-specific. not every framework is structured with tokenizers, feature_extractor, models, and pipelines. I still wonder if we really need all this or if it's just an over-complication. please, help me understand your choice

src/senselab/utils/tasks/chunking.py
```diff
@@ -0,0 +1,25 @@
+"""This module provides functionality for chunking text into overlapping segments."""
+
+from typing import List
+
+from transformers import AutoTokenizer
+
+
+def chunk_text(text: str, tokenizer: AutoTokenizer, max_length: int, overlap: int) -> List[str]:
```
fabiocat93, line 8: 1. this is a text-specific utility and not a general utility.
2. I think chunking may be already managed inside of the text pipeline (at least for HF)

fabiocat93:
> Hi @fabiocat93 - please review. I've made the requested changes and tried to make the code as extensible as possible. If a new model type needs to be added in the future, we won’t need to modify the existing code - just create utils for it and add it to the `MODEL_TYPE_TO_UTILS` dict. These utility classes could also be useful elsewhere in the code.

Thank you @adi611 . i have left some comments and questions. please, feel free to address them to help me understand better your way of thinking.

adi611:
> > Hi @fabiocat93 - please review. I've made the requested changes and tried to make the code as extensible as possible. If a new model type needs to be added in the future, we won’t need to modify the existing code - just create utils for it and add it to the `MODEL_TYPE_TO_UTILS` dict. These utility classes could also be useful elsewhere in the code.
> 
> Thank you @adi611 . i have left some comments and questions. please, feel free to address them to help me understand better your way of thinking.

Hi, I was thinking we could have one `EmotionAnalysis` and one `SentimentAnalysis` class to handle all model types, and use `MODEL_TYPE_TO_UTILS` to get the utils for a model type. The `BaseModelSourceUtils` could have something like a `get_classifier` abstract method, which the concrete utils classes would implement to get the right classifier for each model type. So we wouldn’t have to update existing code if support for more model types are added, just create the new utils class and add it to `MODEL_TYPE_TO_UTILS`. But I get your point, it might oversimplify things across models and not always be practical. I should’ve written up a design doc first, gotten your feedback, and then moved ahead with the code. I’ll make sure to do that going forward to keep things smoother!

src/senselab/text/tasks/sentiment_analysis/sentiment_analysis.py
adi611: not really an advantage, it actually uses the same `TextClassificationPipeline` as `text-classification`. but I’ve noticed in the docs that `sentiment-analysis` is often mentioned as the task when building any sentiment analysis classifier. I guess that’s just to explicitly define the type of text classification. functionality-wise, it’s not really different.

src/senselab/text/tasks/emotion_analysis/constants.py
adi611: using this in the tests to not hardcode any strings, but not really used in the actual classification code.

src/senselab/text/tasks/emotion_analysis/emotion_analysis.py
adi611: The `ChunkPipeline` is meant only for `zero-shot-classification` and `question-answering` tasks, not for `text-classification`. It throws an error if the text goes over the max token limit. I tried it with a few BERT-based models that have a 512 token limit, and it fails when the length exceeds that. But please let me know if I am missing something here.

src/senselab/text/tasks/emotion_analysis/api.py
adi611: Right, I am using the tokenizer to handle chunking for long texts and not to actually create the classifier from the `pipeline`. I haven’t come across a robust method for handling long texts in text classification tasks beyond this. From what I have tested and read on the HF docs, the chunk batching pipeline is only supported for `zero-shot-classification` and `question-answering` and not for `text-classification`.

src/senselab/text/tasks/emotion_analysis/emotion_analysis.py
adi611: Yes, will be making the change. My initial thought was to use the same class for every model type, including HF, and have the utils classes return the classifier with a `get_classifier` method. But you’re right, different model types might have different output formats and my approach would fail then. I should’ve shared an LLD doc first to clear this up.

src/senselab/utils/model_utils.py
adi611: Thank you for your feedback. The plan was to have functions like `get_classifier` (`get_pipeline` now) and `get_tokenizer` to create a standardized way to access these components across different utils, making it easier to work with various models. You’d just call `model_utils.get_classifier()` to get the classifier and `model_utils.get_tokenizer()` to fetch the tokenizer for the model. If tokenizer is not available for a model and a long text is passed that exceeds the token limit, we could show an error. But I understand your concern about complexity and can rework the entire thing to work as it does currently with if-else checks at the `api.py` level.

adi611:
## Proposed Changes

### 1. Base Analysis Classes
- Add `validate_output` method to `BaseAnalysis` and `BaseTextAnalysis`.
- Move `BaseTextAnalysis` to something like `senselab/text/utils/interfaces.py`.

### 2. Chunking Strategy
- If custom chunking needed, move to `senselab/text/utils/`.

### 3. Restructure Utilities
- Remove `BaseModelSourceUtils`. `HFUtils` will not be a derived class and can still be used across the codebase to avoid repeating similar pipeline code.

### 4. Decouple Analysis Logic from Model Type
```python
def analyze_emotion(...):
    if isinstance(model, HFModel):
        return HFEmotionAnalysis.analyze(input_data=pieces_of_text, device=device, **kwargs)
    else:
        raise NotImplementedError(f"The specified model type is not supported.")
```

### 5. Emotion Enum
- Remove `Emotion` enum and use string values in the tests.
- OR, move to test file since only used there.


Please review and provide feedback. cc: @fabiocat93

adi611:
Hi @fabiocat93, let me know if there's another task I can start working on in the meantime.

satra:
Closing — text sentiment now implemented in the SER tutorial (PR #468) using cardiffnlp/twitter-roberta-base-sentiment-latest.

PR #163, opened 2024-09-18
GitHub wf on self-hosted runner

fabiocat93:


fabiocat93:
Changes:
- re-organized unit tests adding SKIP based on GPU (cuda) availability
- integrated self-hosted runner for testing with GPU
- temporarily removed support for python 3.12 (following harsha's suggestion - apparently there is an existing issue with python 3.12 and pytorch)
- removed tests on macos with x64 architecture
- parallelized pytests on multiple cpus
- removed testmon
- removed pytest from pre-commit hooks
- temporarily commented out the transcribe_timestamp workflow because it fail all tests on the GPU (@ibevers can you please double-check that?)
- fixed an embedding_extraction bug on GPU
- enabled caching and authentication when checking whether a github exists
- fixed a plotting bug
- fixed tutorial naming

.github/workflows/e2c-runner-tests-310.yaml
wilke0818, line 1: should we go ahead and add a 3.11 and 3.12 before merging?

wilke0818 (approved):

.github/workflows/e2c-runner-tests-310.yaml
fabiocat93: I have added 3.11. would stay like this for now. We can add 3.12 in the future (I want first to check how much money we use with 3.10 and 3.11).

fabiocat93:
@hvgazula fyi, all set with the self-hosted runner implementation. Thank you again so much for the tutorial and all the brainstorming!!!

hvgazula:
btw, what was the solution for the path-related issue you reached out to me about?

fabiocat93:
> btw, what was the solution for the path-related issue you reached out to me about?

inspired by your recommendation, I copied the workspace folder to a different volume and exported a couple of env variables for `pytest` and `poetry`

hvgazula:
I see. 👍

PR #164, opened 2024-09-25
switch to poetry - last

fabiocat93:


wilke0818 (approved):
I assume this means the new version of poetry fixes the error we were previously seeing with ISO-639?

PR #165, opened 2024-09-25
adjusting windowing by returning an Audio instead of a waveform

fabiocat93:


src/senselab/audio/data_structures/audio.py
```diff
@@ -103,28 +103,53 @@ def __eq__(self, other: object) -> bool:
...
 
-        while current_position < num_samples - window_size:
-            window = self.waveform[:, current_position : current_position + window_size]
-            yield window
+        while current_position < num_samples:
+            print("current_position:", current_position)
+            # Calculate the end position of the window
+            end_position = current_position + window_size
+
+            # If the end_position exceeds the number of samples, take the remaining samples
```
ibevers, line 134: This is not necessary since it is done automatically when slicing tensors

src/senselab/audio/data_structures/audio.py
```diff
@@ -103,28 +103,53 @@ def __eq__(self, other: object) -> bool:
...
+            if end_position > num_samples:
+                end_position = num_samples
+            print("end_position:", end_position)
```
ibevers: This could get verbose. Also, I'm not sure if we want print statements in source code

src/senselab/audio/data_structures/audio.py
```diff
@@ -103,28 +103,53 @@ def __eq__(self, other: object) -> bool:
...
 
         num_samples = self.waveform.size(-1)
+        print("num_samples:", num_samples)
         current_position = 0
 
-        while current_position < num_samples - window_size:
-            window = self.waveform[:, current_position : current_position + window_size]
-            yield window
+        while current_position < num_samples:
+            print("current_position:", current_position)
```
ibevers: This could get verbose. Also, I'm not sure if we want print statements in source code

ibevers (approved):

src/senselab/audio/data_structures/audio.py
fabiocat93: oh sorry, that was there as part of my debugging

PR #174, opened 2024-10-18
Added CHAT file reader

wilke0818:
- Reads and converts .cha files to lists of script lines for use in SenseLab


fabiocat93 (approved):
Looks good. Feel free to add unit tests as needed

wilke0818:
we'd need a .cha file for that. not sure what a minimal file might look like. also this currently won't work as I didn't add `pylangacq` to the requirements because as discussed offline, my poetry is currently broken.

fabiocat93:
> we'd need a .cha file for that. not sure what a minimal file might look like. also this currently won't work as I didn't add `pylangacq` to the requirements because as discussed offline, my poetry is currently broken.


- feel free to create a basic .cha file starting from some you find on CHILDES
- I would recommend cloning the repo from scratch. I have tested the installation today and it works for me

fabiocat93:
@wilke0818 have you ever found what was the source of the problem? (and maybe a solution to it)

wilke0818:
I thought you found that the solution was to not use relative paths?

fabiocat93:
> I thought you found that the solution was not to use relative paths?

If this is completed, I will merge it then!

Issue #140, opened 2024-08-14
Task: The human phenotype vector


satra:
### Description

The idea is to start defining and populating different aspects of the human phenotype estimated from a voice recording.

### Tasks

- [ ] Define buckets (e.g. physiology, voice quality, rate, emotion, ...)
- [ ] Define features in each bucket
- [ ] Prioritize features
- [ ] Implement features 
- [ ] Validate features

### Freeform Notes

ping @900miles and @wilke0818

wilke0818:
It is unclear to me what this even means. perhaps we can discuss at a later point.

fabiocat93:
> It is unclear to me what this even means. perhaps we can discuss at a later point.

@wilke0818 this should be one of the multiple project options we have discussed. I believe the goal here is to establish adn implement a set of descriptors for voice and speech. These descriptors should cover various levels of abstraction and interpretability. For example, they might include low-level acoustic features as well as high-level info like emotions or transcripts. The main motivation is that we want to shift from a class-based description of voice to a dimensional space that captures the complexity of voice and speech.

- input: audio file
- output: a dictionary (or maybe an array given the title of the issue) describing the speaker + their behaviors in the clip

fabiocat93:
Should we see the audio summary (in issue https://github.com/orgs/sensein/projects/45/views/4?pane=issue&itemId=87312536&issue=sensein%7Csenselab%7C139) as a sub-part of the speech phenotype vector? These two may be better discussed and brainstormed together

satra:
i think of #139 as context or metadata for this vector, rather than a sub part necessarily. could be included but the idea vector should be one that should work independent of context. but indeed including it is an easier place to start.

PR #175, opened 2024-10-21
Features extraction

fabiocat93:
This is an attempt to structure audio features extraction. This would contribute to [the human phenotype vector](https://github.com/sensein/senselab/issues/140) with some low-level acoustic and audio quality descriptors. 

Follow-up steps include:

- [x] enlarging the set of features
- [x] edit the code also taking into account children's voices (so far, some functions only have different behaviors for adult males and females
- [x] enable params customization, maybe as in (https://github.com/novoic/surfboard)


src/senselab/audio/tasks/features_extraction/api.py
```diff
@@ -0,0 +1,85 @@
...
+
+    with pydra.Submitter(plugin=plugin) as sub:
+        sub(wf)
+
+    outputs = wf.result()
+
+    formatted_output = []
+    for output in outputs:
+        formatted_output_item = {
+            "opensmile": output.output.opensmile_out[0],
```
wilke0818: do u think it is better to use Dictionaries or create a class for features that we can add to? I worry that the features we're extracting are only going to grow. Additionally from a usability perspective, it is much harder to know the proper output names for a dictionary compared to using a class of some sorts unless we thoroughly document the feature extraction dictionary

wilke0818 (approved):

src/senselab/audio/tasks/features_extraction/api.py
fabiocat93: This is a good point. I am happy to hear from the others as well. I have no strong preference. The doc should be well done anyway. If we go with a class, we should remember to include a method `.to_dict()` because I feel dictionaries may be easier to handle for some people/in some settings

ibevers:
@fabiocat93 is this going to incorporate Nick's features? Am I remembering correctly that he gave us the okay to incorporate them?

fabiocat93:
> @fabiocat93 is this going to incorporate Nick's features? Am I remembering correctly that he gave us the okay to incorporate them?

there already

ibevers:
Cool😎

src/senselab/audio/tasks/features_extraction/api.py
wilke0818: we can worry about this later

fabiocat93:
@satra Quick update: I ... 
- created a general api for audio feats extraction, which returns all the features we implemented using torchaudio, praat-parselmouth, torchaudio-squim and opensmile. The API follows in principle the API of [surfboard](https://github.com/novoic/surfboard), as @danielmlow suggested, but includes multiple improvements (including caching and parallelization params and library-dependent feats extraction implementations). My work should contribute to the human phenotype vector efforts (https://github.com/sensein/senselab/issues/140), which may be @wilke0818 and @900miles are still considering contributing to (?). 
- further restructured the code, organized functions into importable scripts, and documented them
- integrated @nickcummins41's code for feats extraction with praat parselmouth. some feats were already integrated in senselab. Hence I tried to pick the best parts from the two implementations. 
- optimized @nickcummins41 code using pydra plus some minor adjustments, and it is now almost 4 times more performant than the original, tested on my MacBook Pro with 12 CPUs. 
- added a workflow that contains the general interface for health measurement extraction. @nickcummins41 I think we need a name for our feats set. This would make it easily recognizable by the community
- fixed a minor bug that caused the code to crash when no pauses were detected in the speech.
- added some unit tests (we may want to discuss further good ways to test feats extraction functions)
- added caching (good for when you run your code a second time)

fabiocat93:
@nickcummins41 I’ve added your name throughout the docstrings and general documentation to acknowledge your contributions to the audio feature extraction work. If you have any suggestions for more effective ways to give credit or improvements to the documentation, please let me know!

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1380 @@
...
+
+        # remove outliers from wide pitch search
+        pitch_values = pitch_wide.selected_array["frequency"]
+        pitch_values = pitch_values[pitch_values != 0]
+        pitch_values_Z = (pitch_values - np.mean(pitch_values)) / np.std(pitch_values)
+        pitch_values_filtered = pitch_values[abs(pitch_values_Z) <= 2]
+
+        mean_pitch = np.mean(pitch_values_filtered)
+
+        # TODO: what about children?
```
fabiocat93: @satra @nickcummins41, for now the audio feats extraction code only takes into account adult males and females. how about including children in the loop? i have commented with some suggested floor and ceiling values for pitch extraction. please, let me know what you think. also, let me know if you think there are more functions that need to be adapted to children's voices.

satra: for pitch extraction i think female voice range would cover children generally. so it may depend on what aspect of a female range is being considered. also singing voice may yet be a different consideration.

satra:
thanks @fabiocat93  this is great for now. let's get this out and get some feedback on usage. i'm fighting a few fires so won't have time to do an in depth review. perhaps we can do that together sometime in a month or so across the entire codebase.

satra (commented):
based on the last commit, i'm curious how running in series is faster than running in parallel at scale. doesn't mean pydra has to be used, but the change off hand doesn't make sense without some explanation on bottlenecks (memory, gpu, overhead, etc.,.). 

also it looks like this focuses on returning a single feature value per feature independent of the length of the audio. i.e. some features will not make sense over some duration, and some for less than some duration. we could say that should be left up to the user, in which case an example of splitting the audio into chunks would be good. 

(relatedly, for speech like audio, do we have some utility or example for splitting audio based on other targets (e.g. sentences, vad, etc.,.), rather than time?)

src/senselab/audio/tasks/features_extraction/opensmile.py
```diff
@@ -60,11 +63,16 @@ def _extract_feats_from_audio(sample: Audio, smile: opensmile.Smile) -> Dict[str
...
+        try:
+            sample_features = smile.process_signal(audio_array, sampling_rate)
+            # Convert to a dictionary with float values and return it
+            return {
+                k: v[0] if isinstance(v, list) and len(v) == 1 else v
+                for k, v in sample_features.to_dict("list").items()
+            }
+        except Exception as e:
+            print(f"Error processing sample {sample.orig_path_or_id}: {e}")
+            return {feature: torch.nan for feature in smile.feature_names}
```
satra: is this equivalent to `np.nan` or are we saying that torch is a key dependency of senselab (which is fine), and we are using torch constructs elsewhere.

src/senselab/audio/tasks/features_extraction/opensmile.py
```diff
@@ -60,11 +63,16 @@ def _extract_feats_from_audio(sample: Audio, smile: opensmile.Smile) -> Dict[str
...
 
     smile = OpenSmileFeatureExtractorFactory.get_opensmile_extractor(feature_set, feature_level)
     features = [_extract_feats_from_audio(audio, smile) for audio in audios]
```
satra, line 78: parallel support?

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+        audio (Union[Path, Audio]): A path to an audio file or an Audio object.
+        sampling_rate (int, optional): The sampling rate of the audio. Defaults to 16000.
 
-        waveform = audio.waveform
-        sampling_rate = audio.sampling_rate
-        sound = parselmouth.Sound(waveform, sampling_frequency=sampling_rate)
-        duration = _get_duration(sound)
-        return {"duration": duration}
+    Returns:
+        parselmouth.Sound: A Parselmouth Sound object.
```
satra, line 28: should we try to harmonize sound/audio objects?

fabiocat93:
> based on the last commit, i'm curious how running in series is faster than running in parallel at scale. doesn't mean pydra has to be used, but the change off hand doesn't make sense without some explanation on bottlenecks (memory, gpu, overhead, etc.,.).

pydra is still used under the hood. and I confirm that this newer implementation runs faster than previous one.

> also it looks like this focuses on returning a single feature value per feature independent of the length of the audio. i.e. some features will not make sense over some duration, and some for less than some duration. we could say that should be left up to the user, in which case an example of splitting the audio into chunks would be good.

thank you for pointing this out. I agree with that. The way senselab is designed rn, we have some functionalities at the "tasks" level offering all kinds of customization and this is left to the user. the "workflows" level is one level higher in terms of abstraction and includes (/will include) all the best practices/work arounds/heuristics that you are referring to. 

> (relatedly, for speech like audio, do we have some utility or example for splitting audio based on other targets (e.g. sentences, vad, etc.,.), rather than time?)

we do, in the relative sections (vad, time-stamped transcripts, ...)

src/senselab/audio/tasks/features_extraction/opensmile.py
fabiocat93: torch is definitely a key dependency of senselab. Starting from the Audio class, where waveforms are treated as torch tensors

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
fabiocat93: this implementation is because parselmouth.Sound is not pickable, which is a problem with pydra

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+    """Extract speech timing and pausing features from a given sound object.
+
+    Args:
+        snd (Union[parselmouth.Sound, Path, Audio]): A Parselmouth Sound object or a file path or an Audio object.
+
+    Returns:
+        Dict[str, float]: A dictionary containing the following features:
+
+            - speaking_rate (float): Number of syllables divided by duration.
+            - articulation_rate (float): Number of syllables divided by phonation time.
```
satra, line 66: we should clarify what `syllable` means in the description as an acoustically defined syllable. this is an area of praat that means something quite specific based on acoustics and can miss several acoustic phonetic/linguistic considerations (e.g. long syllables, stress, etc.,.). it would be good to know if that has changed. also would this apply to whispered speech? i.e. would it return the same number of syllables between whispered and non-whispered speech of the same sentence?

satra: also pinging @nickcummins41 here.

fabiocat93: > we should clarify what syllable means in the description as an acoustically defined syllable. this is an area of praat that means something quite specific based on acoustics and can miss several acoustic phonetic/linguistic considerations (e.g. long syllables, stress, etc.,.). 

Good point, I can add that. The docs will definitely be refined once the paper we are writing is closer to completion. 


> also would this apply to whispered speech? i.e. would it return the same number of syllables between whispered and non-whispered speech of the same sentence?

Good question. It sounds like an interesting sub-project

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+        dict: A dictionary containing the following keys:
+
+            - f1_mean (float): Mean F1 location.
+            - f1_std (float): Standard deviation of F1 location.
+            - b1_mean (float): Mean F1 bandwidth.
+            - b1_std (float): Standard deviation of F1 bandwidth.
+            - f2_mean (float): Mean F2 location.
+            - f2_std (float): Standard deviation of F2 location.
+            - b2_mean (float): Mean F2 bandwidth.
+            - b2_std (float): Standard deviation of F2 bandwidth.
```
satra, line 760: is it only f1 and f2? if so, perhaps rename function to `measure_f1f2_formants_bandwidths`

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+    """Extract Formant Frequency Features.
+
+    Function to extract formant frequency features from a given sound object. This function is adapted from default
+    Praat code to work with Parselmouth.
+
+    Args:
+        snd (Union[parselmouth.Sound, Path, Audio]): A Parselmouth Sound object or a file path or an Audio object.
+        floor (float): Minimum expected pitch value, set using value found in `pitch_values` function.
+        ceiling (float): Maximum expected pitch value, set using value found in `pitch_values` function.
+        frame_shift (float): Time rate at which to extract a new pitch value, typically set to 5 ms.
```
satra, line 748: is there an option to return the frame by frame values, or should that be done at a meta level (fine from my side)? if the latter, it may be helpful to point to the place in senselab if folks wanted to extract these at a per frame level.

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+        return {"mean_cpp": float("nan"), "std_dev_cpp": float("nan")}
+
+
+def measure_formants(
+    snd: Union[parselmouth.Sound, Path, Audio], floor: float, ceiling: float, frame_shift: float
+) -> Dict[str, float]:
+    """Extract Formant Frequency Features.
+
+    Function to extract formant frequency features from a given sound object. This function is adapted from default
+    Praat code to work with Parselmouth.
```
satra, line 742: may want to add the description of pulses here to indicate that the extraction happens per pitch period.

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
+        output_connections.append(("cpp_out", wf.extract_cpp_descriptors_pt.lzout.out))
+    if duration:
+        output_connections.append(("audio_duration", wf.extract_audio_duration_pt.lzout.out))
+    if jitter:
+        output_connections.append(("jitter_out", wf.extract_jitter_pt.lzout.out))
+    if shimmer:
+        output_connections.append(("shimmer_out", wf.extract_shimmer_pt.lzout.out))
+    wf.set_output(output_connections)
+
+    with pydra.Submitter(plugin=plugin) as sub:
```
satra: wherever we use this, we should also have `plugin_args`. in some settings leaving it to the default number of cores causes slowdowns if other tasks are using the same cores.

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
```diff
@@ -1,143 +1,1377 @@
...
-            "local_jitter": _get_jitter("local", point_process),
-            "localabsolute_jitter": _get_jitter("local, absolute", point_process),
-            "rap_jitter": _get_jitter("rap", point_process),
-            "ppq5_jitter": _get_jitter("ppq5", point_process),
-            "ddp_jitter": _get_jitter("ddp", point_process),
+            "local_jitter": np.nan,
+            "localabsolute_jitter": np.nan,
+            "rap_jitter": np.nan,
+            "ppq5_jitter": np.nan,
+            "ddp_jitter": np.nan,
```
satra, line 1044: should these be `torch.nan`

src/senselab/audio/tasks/features_extraction/torchaudio.py
```diff
@@ -36,7 +37,10 @@ def extract_spectrogram_from_audios(
     )
     spectrograms = []
     for audio in audios:
```
satra, line 40: we should perhaps discuss how to parallelize such loops for large datasets/long recordings in general. let's make this a key part of our code review target.

src/senselab/audio/tasks/features_extraction/torchaudio_squim.py
```diff
@@ -10,64 +11,78 @@
...
 
-    for audio in audio_list:
-        stoi, pesq, si_sdr = objective_model(audio.waveform)
-        features["stoi"].append(stoi.item())
-        features["pesq"].append(pesq.item())
-        features["si_sdr"].append(si_sdr.item())
+    for audio in audios:
+        audio_features = {}
+        try:
+            stoi, pesq, si_sdr = objective_model(audio.waveform)
```
satra, line 41: perhaps add this to description.

```
Wideband Perceptual Estimation of Speech Quality (PESQ)
Short-Time Objective Intelligibility (STOI)
Scale-Invariant Signal-to-Distortion Ratio (SI-SDR)
```

src/senselab/audio/tasks/features_extraction/torchaudio_squim.py
```diff
@@ -10,64 +11,78 @@
...
+        except RuntimeError:
+            audio_features["mos"] = torch.nan
 
-    for i, audio in enumerate(audio_list):
-        mos = subjective_model(audio.waveform, non_matching_references[i].waveform)
-        features["mos"].append(mos.item())
+        features_list.append(audio_features)
 
-    return features
+    return features_list
```
satra: this seems semantically counter to the `audios` change. similar to `audios`, if we left this as `features`, would that hurt anything?

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
satra: we should test this directly with `cloudpickle`, which is what pydra uses. any known reasons for this? perhaps something we could address upstream depending on the reason.

src/senselab/audio/workflows/health_measurements/extract_health_measurements.py
```diff
@@ -0,0 +1,93 @@
...
+"""
+
+from typing import Any, Dict, List, Optional
+
+from senselab.audio.data_structures import Audio
+from senselab.audio.tasks.features_extraction.praat_parselmouth import extract_praat_parselmouth_features_from_audios
+
+
+def extract_health_measurements(audios: List[Audio], cache_dir: Optional[str] = None) -> List[Dict[str, Any]]:
+    """Extract health measurements from audio files.
```
satra, line 15: `health` seems like a broad space, but perhaps this is putting a stake in the ground here :)

while many of these features have been shown to vary with health conditions, it would be good to make this more concrete with pointers and examples. this seems quite related to the phenotype vector considerations at a conceptual level. perhaps, to make this a key output of this project, we should raise the standards and have some principled way in which we add features or composite scores that have direct evidence and relevance to health. for example such principles could include high out of sample positive predictive value (precision) for a disorder/symptom tested in a large diverse dataset. i'm thinking what level of evidence would one need to include such a feature be included alongside someone's EHR.

we could also have a softer version that says these features have some evidence but do not rise to a standard that should be included in an EHR (like a blood test result) for example. perhaps this is that function and we could make a new function that says `EHR_ready_features`, which would have the higher bar.

satra:
@fabiocat93 - great job here. i finally had some time to go through this. i realize there are some inconsistencies in places, and instead of trying to be perfect, fix the easy ones, let's file some of those others as issues/discussions, and let's focus on optimizations in a different PR. 

i'm hoping to update b2aiprep this weekend. let me know if you think a version could be released.

src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
wilke0818: you might be able to write a hash function for parselmouth.Sound, similar to how we had to write one for Torch.Tensor so that Audios would work. I know that was a different issue but it possibly could be solved this way?

fabiocat93: i remember i spent way too much than I wanted trying to do it, but with no success. same thing with opensmile's smile object, which is why opensmile feats extraction is not parallelized. but I agree with you that it would be great it get it done. would be a great optimization

src/senselab/audio/workflows/health_measurements/extract_health_measurements.py
fabiocat93: this is basically we are discussing in the paper with @nickcummins41. we are aware that this is just a first effort and we will need v2, v3, ... of this set. also, the details of the features inclusion and exclusion criteria will be in the paper coming with the feats. once the manuscript is ready, we can surely summarize some of the info in our doc and point to the paper for more details

PR #182, opened 2024-11-13
added extract_mel_filter_bank_from_spectrograms

### LINKED ISSUE #178: Task: refactor extract_mel_filter_bank_from_audios
### Description

the current extract_mel_filter_bank_from_audios extraction (https://github.com/sensein/senselab/blob/e5a935a808782cf7ca7d7836b5e0fab45858cd1f/src/senselab/audio/tasks/features_extraction/torchaudio.py#L121) extracts spectrograms for all audios. it would be nice to have that as an optional input in case they have been already generated or broken into separate functions.

### Tasks

- [x] refactor mfcc function to take optional input or break into two functions, one from audios and one from spectrograms

### Freeform Notes

more generally consider optimizing internal parallelization so that scale of number of files doesn't can be handled with limited memory. (this may apply to multiple functions).

fabiocat93:
This closes the issue https://github.com/sensein/senselab/issues/178 .


src/senselab/audio/tasks/features_extraction/torchaudio.py
```diff
@@ -168,6 +168,35 @@ def extract_mel_filter_bank_from_audios(
     return mel_filter_banks
 
 
+def extract_mel_filter_bank_from_spectrograms(
+    spectrograms: List[Dict[str, torch.Tensor]],
+    sampling_rate: int,
+    n_mels: int = 128,
+    n_fft: int = 1024,
```
satra: isn't this retrievable from the shape of the spectrogram?

fabiocat93: You are right. I have edited the code accordingly extracting `n_stft` from the spectrogram shape

fabiocat93:
This should be ready for review now

src/senselab/audio/tasks/features_extraction/torchaudio.py
```diff
@@ -168,6 +168,33 @@ def extract_mel_filter_bank_from_audios(
...
+        n_mels (int): Number of mel filter banks. Default is 128.
+
+    Returns:
+        List[Dict[str, torch.Tensor]]: List of Dict objects containing mel filter banks.
+    """
+    mel_filter_banks = []
+    for spectrogram in spectrograms:
+        try:
+            melscale_transform = torchaudio.transforms.MelScale(
+                sample_rate=sampling_rate, n_mels=n_mels, n_stft=spectrogram["spectrogram"].shape[0]
```
satra, line 190: can you check whether this is (2*shape[0] + 1) or this?

fabiocat93: It should be fine the way it is. [torch.transforms.Spectrogram](https://pytorch.org/audio/main/generated/torchaudio.transforms.Spectrogram.html) returns a torch.tensor of size (freq, time), where freq is `n_fft // 2 + 1`, where `n_fft` is the number of Fourier bins, and time is the number of window hops (`n_frame`). 
[torchaudio.transforms.MelScale](https://pytorch.org/audio/main/generated/torchaudio.transforms.MelScale.html) wants `n_stft` , which is exactly `n_fft // 2 + 1`. hence, `spectrogram["spectrogram"].shape[0]` works as expected in our case.

satra: thanks for looking into this.

satra (approved):

PR #201, opened 2024-11-18
Created SER tutorial

### LINKED ISSUE #197: Task: create an abstract interface for senselab Speech Emotion Recognition and Audio Classification
Create an API for audio classification and emotion recognition. For now, this will only include huggingface models. @wilke0818 has been working on it. This new issue is to give some structure to the repo.

wilke0818:
## Description
Creates a tutorial using Senselab for SER.

## Related Issue(s)
#197 

## Motivation and Context
We were lacking a tutorial using the functionality for audio classification, which currently has one specific implementation, speech emotion recognition, and so now novice users can better understand how to Senselab might be useful for this task.

## How Has This Been Tested?
Through Colab


fabiocat93:
Thanks @wilke0818 for the tutorial! It’s helpful and does a good job of addressing real challenges that users might face. I like it, and I’m curious to see how actual users respond.

A few things are still missing to make this more complete:

1. **Senselab installation**: We should install `senselab` only when running in Colab. If the user is running elsewhere, we can assume they’ve already set it up. You can use a function like this:

```python
def is_colab():
    try:
        import google.colab
        return True
    except ImportError:
        return False

if is_colab():
    %pip install senselab
else:
    print("Not running on Colab. Skipping installation.")
```

2. **API for classification and SER**: Add an API for audio classification and speech emotion recognition (SER). For now, this API should call the Hugging Face-based function only if the model uses Hugging Face. If other types of models are used, raise a `NotImplementedError`. This will make this task more in line with the others in senselab and more easily maintainable. 

3. **Documentation for classification and SER**: Add documentation for these tasks. It should explain what the tasks are and link to the tutorial. You can use resources like this to get started: https://huggingface.co/tasks/audio-classification.

4. **This branch is out-of-date with the base branch**: Please, update the branch before requesting a new review.

Let me know if you need help with any of these points!

fabiocat93 (changes requested):
Thanks @wilke0818 . I have commented some required changes

fabiocat93:
hi, @wilke0818 ! Did you have any time to work on this?

wilke0818:
Nope. Added the tutorial change that you gave (might not have updated). Need to refactor for the API per the other issue on this topic. It is unclear to me what you would want for documentation. The functionalities themselves are documented already and this tutorial provides the information about the task a user might need (it is pretty akin to the link you sent).

fabiocat93:
> It is unclear to me what you would want for documentation. The functionalities themselves are documented already and this tutorial provides the information about the task a user might need (it is pretty akin to the link you sent).

Every task has a documentation page that explains what the task is, how it's commonly evaluated, what are the popular datasets and models. You can see the doc: https://sensein.group/senselab/senselab/audio/tasks/text_to_speech.html

wilke0818:
I mean that makes sense but do we want this to be an SER task or a generic audio classification task (which is what the HuggingFace pipeline is) which doesn't have a specific task/dataset but where SER is just an example usage of the task?

fabiocat93:
> I mean, that makes sense, but do we want this to be an SER task or a generic audio classification task (which is what the HuggingFace pipeline is) that doesn't have a specific task/dataset but where SER is just an example usage of the task?

Following https://github.com/sensein/senselab/issues/197, both. I would implement both a classification task (as HuggingFace has) and a SER task and would make it so that SER exploits the classification interfaces and employs some checks before and after (e.g., outputs should be emotion-related)

wilke0818:
Need to write the documentation page for SER but otherwise this task (and the corresponding issue #197) are finished.

wilke0818:
@fabiocat93 feel free to check and review the changes. Documentation pages have been added. SER tutorial should now be mostly complete (dimensional classification was removed since models for it don't really work with the audio-classification HF pipeline).

src/senselab/audio/tasks/classification/speech_emotion_recognition/speech_emotion_recognition.py
fabiocat93, line 1: is this still required? @wilke0818

fabiocat93 (approved):
Looks good to me. @wilke0818: you may want to remove `src/senselab/audio/tasks/classification/speech_emotion_recognition/speech_emotion_recognition.py` if not required anymore.

src/senselab/audio/tasks/classification/speech_emotion_recognition/speech_emotion_recognition.py
fabiocat93: Hey @wilke0818 , pinging you on this. I only need a yes/no. I can handle it from here. Thanks!

wilke0818: Yes it can be removed. I thought there might be one other requested change I needed to go back and do but if not then feel free.

PR #203, opened 2024-11-19
Media Pipe Pose Estimation + Visualization

### LINKED ISSUE #173: Task [pose estimation]: implement the general pose estimation API and utilities, plus integrate some initial models (e.g., mediapipe, DeepLabCut)
# Senselab Pose Estimation

**Goal:**  
Integrate robust pose estimation workflows within Senselab

[ViTPose performs best on infants - mediapipe and deeplabcut are low on accuracy](https://arxiv.org/pdf/2406.17382)

---

### Support Plan
- **Start with [MediaPipe](https://ai.google.dev/edge/mediapipe/solutions/vision/pose_landmarker)**  
  - Easy to implement, lightweight (in terms of computation and speed)
  
- **Follow up with:**
  - [**OpenPose**](https://cmu-perceptual-computing-lab.github.io/openpose/web/html/doc/md_doc_03_python_api.html): Best performance across humans and animals (depending on the specific model): Better performance than MediaPipe, but more computationally intensive
  - [**ViTPose**](https://github.com/ViTAE-Transformer/ViTPose?tab=readme-ov-file): Best performance across humans and animals (depending on the specific model)
  - [**DeepLabCut**](https://github.com/ViTAE-Transformer/ViTPose?tab=readme-ov-file): High performance on animal pose tracking

---

### Workflow
1. Upload media
2. Receive custom output 
3. Perform further analysis using the output

---

### Version Planning

- **V1:**
  - Create pose estimations for images and videos
    - Custom datatypes for output
  - Support for MediaPipe
  - Visualization support
  
- **V2:**
  - Expand support to OpenPose
  
- **V3:**
  - Real-time pose detection across supported models (?)

---

### Inputs/Outputs

- **Inputs:**  
  - Image/Video file  
  - Configurations (e.g., [MediaPipe example](https://ai.google.dev/edge/mediapipe/solutions/vision/pose_landmarker/python#configuration_options))  
  - Model  

- **Outputs:**

  - **PoseImage**
    - Keypoints representing the pose skeleton
    - Visualization methods

  - **PoseVideo**
    - Keypoints representing the pose skeleton per video frame
    - Visualization methods

---

At each step, proper documentation and tests are expected. Tutorials will also be implemented for each workflow (across models and modalities).

brukew:
## Description
Implemented Media Pipe pose estimation. Given an image path, will return a PoseSkeleton object with landmarks of each individual in the image. 

## Related Issue(s)
https://github.com/orgs/sensein/projects/45/views/3?pane=issue&itemId=82951656&issue=sensein%7Csenselab%7C173

## Motivation and Context
This is the initial structure for pose estimation which is a valuable signal for behavior analysis. I will expand to more models and functionality, and with that, this will be more generalized.

## How Has This Been Tested?
I tested with different kinds of images and attempts to access invalid properties of the PoseSkeleton object. Unit tests for the new functions + also manually tested for proper visualization.

## Types of changes
Created PoseSkeleton object that contains pose information for individuals in an image. Currently supports MediaPipe pose estimation + visualization functionality.  

## Checklist:


fabiocat93:
Thank you, @brukew, for the updates. I have a few suggestions and points to address based on our previous discussions:

1. **Reorganize Code Structure**  
   Please re-organize your code by separating **data structures** from **functionalities** (what we refer to as tasks in Senselab).  
   - The **skeleton** should be treated as a data structure. It should define the skeleton and optionally include utilities for visualization. Alternatively, visualization can be handled as a standalone task.  
   - **Pose estimation** should be a task (generating the skeleton as an output). All human pose estimation models (e.g., MediaPipe, AlphaPose) should conform to a consistent data structure.  
   - To encourage generalizability, I suggest integrating a second pose estimation tool, such as [YOLO](https://docs.ultralytics.com/tasks/pose/) due to its simplicity.

2. **Model Inclusion**  
   The current approach of including a model within the source code (e.g., `src/senselab/video/tasks/pose_estimation/models/pose_landmarker.task`) makes the package unnecessarily heavy. Instead, please ensure models are downloaded as needed. You can take inspiration from [this example](https://colab.research.google.com/github/googlesamples/mediapipe/blob/main/examples/pose_landmarker/python/%5BMediaPipe_Python_Tasks%5D_Pose_Landmarker.ipynb?hl=it).

3. **Documentation**  
   Please add a dedicated documentation page:  
   - Explain **human pose estimation** as a task, its purpose, and supported models.  
   - For instance, you can reference [this documentation](https://github.com/google-ai-edge/mediapipe/blob/master/docs/solutions/pose.md) for MediaPipe.  
   - Feel free to draw inspiration from the existing audio task documentation (though some sections are incomplete).

4. **Tutorial**  
   Create a Jupyter Notebook tutorial to demonstrate:  
   - How to use the interface.  
   - What functionalities are available.  
   Add this under a `video` folder in the `tutorial/` directory.

5. **Failing Tests**  
   I noticed two tests are failing:  
   - `test_valid_image_single_person`  
     - AssertionError: "Input and output image shapes should match."  
   - `test_visualization_single_person`  
     - ValueError: "Input image must contain three-channel BGR data."  
   Please double-check these tests to ensure they pass.

fabiocat93 (changes requested):
@brukew, I have commented some required changes.

brukew:
Nice, thank you for the feedback @fabiocat93. I will address your comments and ask questions as I go.

fabiocat93:
hi @brukew , did you have any time to work further on this?

brukew:
hey @fabiocat93, slowed down on development towards the end of the semester. I will continue this month.

brukew:
@fabiocat93 lmk what you think about having separate estimator classes for different models. also, I kept the landmark names as they are originally listed but wondering if I should homogenize them at all.

fabiocat93:
> @fabiocat93 lmk what you think about having separate estimator classes for different models. also, I kept the landmark names as they are originally listed but wondering if I should homogenize them at all.

Good question. You can have multiple estimators, each being a child of the same abstract class and producing a skeleton following the same universal structure

brukew:
addressed model download by downloading models to video/tasks/pose_estimation/models upon use. Let me know if there is a better alternative to this that is more in line with senselab use.

brukew:
also wondering how I can ensure tutorial works (and cells are run in tutorial) if changes haven't been pushed to main branch @fabiocat93. should I wait until after?

fabiocat93:
> addressed model download by downloading models to video/tasks/pose_estimation/models upon use. Let me know if there is a better alternative to this that is more in line with senselab use.

Good for now. But you raised a good point. We will need to implement a customizable cache folder for the package (of course with a default value)

fabiocat93:
> also wondering how I can ensure tutorial works (and cells are run in tutorial) if changes haven't been pushed to main branch @fabiocat93. should I wait until after?


When testing the tutorial locally, you can install the package directly from a specific branch instead of relying on `pip install senselab`. This allows you to ensure the tutorial runs correctly with the latest changes before they are merged into the main branch

fabiocat93 (commented):
@brukew : great work on this! I have a few suggestions and questions for consideration:

- I noticed the visualization method in the tutorial doesn't actually visualize the image but returns the array representation. Don’t you think it would make more sense for a visualization method (with the word visualize in the name) to display the image directly? Returning the array (and optionally saving it) could still be included as additional functionality.
- Is there a specific reason for having both IndividualPose and ImagePose? Would it make sense to simplify by treating IndividualPose as a special case of ImagePose where there’s only one subject? It might reduce redundancy and improve maintainability.
- I noticed there’s no mapping of the skeletons into a more general senseLab skeleton yet, and the visualization method is method-specific rather than generic for senselab. Was there a challenge in unifying this? Having two different skeletons and implementations feels less maintainable. Is there a particular reason for this approach?

Let me know if you’d like to discuss any of these points. I’m happy to help! Also, let me know if you prefer to fix the minor questions and merge and keep the rest of the changes for a future PR or if you want to keep it more ordered and work on this until everything is ready.

fabiocat93:
A quick note for the entire group about installation: the number of dependencies, especially for video, makes installation quite heavy. Moving forward, we might want to split the package into modules like senselab[audio], senselab[video], and senselab[text]. This would simplify installation based on the user’s needs

brukew:
Hey @fabiocat93, thanks for the comments - I will make all changes before merging. Here are my thoughts:

1. Yes, that makes sense. I will do that.

2. I think it gives us greater power for anything we want to do later. For example, if a video has different people over time, we could combine face analysis and pose estimation to aggregate the poses of an individual across frames which would be easier with IndividualPose object (with tweaks). It's also just intuitive to have a separate object for each individual detected. I could take it out for now though.

3. I can do this. My thinking was that users could have access to whatever the given model output is and they can work with it how they want to, but I figure that defeats the point of having these in Senselab. So I guess I'll make a Senselab keypoint mapping and attempt to not lose any information across the models - so it includes all possible key points across models. What do you think about that?

fabiocat93:
> I think it gives us greater power for anything we want to do later. For example, if a video has different people over time, we could combine face analysis and pose estimation to aggregate the poses of an individual across frames which would be easier with IndividualPose object (with tweaks). It's also just intuitive to have a separate object for each individual detected. I could take it out for now though.

You convinced me with this. but how about always returning back an ImagePose object (which may include as many IndividualPose objects as the number of people detected?)

> I can do this. My thinking was that users could have access to whatever the given model output is and they can work with it how they want to, but I figure that defeats the point of having these in Senselab. So I guess I'll make a Senselab keypoint mapping and attempt to not lose any information across the models - so it includes all possible key points across models. What do you think about that?

Sounds great! And please let me know if you need or want to brainstorm it further before implementing it.

brukew:
@fabiocat93 Addressed the visualization and keypoint mapping comments. The unit tests kept failing on Git Hub when a plot was created - my workaround was adding a bool parameter for plotting but let me know if there is another way that you want me to handle that.

fabiocat93 (approved):
Looks good to me. Nice job @brukew ! Looking forward to seeing more models integrated and some cools analyses based on this!!

PR #256, opened 2025-02-26
253 task implement initial version of bioacoustic quality control

ibevers:
## Description
Uses a bioacoustic task taxonomy tree with unique nodes (`BIOACOUSTIC_TASK_TREE`). Each node has associated checks. When `check_quality` is run, a subtree for an input list of Audio objects is created based on which task metadata labels (maximum of one per audio because of unique nodes) are in the audio files. The nodes in this subtree are recursively iterated over, and the checks at each node are run for audio files that have that node in the path to their associated task in the tree. Exclude and review Audio objects are stored in relevant fields for each node for each check. The filtered Audio objects (including review) are passed to the next node. Users can input a custom task taxonomy.

In separate PRs, I plan to:
- Convert the dataset subtree to a pydra workflow and use a generator for scalability.
- Extend the initial bioacoustic task taxonomy based on specific project needs. 
- Add filtering of checks by importance and computational complexity.
- Add more checks.
- Possibly adapt this method for other data types.
## Motivation and Context

Implements a systematic, flexible, and extensible approach to bioacoustic quality control as discussed in our Biometrics meeting.

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:


src/senselab/audio/tasks/bioacoustic_qc/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+
+__all__ = [
+    "audios_to_task_dict",
+    "check_quality",
+    "task_dict_to_dataset_taxonomy_subtree",
+    "task_to_taxonomy_tree_path",
+    "taxonomy_subtree_to_pydra_workflow",
+    "check_node",
+    "run_taxonomy_subtree_checks_recursively",
+    "check_quality",
```
fabiocat93, line 20: `check_quality` is listed twice

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
```diff
@@ -0,0 +1,222 @@
...
+from copy import deepcopy
+from typing import Any, Dict, List, Optional, Set, Tuple
+
+from pydra.engine.core import Workflow  # Assuming you're using Pydra workflows
+
+from senselab.audio.data_structures import Audio
+from senselab.audio.tasks.bioacoustic_qc.constants import BIOACOUSTIC_TASK_TREE
+
+
+def audios_to_task_dict(audios: List[Audio]) -> Dict[str, List[Audio]]:
```
fabiocat93: I wouldn't use the term `task` for describing speech activity - e.g., breathing, cough, word repetition, ... - since it already means something else in the package

fabiocat93 (commented):
thank you @ibevers . This looks like a great start. One question/comment: I see how if an audio file has no task label, it will still be processed under "bioacoustic". Do you think it could me possible in the future to automatically estimate whether an audio sample contains a task in the provided hierarchical organization? (sounds more like a future work)

src/senselab/audio/tasks/bioacoustic_qc/__init__.py
ibevers: Good catch, thanks!

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
ibevers: @fabiocat93  Good point! What do you think of using `activity`?

ibevers:
> thank you @ibevers . This looks like a great start. One question/comment: I see how if an audio file has no task label, it will still be processed under "bioacoustic". Do you think it could me possible in the future to automatically estimate whether an audio sample contains a task in the provided hierarchical organization? (sounds more like a future work)

Great question--I think it depends. For example, if you have speech recordings, then labeling those automatically as speech is not hard (although not perfect), but if you have highly specific speech tasks that no existing model can classify, and you want separate labels for all of them, then it would be harder. 

However, if the file names or other metadata that comes with the audio files reflects task differences, it could be doable.

My guess is that the effort of trying to automate this in general is not worth it, except perhaps for a few high-level classes, since there are many cases and different forms of contextual information might be available to users to label their Audio objects

fabiocat93:
> > thank you @ibevers . This looks like a great start. One question/comment: I see how if an audio file has no task label, it will still be processed under "bioacoustic". Do you think it could me possible in the future to automatically estimate whether an audio sample contains a task in the provided hierarchical organization? (sounds more like a future work)
> 
> Great question--I think it depends. For example, if you have speech recordings, then labeling those automatically as speech is not hard (although not perfect), but if you have highly specific speech tasks that no existing model can classify, and you want separate labels for all of them, then it would be harder.
> 
> However, if the file names or other metadata that comes with the audio files reflects task differences, it could be doable.
> 
> My guess is that the effort of trying to automate this in general is not worth it, except perhaps for a few high-level classes, since there are many cases and different forms of contextual information might be available to users to label their Audio objects

fair. I agree with you that some automation for speech may be cool, though

ibevers:
> fair. I agree with you that some automation for speech may be cool, though

We could definitely explore this in the future

src/senselab/audio/tasks/bioacoustic_qc/constants.py
```diff
@@ -0,0 +1,85 @@
+"""Taxonomy of bioacoustic tasks emphasizing human tasks."""
+
+from senselab.audio.tasks.bioacoustic_qc.checks import audio_intensity_positive_check, audio_length_positive_check
+
+BIOACOUSTIC_TASK_TREE = {
```
satra: where did this task tree come from?

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
```diff
@@ -0,0 +1,222 @@
...
+        >>> audio1 = Audio(waveform=torch.rand(1, 16000), sampling_rate=16000, metadata={"task": "breathing"})
+        >>> audio2 = Audio(waveform=torch.rand(1, 16000), sampling_rate=16000, metadata={"task": "cough"})
+        >>> audio3 = Audio(waveform=torch.rand(1, 16000), sampling_rate=16000, metadata={})  # No task
+        >>> audios_to_task_dict([audio1, audio2, audio3])
+        {'breathing': [audio1], 'cough': [audio2], 'bioacoustic': [audio3]}
+    """
+    task_dict: Dict[str, List[Audio]] = {}
+
+    for audio in audios:
+        task = audio.metadata.get("task", "bioacoustic")  # Default to "bioacoustic" if no task
```
satra: shouldn't this be `unknown` by default?

satra (commented):
this PR seems to make some strong assumptions about what's in an audio file.  perhaps start with functions for audio quality metrics (categorized into scalar - single value independent of duration, and temporal - values change over time - can be computed over a hopped window), e.g.

- proportion of silence
- silence at beginning
- silence at end
- amplitude headroom
- peak SNR
- percentage clipping
- clipping
- etc.,.

to simplify implementation can we focus on generating a set of AQM's that should simply be metrics not decisions. decisions can be implemented later more along the lines of the flowchart you are trying to get at. however, let's consider the flow later.

it may also be helpful to take a look at the descriptions of QA/QC here: https://www.nipreps.org/qc-book/preamble/qa-qc-basics.html (it's for imaging, but some of the concepts are portable).

ibevers:
@satra 
> PR seems to make some strong assumptions about what's in an audio file

What do you mean? What assumptions?

> * proportion of silence
> * silence at beginning
> * silence at end
> * amplitude headroom
> * peak SNR
> * percentage clipping
> * clipping
> * etc.,.

Will prioritize these in: https://github.com/sensein/senselab/issues/270

> perhaps start with functions for audio quality metrics 

Will implement here: https://github.com/sensein/senselab/issues/270

> can we focus on generating a set of AQM's that should simply be metrics not decisions

Yes.

> however, let's consider the flow later.

I will make the AQM generation at the top level of the tree for now, and we can decide whether to move or add others further down in the future.

> https://www.nipreps.org/qc-book/preamble/qa-qc-basics.html

Thanks, I'll take a look.

src/senselab/audio/tasks/bioacoustic_qc/constants.py
ibevers: I created it.

ibevers: (with help from ChatGPT)

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
ibevers: Since it's bioacoustic_qc and senselab is a biometrics package, I see it as the user's responsibility to only run this qc pipeline on bioacoustic audio files.

satra:
> What do you mean? What assumptions?

it currently assumes a single task (path through the tree) per audio file. i didn't see anything that would suggest otherwise (let me know if i missed something).

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
satra: i would say that it's up to senselab to detect if there is bioacoustic signal in the audio stream rather than saying individuals need to know what it's running on. doesn't have to be MVP, but starting with such a constraint precludes other operations. think scale, usability (from a user perspective), and operations on millions of files potentially from youtube. for example, even in the bridge2ai dataset there are recordings where only the environmental noise is captured. i'm trying to separate intent of recording from the recording itself, which we may not know has any bioacoustic signal.

ibevers:
> > What do you mean? What assumptions?
> 
> it currently assumes a single task (path through the tree) per audio file. i didn't see anything that would suggest otherwise (let me know if i missed something).

@satra It does use a single path through the tree per subset of the audio files. However, this is not an assumption about what's in an audio file. It's an _inference_ about what _should_ be in an audio file from the definition of the task during which it was recorded. For example, if the audio file was recorded in a three quick breaths task, then there should be three breaths in the file, and if there are only two, the audio file should be flagged for review or exclusion.

satra:
> then there should be three breaths in the file, and if there are only two, the audio file should be flagged for review or exclusion.

see my other note. this is perhaps the area where i would want to see this split into protocol-specific components (in b2aiprep) and generic components (for example, if there is a reasonable taxonomy in senselab). so `should be` is a separate consideration from basic elements of quality for a signal in a file. 

hence let's start with measures of information rather than whether that information makes sense. knowing what you expect itself can be a function of information (e.g., is breathing present in this file?). however, that will still likely require functions on the file itself. so separating out computable objects (detect breath/amount of breath) from decisions (is breath present, or two breaths present) will help with separation of concerns.

src/senselab/audio/tasks/bioacoustic_qc/constants.py
satra: let's revisit this after framework for metrics (and basic metrics) are in place.

ibevers: Okay, sounds good. It's not meant to be complete, just a starting point.

src/senselab/audio/tasks/bioacoustic_qc/bioacoustic_qc.py
ibevers: @satra This is a particular case of automated activity labelling, which Fabio suggested. I agree that automated labelling as bioacoustic or not would be useful. 

Planned here:
https://github.com/sensein/senselab/issues/271

For now, let's say this is the user's responsibility since many AQMs (e.g. the ones you listed) and checks don't necessarily apply in every case outside of the scope of bioacoustic files.

ibevers:
@satra 
> see my other note

I'm guessing you're referring to prioritizing AQMs. When I gave the example, I was not indicating in any way that decision checks should be prioritized over AQMs. What I said before about focusing on AQMs in https://github.com/sensein/senselab/issues/270 still applies.

> this is perhaps the area where i would want to see this split into protocol-specific components

It's designed so that a custom protocol-specific (e.g. b2ai) taxonomy can be used. 

> let's start with measures of information rather than whether that information makes sense

Understood.

> separating out computable objects (detect breath/amount of breath) from decisions (is breath present, or two breaths present)

Breath presence or count label is still an AQM. The only decision this approach is designed to make currently is labelling exclude, include, or review. That labelling can be done with deterministic rules for obvious cases (e.g. 0 amplitude audio files), but for more ambiguous cases it may be better for fuzzy rule prioritization to be inferred using Snorkel, as discussed in Slack. https://github.com/sensein/senselab/issues/269 will enable future integration with Snorkel.

ibevers:
@fabiocat93 Are there any remaining barriers to approve this?

ibevers:
@fabiocat93 Remembered I need to change things named 'task' to 'activity'

fabiocat93:
> @fabiocat93 Are there any remaining barriers to approve this?

Hey @ibevers! I’d flip the question back to you: do we gain anything by merging this now, knowing it’s still incomplete? @satra raised some good points and suggested starting with audio quality assessment functions, which sounds like a solid approach to me. It’ll allow you to explore how rules work in practice before deciding on their structure. Otherwise, it’s just a conceptual exercise with no real impact. Leet me know what you think!

ibevers:
@fabiocat93 This is complete within the scope of the PR. It provides an initial functional version of bioacoustic QC. My intent here is to to do smaller PRs and move towards continuous integration in line with this approach:
https://sourcery.ai/blog/small-changes-big-impact/#:~:text=Small%20PRs%20are%20easier%20to,when%20the%20change%20is%20minimal.

I was talking to a senior software engineer recently who says he usually implements a complete functional feature and then rewrites it as a series of linked smaller PRs that he aims to keep under ~100 lines of code to facilitate better code quality, make reviews more bite-sized, and increase integration velocity.

Audio quality metrics will be implemented here: https://github.com/sensein/senselab/issues/270

fabiocat93:
> @fabiocat93 This is complete within the scope of the PR. It provides an initial functional version of bioacoustic QC. My intent here is to to do smaller PRs and move towards continuous integration in line with this approach: https://sourcery.ai/blog/small-changes-big-impact/#:~:text=Small%20PRs%20are%20easier%20to,when%20the%20change%20is%20minimal.
> 
> I was talking to a senior software engineer recently who says he usually implements a complete functional feature and then rewrites it as a series of linked smaller PRs that he aims to keep under ~100 lines of code to facilitate better code quality, make reviews more bite-sized, and increase integration velocity.
> 
> Audio quality metrics will be implemented here: #270

Totally agree on keeping PRs small. The divide et impera was by the Romans after all, and I definitely won't ignore my origins lol 

That said, my main concern here is merging something into main that isn’t actually usable yet. senselab is meant to provide working functionalities, so users expect released code to be functional. Like @satra  mentioned in our biometrics meetings some time ago, we should avoid merging and keeping non-working or placeholder code in main. 

How about merging these PRs into a feature branch instead? That way, we can bring everything into main once the audio quality metrics are fully working and useful. Let me know what you think!

satra:
this functional framework needs a significant amount of design considerations. you can always leave it as a branch and evaluate it as you implement the metrics. i.e. keep this as a draft pr as you work and incorporate other pieces into it. i agree with @fabiocat93 that there is no good reason for this to be merged into main at this point. personally i like main to be a functionally complete state of the project that can be released at any point. git provides a completely reasonably mechanism to have feature branches that can be developed on their own.

ibevers:
@fabiocat93 @satra Okay, I will make a feature branch and merge into that.

PR #258, opened 2025-02-28
Restructuring senselab in submodules (audio, text, video)

### LINKED ISSUE #145: Modularize dependencies into audio vs video vs text
### Description

As our list of dependencies grows longer and longer, we should think about how to best manage them. A good start might be to split them into three categories: audio, video, and text, so that users can install only the dependencies they plan to use. In the future we may need to further break them apart further into task-level modularization.

### Tasks

- [ ] Determine which dependencies fall into the categories: `core`, `audio`, `video`, and `text`
- [ ] Update `pyproject.yaml` accordingly
- [ ] Update `README.txt` with installation instructions
- [ ] Add decorator or some other feature to interfaces to indicate that it cannot be imported because a dependency is missing.

### Freeform Notes

_No response_

fabiocat93:
## Description
This PR is about restructuring senselab in submodules (audio, text, video). Once this is done, it will be possible to install senselab with `pip install senselab['audio']` if you are only interested in our audio pipelines and utilities, or  `pip install senselab['all']`, if you are interested in the entire thing.

## Motivation and Context
This code restructure, will make senselab more modular and hopefully more easily maintainable and will enable users to install only the dependencies they require.

## How Has This Been Tested?
Unit tests. I have added some more checks that happen when only the core dependencies are installed.

## Types of changes

## Checklist:


fabiocat93:
So far, I have organized dependencies in core + submodules (audio, text, video). Still need to:

- [x] organize code and imports in the submodules
- [x] organize tests
- [x] organize CI

@satra @wilke0818 @ibevers @900miles  if you happen to have suggestions, pointers, ... please let me know. Any help with this is appreciated

wilke0818:
I mean if we are considering something like this which feels like a major refactoring, we should consider what else might need to/could change at the same time (unless you don't see this as a major refactoring/senselabV2.0)? For me, some things that have come up before are how much abstraction we give and with that the need to think of how to make code more performative (batching) as well as recently the idea of lazy-loading audio and videos which the construction of such classes isn't difficult, but how we would rework existing tasks to fit it is

900miles:
> I mean if we are considering something like this which feels like a major refactoring, we should consider what else might need to/could change at the same time (unless you don't see this as a major refactoring/senselabV2.0)? For me, some things that have come up before are how much abstraction we give and with that the need to think of how to make code more performative (batching) as well as recently the idea of lazy-loading audio and videos which the construction of such classes isn't difficult, but how we would rework existing tasks to fit it is

I believe this is a relatively small change implementation-wise, whereas those seem like bigger changes that might require discussion. I think those are good things to discuss but I don't think it's necessary to batch them with this PR.

900miles:
> So far, I have organized dependencies in core + submodules (audio, text, video). Still need to:
> 
>     * [ ]  organize code and imports in the submodules
> 
>     * [ ]  organize tests
> 
>     * [ ]  organize CI

I would add "update docs" to this, to explain to people the updated installation method.

fabiocat93:
> I mean if we are considering something like this which feels like a major refactoring, we should consider what else might need to/could change at the same time (unless you don't see this as a major refactoring/senselabV2.0)? For me, some things that have come up before are how much abstraction we give and with that the need to think of how to make code more performative (batching) as well as recently the idea of lazy-loading audio and videos which the construction of such classes isn't difficult, but how we would rework existing tasks to fit it is

I agree that this first phase of `senselab` was helpful for everyone to assess some of the existing challenges (conceptual and technical) and a major refactoring addressing (some of) these would be helpful.  **Let's see if the others have some more suggested fixes to be included in @wilke0818 's list.** after that, I suggest we meet as @900miles said and decide on every point. I am not against making this PR the major change and releasing senselab 1.0 (we are still in our 0 version for now). 

In terms of timeline, I would like to have this done in the next month, since senselab will be presented at the b2ai consortium meeting in April and would be a nice opportunity to advertise it and to receive a feedback from the community (there is also a hackathon at the event where people could play with it)

wilke0818:
is pip install senselab['all'] equivalent to pip install senselab

.github/workflows/tests.yaml
wilke0818, line 1: just gonna trust you on this one tbh

README.md
```diff
@@ -43,6 +43,15 @@ Or get the newest development version via:
 pip install git+https://github.com/sensein/senselab.git
 ```
 
+If you want to install only audio dependencies, you do:
+```sh
+pip install senselab['audio']
+```
+To install video and text extras, please do:
+```sh
+pip install senselab[video,text]
```
wilke0818, line 52: do these need to be strings?

pyproject.toml
```diff
@@ -10,13 +10,13 @@ maintainers = [
...
 ]
-license = {text = "Apache-2.0"}
+license = {file = "LICENSE.txt"}
 readme = "README.md"
 requires-python = ">=3.10,<3.13"
 homepage = "https://github.com/sensein/senselab"
 repository = "https://github.com/sensein/senselab"
 documentation = "https://sensein.github.io/senselab"
-keywords = ["voice", "speech", "audio"]
+keywords = ["audio", "voice", "speech", "video", "pose", "face", "text", "nlp"]
```
wilke0818, line 19: maybe remove nlp?

src/senselab/audio/data_structures/audio.py
```diff
@@ -68,6 +74,15 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
             filepath: Filepath of the audio file to read from
             metadata: Additional information associated with the audio file
         """
+        if not TORCHAUDIO_AVAILABLE:
```
wilke0818, line 77: this feels a little messy. is there no better way to do this?

src/tests/audio/data_structures/audio_test.py
```diff
@@ -6,17 +6,31 @@
...
 
 def load_audio(file_path: str) -> Tuple[torch.Tensor, int]:
     """Loads audio data from the given file path."""
     return torchaudio.load(file_path)
 
 
+@pytest.mark.skipif(TORCHAUDIO_AVAILABLE, reason="torchaudio is installed.")
+def test_audio_creation_error() -> None:
+    """Tests audio creation with invalid input."""
+    with pytest.raises(ImportError):
```
wilke0818: is there a way to do mocking to act as if a module is not installed rather than only running this if it isn't installed?

.github/workflows/tests.yaml
wilke0818: This could add a lot of tests but prob best to have a test that runs each of the single modules being installed, as well as all combinations of them? Don't know if the changes here currently do this.

wilke0818 (approved):
Appreciate the work! I think that my main issue with this change is that it tries to make user interfacing with the package easier/nicer, which I think it generally succeeds at that (tho if the user can't just do `pip install senselab` anymore I do think we might've made it harder on the average user), but that in making it easier on the user, we've made it harder/worse for developers. It's changes like this when it is a tradeoff between these two things that I'm not sure how far we want to swing one way or the other

src/tests/audio/data_structures/audio_test.py
wilke0818: I think in general we have not done a good job at mocking our dependencies though that definitely is something done more in software where I think you can have a little more trust in your dependencies (other teams). IDK if @satra has thoughts on that.

fabiocat93:
> is pip install senselab['all'] equivalent to pip install senselab

nope. `pip install senselab` only installs the core dependencies, while `pip install senselab['all ']` installs it all

.github/workflows/tests.yaml
fabiocat93: I see your point about usability. It must be the same problem that other packages organized in submodules have (e.g., https://github.com/NVIDIA/NeMo). Unfortunately, poetry does not provide a built-in way to make `pip install senselab` behave like `pip install senselab[all]`. 

After these changes, in CI, we test it all with all extra dependencies installed (all submodules) and with only the core dependencies installed. originally, I was thinking to include all combinations of installation conditions, but realized it's not really necessary for how the package is structured rn (we don't have dependency overalps) and also I am not sure we have the budget for that lol

README.md
fabiocat93: from what i understand, it depends on the environment where you install the package

pyproject.toml
fabiocat93: we have people working on integrating language features extraction and more functionalities in that space

src/senselab/audio/data_structures/audio.py
fabiocat93: I am open to suggestions. The idea is: whenever I know an extra dependency is required, I check if it's installed and if not I raise an ImportError.

src/tests/audio/data_structures/audio_test.py
fabiocat93: Mocking was my first attempt (I have tried with `unittest.mock`), but didn't work as expected and I could import all packages anyway. I know there are some workarounds for poetry (https://medium.com/@zalun/mocking-the-unmockable-1bf3cc31dad6), but I am not sure they make our life actually easier or our code cleaner...

fabiocat93:
> > is pip install senselab['all'] equivalent to pip install senselab
> 
> nope. `pip install senselab` only installs the core dependencies, while `pip install senselab['all ']` installs it all

thank you @wilke0818 . I will wait a couple of more days in case someone else has some more suggestions on how to address this @900miles @satra @ibevers

src/senselab/audio/tasks/forced_alignment/constants.py
```diff
@@ -10,6 +10,8 @@
 
 LANGUAGES_WITHOUT_SPACES = ["ja", "zh"]
 
+# TODO: models shouldn't be initialized here, but only in the align function
```
ibevers, line 13: Better to put this in an issue:
https://github.com/sensein/senselab/issues/262

src/senselab/audio/tasks/text_to_speech/api.py
```diff
@@ -19,7 +19,7 @@ def synthesize_texts(
     """Synthesizes speech from all texts using the given model.
 
     This function synthesizes speech from a list of text strings using the specified text-to-speech (TTS) model.
-    It supports models from HuggingFace and `Mars5TTS` and `StyleTTS2` for now.
+    It supports models from HuggingFace and `Mars5TTS` for now.
```
ibevers, line 22: Removing StyleTTS seems out-of-scope for this PR. I would expect to only see changes related to restructuring into submodules. Keeping things tightly within scope keeps our PRs and issues more identifiable with the changes they contain. 

This feels a bit like when legislators make some bill like "Fixing Our Infrastructure" bill and then it contains unrelated things.

ibevers (commented):
I'm happy we are making this change. From reading the code, this seems to be mergable.

src/senselab/audio/tasks/text_to_speech/api.py
fabiocat93: hahaha you can see it in the scope of this PR if you consider that if we want to keep styletts2 we need to include all its dependencies as part of our "audio" extras. And part of the reasons why we are getting rid of styletts2 is it relies on conflicting dependencies which make it non-trivial to integrate

src/senselab/audio/data_structures/audio.py
```diff
@@ -5,14 +5,20 @@
 ease of maintaining the codebase and offering consistent public APIs.
 """
 
+try:
+    import torchaudio
+
+    TORCHAUDIO_AVAILABLE = True
+except ImportError:
```
900miles: We should catch a `ModuleNotFoundError` rather than an `ImportError`. The former is triggered only when a module isn't found, while the latter is triggered also in cases where the module IS found but can't be imported correctly for whatever one of a million reasons. Because we catch this error in order to recommend installing `torchaudio`, we do not want to catch the generic `ImportError`, as this might lead to the rare but frustrating case of `torchaudio` being installed but having an unrelated import error, which we hide and instead tell the user that it's not installed.

900miles: Also do you think it would be better to raise the error here rather than having a flag set which causes the error to appear later on? My inclination is to raise the error here since that's where we're catching it, but I don't have a sense of what that might entail vs raising it later on.

src/senselab/audio/data_structures/audio.py
```diff
@@ -68,6 +74,15 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
             filepath: Filepath of the audio file to read from
             metadata: Additional information associated with the audio file
         """
+        if not TORCHAUDIO_AVAILABLE:
+            raise ImportError(
```
900miles: Fairly pedantic but we should raise a `ModuleNotFoundError` since that is the more descriptive error in python 3.6+.

src/senselab/audio/data_structures/audio.py
```diff
@@ -68,6 +74,15 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+                "`torchaudio` is not installed. "
+                "Please, install senselab audio dependencies using `pip install senselab['audio']`."
```
900miles: Even more pedantic but this shouldn't have a comma in it :)

src/senselab/audio/data_structures/audio.py
```diff
@@ -184,6 +199,12 @@ def save_to_file(
         Note:
             - https://pytorch.org/audio/master/generated/torchaudio.save.html
         """
+        if not TORCHAUDIO_AVAILABLE:
+            raise ImportError(
+                "`torchaudio` is not installed. "
+                "Please, install senselab audio dependencies using `pip install senselab['audio']`."
```
900miles: Same here

src/senselab/audio/tasks/data_augmentation/api.py
```diff
@@ -2,18 +2,29 @@
...
 from senselab.audio.data_structures import Audio
 from senselab.audio.tasks.data_augmentation.audiomentations import augment_audios_with_audiomentations
 from senselab.audio.tasks.data_augmentation.torch_audiomentations import augment_audios_with_torch_audiomentations
 from senselab.utils.data_structures import DeviceType
 
+try:
+    from audiomentations import Compose as AudiomentationsCompose
+
+    AUDIOMENTATIONS_AVAILABLE = True
+except ImportError:
```
900miles: Same thing as above wrt `ModuleNotFoundError`

src/senselab/audio/tasks/data_augmentation/api.py
```diff
@@ -2,18 +2,29 @@
...
+    AUDIOMENTATIONS_AVAILABLE = False
+
+try:
+    from torch_audiomentations import Compose as TorchAudiomentationsCompose
+
+    TORCH_AUDIOMENTATIONS_AVAILABLE = True
+except ImportError:
```
900miles: and here

src/senselab/audio/tasks/data_augmentation/api.py
```diff
@@ -30,6 +41,14 @@ def augment_audios(
     Returns:
         List of augmented audios.
     """
+    if not AUDIOMENTATIONS_AVAILABLE:
+        raise ImportError(
+            "`audiomentations` is not installed. Please install it using:\n\n" "    pip install senselab['audio']"
```
900miles: Not super important but we should have a consistent error message for missing dependencies, and this one is worded differently to the torchaudio ones. Also should be `ModuleNotFoundError`

src/senselab/audio/tasks/data_augmentation/api.py
```diff
@@ -30,6 +41,14 @@ def augment_audios(
...
+        )
+    if not TORCH_AUDIOMENTATIONS_AVAILABLE:
+        raise ImportError(
```
900miles: `ModuleNotFoundError`

src/senselab/utils/data_structures/__init__.py
```diff
@@ -12,9 +11,15 @@
...
     SpeechBrainModel,
-    TorchAudioModel,
     TorchModel,
     check_hf_repo_exists,
 )
+
+# Conditionally import TorchAudioModel only if torchaudio is available
+try:
+    from .model import TorchAudioModel
+except ImportError:
```
900miles, line 21: Actually this one should remain `ImportError` , because `.model` IS found, but has an error upon importing, so it uses the more generic error class.

tutorials/audio/features_extraction.ipynb
```diff
@@ -35,13 +81,23 @@
...
+                        "\u001b[0;31mImportError\u001b[0m: `torchaudio` is not installed.Please, Install senselab audio dependencies using `pip install senselab['audio']`."
```
900miles: Not sure we want to include this, especially if it's just showing an import error :)

src/senselab/audio/tasks/data_augmentation/api.py
900miles: I wonder if it would be good to create our own error class template, so we can just do something like `raise ModuleNotFoundError("audiomentations", "audio")` or whatever, and have it generate the message for us.

src/senselab/audio/data_structures/audio.py
wilke0818: So I think raising it later helps in the case of Video where say you get an Audio object but you're not actually using it at all. I guess the idea is like you can use certain functionality without having all the dependencies, it's only when you try to use something you don't have that you raise an error. But I can see the other way working fine too

src/senselab/audio/tasks/data_augmentation/api.py
wilke0818: I agree that there are probably ways to clean this up to give a little bit more consistency/readability

tutorials/audio/features_extraction.ipynb
fabiocat93: lol. you are right

src/senselab/audio/tasks/data_augmentation/api.py
fabiocat93: I have just made errors consistent for now. Let me open an issue for a custom error class, which sounds like a good idea

src/senselab/audio/data_structures/audio.py
fabiocat93: Fixed

src/senselab/audio/data_structures/audio.py
fabiocat93: Nice catch @900miles !! I have changed all relevant `ImportError` and made them `ModuleNotFoundError `. About when to raise the error message, for now I like it where it is because the current approach allows having some functionalities working, when they don't require the missing package

900miles: Sounds good. Thanks @fabiocat93 !

wilke0818:
@fabiocat93 I noticed you might need to change the tutorials where we install senselab (if it is not installed) to install the appropriate modules.

fabiocat93:
> @fabiocat93 I noticed you might need to change the tutorials where we install senselab (if it is not installed) to install the appropriate modules.

Nice catch. Let me fix this before I merge

PR #264, opened 2025-03-10
Initialize HF models using their preferred data type

fabiocat93:
## Description
This PR updates the model initialization logic to rely on the Hugging Face model configuration (`config.torch_dtype`) for selecting the best `dtype`, rather than manually forcing it. This change was suggested by @ wilke0818 long time ago.

## Motivation and Context
This ensures compatibility with various model architectures and optimizes performance based on the model's intended precision. 

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/audio/tasks/data_augmentation/torch_audiomentations.py
+++ b/src/senselab/audio/tasks/data_augmentation/torch_audiomentations.py
@@ -50,7 +50,7 @@ def augment_audios_with_torch_audiomentations(
         )
 
     augmentation.output_type = "dict"
-    device_type, dtype = _select_device_and_dtype(
+    device_type, _ = _select_device_and_dtype(
         user_preference=device, compatible_devices=[DeviceType.CUDA, DeviceType.CPU]
     )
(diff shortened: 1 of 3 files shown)


src/senselab/audio/tasks/data_augmentation/torch_audiomentations.py
```diff
@@ -50,7 +50,7 @@ def augment_audios_with_torch_audiomentations(
         )
 
     augmentation.output_type = "dict"
-    device_type, dtype = _select_device_and_dtype(
+    device_type, _ = _select_device_and_dtype(
```
wilke0818, line 53: are there cases where we still want the dtype returned? If not, then perhaps we should just return device type

wilke0818 (approved):

src/senselab/audio/tasks/data_augmentation/torch_audiomentations.py
fabiocat93: all those cases that are not huggingface-based

PR #265, opened 2025-03-10
adding freevc, openvoice and knnvc (new implementation) for voice cloning

### LINKED ISSUE #195: Task [voice cloning] integrate freevc
Integrate [FreeVC](https://github.com/OlaWod/FreeVC) to senselab's voice cloning API.

fabiocat93:
## Description
This PR is for adding coqui-tts-based models for voice cloning. 

## Motivation and Context
With this PR we will enlarge the pool of models for voice cloning.

## How Has This Been Tested?
Unit tests.

## Types of changes

## Checklist:


src/senselab/utils/data_structures/model.py
```diff
@@ -101,6 +108,32 @@ class SentenceTransformersModel(HFModel):
     pass
 
 
+class CoquiTTSModel(SenselabModel):
```
wilke0818, line 111: Thoughts on whether this belongs here or if we should have it somewhere within Audio? Theoretically the try/except shouldn't give users issues unless they're trying to Audio related stuff but one might expect that something in the Utils (which would be part of the core) be runnable with only core dependencies. Just a thought.

src/senselab/audio/tasks/voice_cloning/coqui.py
```diff
@@ -0,0 +1,118 @@
...
+from senselab.utils.data_structures import CoquiTTSModel, DeviceType, _select_device_and_dtype
+
+
+class CoquiVoiceCloner:
+    """A factory for managing voice cloning pipelines using Coqui TTS."""
+
+    _models: Dict[str, "TTS"] = {}
+
+    @classmethod
+    def _get_tts_model(cls, model_id: Union[str, Path], user_preference: Optional[DeviceType] = None) -> "TTS":
```
wilke0818, line 24: tts is a little bit of a misnomer given we're using it for voice conversion. Idk if I would change it since it is their naming scheme though

wilke0818 (approved):
I'm a little confused where the freevc and openvoice stuff is? Are those just other Coqui models?

src/senselab/audio/tasks/voice_cloning/coqui.py
fabiocat93: I know haha I understand how this may be confusing, but that's just the name of their package

src/senselab/utils/data_structures/model.py
fabiocat93: Fair point. I followed the same logic until now, considering that torchaudio models are also part of the same util. How about keeping it like this for now and considering moving those classes as part of a different PR?

fabiocat93:
> I'm a little confused where the freevc and openvoice stuff is? Are those just other Coqui models?

Yup. `coqui-tts` offers an interface to different models, including freevc, openvoice and knnvc

PR #215, opened 2024-12-05
Add Word-Level Alignment

ibevers:
## Description

- Fixes a bug that was leading to incorrect alignments.
- Updates the output to give transcription, sentence, word, and character timestamps.
- Adds a recursive evaluation function for ensuring the differences in timestamps between two alignments are under a certain threshold.
- Adds a test that validates that the generated alignments match a publicly available alignment with a tolerance of .1 second for every start and end timestamp for every aligned level present in both alignments.
- Uses SenseLab native data types more (can be fully converted to SenseLab types in the future).
- Adds optional levels so the user can include or exclude utterance, word, and character-level alignments.
- Adds a tutorial for forced alignment.
- Adds a docs.md for forced alignment.

## Related Issue(s)

https://github.com/sensein/senselab/issues/14

## Motivation and Context

Corrects buggy forced alignment and adds more granular alignment (word and character), which is useful for various purposes. This is a character-based approach. I plan to add a phoneme-based approach in the future to compare to this.

## How Has This Been Tested?

Added test to compare with publicly available alignment. Plan to do more comprehensive evaluation on TIMIT as soon as possible. 

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [x] I have added tests to cover my changes.
- [x] All new and existing tests passed.
- [x] My code follows the code style of this project.

PR #273, opened 2025-03-13
Added sparc for voice cloning

fabiocat93:
## Description
Added [SPARC](https://github.com/Berkeley-Speech-Group/Speech-Articulatory-Coding) model for voice cloining.

## Motivation and Context
This voice cloning model is based on Vocal Tract Kinematics (see https://arxiv.org/abs/2406.12998), which may enable preserving articulatory features as in the source voice while changing the voice texture. 

## How Has This Been Tested?
Unit tests.

## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/audio/tasks/voice_cloning/api.py
+++ b/src/senselab/audio/tasks/voice_cloning/api.py
@@ -4,6 +4,7 @@ from typing import Any, List, Optional
 
 from senselab.audio.data_structures import Audio
 from senselab.audio.tasks.voice_cloning.coqui import CoquiVoiceCloner
+from senselab.audio.tasks.voice_cloning.sparc import SparcVoiceCloner
 from senselab.utils.data_structures import CoquiTTSModel, DeviceType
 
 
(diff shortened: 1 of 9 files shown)


src/senselab/audio/tasks/voice_cloning/api.py
```diff
@@ -20,48 +21,54 @@ def clone_voices(
...
     """
-    if model is None:
-        model = CoquiTTSModel(path_or_uri="voice_conversion_models/multilingual/multi-dataset/knnvc")
     if len(source_audios) != len(target_audios):
         raise ValueError("The list of source and target audios must have the same length.")
 
-    if isinstance(model, CoquiTTSModel):
+    if model is None:
+        lang = kwargs.pop("lang", None)
+        return SparcVoiceCloner.clone_voices(
```
wilke0818, line 66: why are we using this as a default for providing none? We should prob have a way to specify a Sparc model similar to how we do with Coqui.

wilke0818 (approved):
Looks alright. Not sure I am a big fan of how we're using Sparc as a default/no other way to specify it. Seems a bit inconsistent with the rest of the package.

fabiocat93:
@wilke0818 fyi, in order to pass the tests on the remote runner I had to setup the `HF_HOME` differently (we were having out of memory issues otherwise)

ibevers:
>  in order to pass the tests on the remote runner I had to setup the HF_HOME differently (we were having out of memory issues otherwise)

@fabiocat93 Could that approach also work for https://github.com/sensein/senselab/pull/215?

fabiocat93:
> > in order to pass the tests on the remote runner I had to setup the HF_HOME differently (we were having out of memory issues otherwise)
> 
> @fabiocat93 Could that approach also work for #215?

I don't think so, given that here we had an explicit "running out of space" error, which I don't see in [#215](https://github.com/sensein/senselab/pull/215). In [#215](https://github.com/sensein/senselab/pull/215), I see a segmentation fault rn

Issue #270, opened 2025-03-12
Task: Add 20 AQMs to bioacoustic QC


ibevers:
### Description

Depends on: https://github.com/sensein/senselab/issues/269
Depended on: https://github.com/sensein/senselab/issues/271

Adds 20 initial AQMs to bioacoustic QC to make it useful for an initial run on the Bridge2AI dataset.

### Tasks

- [x] Decide on AQMs to add. See here for starters: https://github.com/sensein/senselab/pull/256
- [x] Implement AQMs

Veronika271:
Hello, @fabiocat93! I'm working on writing a program to mark which audio samples have too few words (and therefore not enough data to produce meaningful statistical analysis.) Do you have any suggestions for what the procedure should be if the number of words is below the threshold?
I'm still trying to determine what kinds of statistical analysis are important enough to be used to set the threshold - so far, I've read some literature supporting setting it at 75 or 100 words, but I'm worried that will exclude way too much data.
Thank you!
Veronika

fabiocat93:
> Hello, [@fabiocat93](https://github.com/fabiocat93)! I'm working on writing a program to mark which audio samples have too few words (and therefore not enough data to produce meaningful statistical analysis.) Do you have any suggestions for what the procedure should be if the number of words is below the threshold? I'm still trying to determine what kinds of statistical analysis are important enough to be used to set the threshold - so far, I've read some literature supporting setting it at 75 or 100 words, but I'm worried that will exclude way too much data. Thank you! Veronika

hi @Veronika271 , I am a bit confused here. The number of words varies depending on the task/experiment, right? Are you targeting a specific condition here?

Veronika271:
Hi! My goal is to try to remove audio samples that have too few words to be analyzable, and hopefully not to accidentally exclude any samples based on condition. My logic was that a sample with something like five words probably will likely not have enough data to yield a good understanding of mean length of utterance or some other vocal features. Hopefully that applies to most tasks!

fabiocat93:
> Hi! My goal is to try to remove audio samples that have too few words to be analyzable, and hopefully not to accidentally exclude any samples based on condition. My logic was that a sample with something like five words probably will likely not have enough data to yield a good understanding of mean length of utterance or some other vocal features. Hopefully that applies to most tasks!

I agree it's a good plan to prepare a pipeline for extracting the number of words and comparing it against an expected range. But I think you should make the range a parameter that the user can customize

Veronika271:
[...]

Issue #269, opened 2025-03-12
Task: Add storage of QC AQMs and check results in a dataframe


ibevers:
### Description

Depends on: https://github.com/sensein/senselab/pull/256
Depended on: https://github.com/sensein/senselab/issues/270

Updates the bioacoustic computation graph to create or use an input dataframe that stores AQMs and check results. This will facilitate integration with Snorkel and easier downstream processing.

### Tasks

- [x] Implement input data frame
- [x] Implement storage of AQMs and check results in the dataframe

ibevers:
Completed by:
https://github.com/sensein/senselab/pull/275

Issue #253, opened 2025-02-21
Task:  Implement initial version of bioacoustic quality control


ibevers:
### Description

Implement an initial version of the bioacoustic quality control pipeline Rahul and I have designed.

### Tasks

- [x] Set up tree structure
- [x] Add a few functions for critical checks for testability

PR #277, opened 2025-03-14
276 Feature Implement Quality Control Framework

### LINKED ISSUE #276: Feature: bioacoustic quality control framework
### Description

Implements a bioacoustic quality control framework that is mergable into main.

### Tasks

- [x] Implement initial version: https://github.com/sensein/senselab/pull/256
- [x] Add storage of AQMs and check results in a dataframe: https://github.com/sensein/senselab/issues/269
- [x] Add 20 AQMs: https://github.com/sensein/senselab/issues/270

### Freeform Notes

_No response_

ibevers:
## Description

Disclaimer: This PR is very long because we decided we wanted this feature to be relatively mature before merging to main, and there were complicated dependencies between the branches that fed into this, leading to them being harder to review independently. 

Implements a quality control framework for audio files that uses a taxonomy-based approach.

For a given audio file, there are three steps for quality control. First, relevant quality metrics are computed from the audio waveform and metadata in the Audio object. Second, checks are computed on these metrics that inconclusively indicate whether a given file should be included or excluded from the dataset. Third, a weak supervision model (Snorkel) is trained that conclusively labels audio files as include or exclude based on the check results. The actual audio files are not modified in any way to give the user maximum agency in how they incorporate information from the quality control process. 

What metrics and checks (evaluations) are computed for each file in the dataset is determined by a taxonomy. When an audio file is labeled according to a particular node in the input taxonomy, the evaluations for that node and all sub-nodes in the tree are computed for that audio file. 

While multiple evaluations are provided, the default taxonomy has a single node (labeled 'bioacoustic') and the following evaluations.

Default metrics:
- amplitude_modulation_depth_metric
- peak_snr_from_spectral_metric
- proportion_clipped_metric
- proportion_silent_metric
- spectral_gating_snr_metric

Default checks:
-  completely_silent_check
-  high_proportion_clipped_check
-  low_spectral_gating_snr_check
-  very_low_amplitude_modulation_depth_check
-  very_low_peak_snr_from_spectral_check

These were selected for aligning well with the quality labels (bad, good, excellent) on the Coswara dataset. On Coswara, this baseline approach excluded 49.0% of bad (1306/2666), 4.7% of good (145/3054), and 3.1% of excellent (592/18992) recordings (precision=.639, recall=.490, F1=0.555). A dataset-specific taxonomy (`taxonomies.BRIDGE2AI_VOICE_TAXONOMY`) provides an example of a more customized taxonomy that can be modified for similar audio datasets. **Users are encouraged to make customized taxonomies and evaluations for their use cases and datasets and contribute them to senselab!**

## Related Issue(s)

[Task: Implement initial version of bioacoustic quality control #253](https://github.com/sensein/senselab/issues/253)
[Task: Add storage of QC AQMs and check results in a dataframe #269](https://github.com/sensein/senselab/issues/269)
[Task: Add 20 AQMs to bioacoustic QC #270](https://github.com/sensein/senselab/issues/270)
[Task: Update Bioacoustic QC Taxonomy Tree with B2AI Activities #298](https://github.com/sensein/senselab/issues/298)
[Task: Add bioacoustic QC checks based on B2AI #300](https://github.com/sensein/senselab/issues/300)
[Task: Recursive Bioacoustic QC Class Structure #296](https://github.com/sensein/senselab/issues/296)
[Task: parallelize bioacoustic qc #297](https://github.com/sensein/senselab/issues/297)
[Task: Add Computation Reuse to Bioacoustic QC #299](https://github.com/sensein/senselab/issues/299)
[Task: Update Bioacoustic QC to Run on B2AI dataset #305](https://github.com/sensein/senselab/issues/305)
[Task: Calculate Metrics over Windows #326](https://github.com/sensein/senselab/issues/326)
[Task: Integrate Rahul's bioacoustic checks to senselab #330](https://github.com/sensein/senselab/issues/330)
[Task: Add Snorkel Automated Review #347](https://github.com/sensein/senselab/issues/347)
[Task: Replace Pydra with Joblib #409](https://github.com/sensein/senselab/issues/409)
[Task: Support Arbitrary Arguments to Evaluation Functions #410](https://github.com/sensein/senselab/issues/410)

## Motivation and Context

This enables quality control of audio. Motivated by need for quality control on the Bridge2AI Voice dataset. 

## How Has This Been Tested?

- Unit tests
- Experimental usage on Coswara and Bridge2AI Voice

## Types of changes

## Checklist:


ibevers:
Fabio's review feedback from this branch: https://github.com/sensein/senselab/pull/322

[fabiocat93](https://github.com/fabiocat93) [3 weeks ago](https://github.com/sensein/senselab/pull/322#discussion_r2217063819)
If I undertsand it right, the get_metric function introduces logic to avoid recomputing metrics already present in the df. I like this in concept, but why not manage this using the cache mechanism already built in in Pydra as we do for the other audio features?

Member
@[fabiocat93](https://github.com/fabiocat93) fabiocat93 [3 weeks ago](https://github.com/sensein/senselab/pull/322#discussion_r2217064652)
The extract_features_from_audios function tries to cleanly separate feature extractor configuration, execution, and result aggregation. Consider aligning get_metric with the structure of extract_features_from_audios. You could define a extract_metrics_from_audios function that:

Accepts a list of audios
Accepts a list of metrics
Returns a list of dicts
Uses the same plugin/config pattern as the rest of the package
Having different logics inside of the same package may confuse future developers or users

Member
@[fabiocat93](https://github.com/fabiocat93) fabiocat93 [3 weeks ago](https://github.com/sensein/senselab/pull/322#discussion_r2217065163)
As a minor thing: get_metric uses metric_function.__name__ as the key. This assumes that all metric functions have unique names but if two different metrics are implemented with the same function name (e.g., via wrapper functions), this could siilently overwrite or misread cached values. You may want to use a namespacing approach like f"{metric_function.__module__}.{metric_function.__name__}"

Member
@[fabiocat93](https://github.com/fabiocat93) fabiocat93 [3 weeks ago](https://github.com/sensein/senselab/pull/322#discussion_r2217065435)
Please treat filepaths as Path-like objects instead of strings, unless you do that for a reason

Member
@[fabiocat93](https://github.com/fabiocat93) fabiocat93 [3 weeks ago](https://github.com/sensein/senselab/pull/322#discussion_r2217066889)
The current setup works one metric per file. If I want 10 metrics across 5000 files, that’s 50,000 function calls... Integrating pydra would help optimize execution time

src/senselab/audio/data_structures/audio.py
```diff
@@ -311,7 +311,7 @@ def save_to_file(
             if not os.path.exists(output_dir):
                 os.makedirs(output_dir)
             torchaudio.save(
-                uri=file_path,
+                uri=str(file_path),
```
wilke0818: torchaudio should be able to take either Path or str

src/senselab/audio/tasks/quality_control/taxonomies/__init__.py
wilke0818, line 1: perhaps a larger discussion to be had of should classes related to Taxonomy (and shared classes for other tasks) live in data_structures either for audio or utilities

src/senselab/audio/tasks/quality_control/taxonomies/bioacoustic.py
wilke0818, line 1: larger comment not neccessarily to just this file: could you create some documentation like the (doc.md or whatever we have elsewhere) sort of explaining the choices you've made and what the differences are between some of these. I know you have a draft of the paper so perhaps you might be able to just do some copy and pasting.

src/senselab/audio/tasks/quality_control/taxonomies/bridge2ai_voice.py
```diff
@@ -0,0 +1,270 @@
...
+
+    glides.add_child("glides_high_to_low", glides_high_to_low)
+    glides.add_child("glides_low_to_high", glides_low_to_high)
+
+    return root
+
+
+# Create the global Bridge2AI voice taxonomy tree instance
+BRIDGE2AI_VOICE_TAXONOMY = build_bridge2ai_voice_taxonomy()
+BRIDGE2AI_VOICE_TAXONOMY.print_tree()
```
wilke0818: debug statement?

src/senselab/audio/tasks/quality_control/checks.py
```diff
@@ -0,0 +1,860 @@
...
+
+    Args:
+        audio_or_path: An Audio instance or filepath to the audio file.
+        df: Optional DataFrame with ``dynamic_range_metric``.
+
+    Returns:
+        True if the audio has completely zero dynamic range, None if evaluation fails.
+    """
+    result = get_evaluation(audio_or_path, dynamic_range_metric, df)
+    if result is None:
```
wilke0818, line 71: under what circumstances are you getting None?

src/senselab/audio/tasks/quality_control/checks.py
wilke0818, line 1: on the one hand I like that these are all well described, on the other hand, the code here is very redundant and feels like there might be a motif that could be re-used (though perhaps that is what get_evaluation already helps with (going through files in order so haven't seen it yet lol)

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+
+    Returns:
+        The evaluation result for the audio, or None if evaluation fails
+    """
+    # Convert string path to Audio object if needed
+    if isinstance(audio, str):
+        try:
+            audio = Audio(filepath=audio)
+        except Exception as e:
+            print(f"Warning: Failed to load audio from {audio}: {e}")
```
wilke0818: I would like to be more consistent with our erroring and try/except. Can you use the logger instead of printing? There's also a question of should the whole thing fail if the user's audio fails. I'm fine with it not but maybe be clearer in the description

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+            has_windowed = (
+                "windowed_evaluations" in existing_results and function_name in existing_results["windowed_evaluations"]
+            )
+            if has_windowed:
+                windowed_result = existing_results["windowed_evaluations"][function_name]
+                if isinstance(windowed_result, list) and len(windowed_result) > window_idx:
+                    return windowed_result[window_idx]
+        else:
+            # For non-windows, look in evaluations
+            has_evaluations = "evaluations" in existing_results and function_name in existing_results["evaluations"]
```
wilke0818, line 70: seems like you are expecting a given structure. I assume this means that you expect the user to never really pass this cache but rather for outside code to call this method and have this internally cache results?

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+            if has_evaluations:
+                return existing_results["evaluations"][function_name]
+
+    # Compute result
+    try:
+        result = evaluation_function(audio)
+        return result
+    except Exception as e:
+        print(f"Warning: Failed to compute {function_name}: {e}")
```
wilke0818: again logger

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+        List of evaluation results for each window, or None if evaluation fails
+    """
+    window_size = int(window_size_sec * audio.sampling_rate)
+    step_size = int(step_size_sec * audio.sampling_rate)
+
+    values: List[EvalResult] = []
+
+    try:
+        # Generate windows and compute evaluations
+        windows = audio.window_generator(window_size, step_size)
```
wilke0818, line 111: not a you issue, but should the Audio class be more adaptive (e.g. not force users to calculate these sizes when passing them in?

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+        function_name = evaluation_function.__name__
+        msg = f"Warning: Failed to compute windowed evaluation for '{function_name}': {e}"
+        print(msg)
+        return None
+
+
+def evaluate_audio(
+    audio_path: str,
+    activity: str,
+    evaluations: Sequence[Callable[[Audio], Union[float, bool, str]]],
```
wilke0818: swap these with your more readable names?

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+    output_dir: Optional[Path] = None,
+    window_size_sec: float = 0.025,
```
wilke0818: why these defaults? Any literature/other packages that make these the optimal choices? I ask because for one, a lot of encoding/SSL models will return values at (0.02 sec or 20ms)

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,744 @@
...
+
+
+def primary_speaker_ratio_metric(audio: Audio) -> float:
+    """Calculates the ratio of the primary speaker's duration to the total duration.
+
+    The primary speaker ratio is computed from speaker diarization results.
+    If diarization is not available in audio.metadata, it will be computed automatically.
+
+    Args:
+        audio: The SenseLab Audio object.
```
wilke0818: just senselab I think (spelling)

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,744 @@
...
+                If there are no speakers, return nan.
+                If only one speaker, return 1.0.
+                Else return a number between 0.0 and 1.0.
+    """
+    # Check Audio metadata for precomputed diarization
+    diarization_result: Optional[List[ScriptLine]] = None
+    if audio.metadata:
+        metadata_diarization = audio.metadata.get("diarization")
+        if metadata_diarization is not None:
+            # Handle different storage formats: List[List[ScriptLine]] or List[ScriptLine]
```
wilke0818, line 544: when would we expect so many different formats of results?

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,744 @@
...
+
+    # Compute diarization if not in metadata
+    if diarization_result is None:
+        try:
+            diarization_results = diarize_audios([audio])
+            if not diarization_results or len(diarization_results) == 0:
+                return np.nan
+            diarization_result = diarization_results[0]
+        except Exception as e:
+            print(f"Warning: Failed to compute diarization for primary_speaker_ratio_metric: {e}")
```
wilke0818: logger

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,744 @@
...
+            correlation_values.append(np.mean(valid_corrs))
+
+    return float(np.mean(correlation_values)) if correlation_values else 0.0
+
+
+def primary_speaker_ratio_metric(audio: Audio) -> float:
+    """Calculates the ratio of the primary speaker's duration to the total duration.
+
+    The primary speaker ratio is computed from speaker diarization results.
+    If diarization is not available in audio.metadata, it will be computed automatically.
```
wilke0818, line 528: I don't think we've ever used metadata like this previously. I think it's a good idea, but we should try to standardize our practices with it

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,744 @@
...
+
+    Returns:
+        float: Duration of voice activity in seconds.
+               Returns 0.0 if no voice is detected.
+               Returns np.nan if VAD computation fails.
+    """
+    # Check Audio metadata for precomputed VAD
+    vad_result: Optional[List[ScriptLine]] = None
+    if audio.metadata:
+        metadata_vad = audio.metadata.get("vad")
```
wilke0818, line 612: what if we have diarization and not vad?

src/senselab/audio/tasks/quality_control/quality_control.py
```diff
@@ -0,0 +1,111 @@
+"""Runs bioacoustic activity recording quality control on audio files."""
```
wilke0818: The functions are more generalizable than to just bioacoustic QC

src/senselab/audio/tasks/quality_control/review.py
```diff
@@ -0,0 +1,297 @@
...
+import pandas as pd
+from snorkel.labeling import PandasLFApplier, labeling_function
+from snorkel.labeling.model import LabelModel
+
+from senselab.audio.tasks.quality_control.taxonomies import (
+    BIOACOUSTIC_ACTIVITY_TAXONOMY,
+)
+from senselab.audio.tasks.quality_control.taxonomy import TaxonomyNode
+
+INCLUDE: int = 1
```
wilke0818: I would create an enum or something

src/senselab/audio/tasks/quality_control/review.py
```diff
@@ -0,0 +1,297 @@
...
+    cols: Sequence[str],
+) -> Callable[[pd.Series], int]:
+    """Include a file if all given checks are False or None.
+
+    Args:
+        cols: Sequence of column names to check for failed quality checks.
+
+    Returns:
+        A labeling function that returns INCLUDE if all checks are False/None,
+        otherwise ABSTAIN.
```
wilke0818: the function EXCLUDES otherwise. is that the same as ABSTAINing?

src/senselab/audio/tasks/quality_control/review.py
```diff
@@ -0,0 +1,297 @@
...
+
+    if not taxonomy_check_cols:
+        raise ValueError(f"No check columns found that match taxonomy checks for " f"activity '{activity}'")
+
+    df_checks = df[taxonomy_check_cols]
+
+    if prune_checks:
+        keep_cols, dropped = prune_check_columns(df_checks, correlation_threahold=correlation_threahold)
+        dropped_total = sum(len(v) for v in dropped.values())
+        print(f"Checks total: {df_checks.shape[1]} | kept: {len(keep_cols)} | " f"dropped: {dropped_total}")
```
wilke0818: we should have no print statements, just logging when necessary

src/senselab/audio/tasks/quality_control/review.py
```diff
@@ -0,0 +1,297 @@
...
+    lf_list.append(include_no_failed_checks_label_function(keep_cols))
+    lf_names.append("no_checks_true_include_label_function")
+
+    # Apply labeling functions
+    applier = PandasLFApplier(lfs=lf_list)
+    df_checks_filtered = df_checks[keep_cols]
+    L_train = applier.apply(df_checks_filtered)
+
+    # Verify alignment
+    assert len(L_train) == len(df), f"Mismatch: L_train has {len(L_train)} rows, df has {len(df)} rows"
```
wilke0818: I would make this a check and throw a related error? Unless you feel assertions are fine practice (this is an area I don't know what the best practice is for readability and proper error management)

src/senselab/audio/tasks/quality_control/taxonomy.py
wilke0818, line 1: Should this be placed in data structures/elsewhere?

src/senselab/audio/tasks/quality_control/utils.py
```diff
@@ -0,0 +1,61 @@
+"""Utility functions for bioacoustic quality control."""
+
+from pathlib import Path
+from typing import List, Set
+
+
+def get_audio_files_from_directory(
+    directory_path: str,
+    audio_extensions: Set[str] = {".wav", ".mp3", ".flac", ".m4a", ".ogg", ".aac", ".wma"},
```
wilke0818: Does torch support all of these extensions

src/senselab/audio/tasks/quality_control/taxonomies/bridge2ai_voice.py
```diff
@@ -0,0 +1,271 @@
...
+def build_bridge2ai_voice_taxonomy() -> TaxonomyNode:
+    """Build the Bridge2AI voice assessment taxonomy tree.
+
+    This reuses the bioacoustic taxonomy as the base.
+
+    Returns:
+        The root node of the Bridge2AI voice taxonomy tree
+    """
+    # Reuse the bioacoustic taxonomy (deep copy to avoid mutating the original)
+    root = copy.deepcopy(build_bioacoustic_activity_taxonomy())
```
wilke0818: this is returning a new Node every time. Why do you need the copy? Were you originally using BIOACOUSTIC_ACTIVITY_TAXONOMY? In that case you would need to copy I think

src/senselab/audio/tasks/quality_control/utils.py
```diff
@@ -0,0 +1,100 @@
+"""Utility functions for bioacoustic quality control."""
+
+import os
+from pathlib import Path
+from typing import List, Set, Union
+
+from senselab.utils.data_structures.logging import logger
+
+
+def validate_audio_paths(
```
wilke0818: this is also done when audio objects are created I believe

src/senselab/audio/tasks/quality_control/utils.py
wilke0818, line 1: these functions feel more generic than for just quality control. Not sure if you have thoughts on a placement elsewhere for them

src/tests/audio/data_structures/audio_test.py
```diff
@@ -228,7 +228,7 @@ def test_audio_save_to_file(audio_fixture: str, request: pytest.FixtureRequest)
         assert temp_file_path.exists(), "The audio file was not saved."
 
         # Load the saved file and verify its content
-        loaded_waveform, loaded_sampling_rate = torchaudio.load(temp_file_path)
+        loaded_waveform, loaded_sampling_rate = torchaudio.load(str(temp_file_path))
```
wilke0818: why do we need string?

wilke0818 (approved):

src/senselab/audio/tasks/quality_control/evaluate.py
```diff
@@ -0,0 +1,347 @@
...
+            window_size_sec=window_size_sec,
+            step_size_sec=step_size_sec,
+            skip_windowing=skip_windowing,
+        )
+        records.append(record)
+
+    return records
+
+
+def evaluate_dataset(
```
wilke0818, line 251: I think the `check_quality` in quality_control.py is fine to take in paths, but I think there should be a somewhat high level function that begins using/receives Audio objects especially since there are downstream metrics/checks that might use Audio metadata which will not be possible if the user isn't passing the Audio objects themselves in

ibevers: Good point! @wilke0818

src/senselab/audio/tasks/quality_control/review.py
```diff
@@ -0,0 +1,298 @@
...
+            }
+        )
+
+    return pd.DataFrame(reliability_rows).sort_values(
+        ["agreement_with_label_model", "voted_include"], ascending=[False, False]
+    )
+
+
+def review_files(
+    df_path: str,
```
wilke0818: quality_control does not return the filename but rather the DataFrame. Would recommend either accepting the df as well, having quality control return the filename of the saved file as well, or changing the input here

src/senselab/audio/tasks/quality_control/metrics.py
```diff
@@ -0,0 +1,745 @@
...
+                If only one speaker, return 1.0.
+                Else return a number between 0.0 and 1.0.
+    """
+    # Check Audio metadata for precomputed diarization
+    diarization_result: Optional[List[ScriptLine]] = None
+    if audio.metadata:
+        metadata_diarization = audio.metadata.get("diarization")
+        if metadata_diarization is not None:
+            # Handle different storage formats: List[List[ScriptLine]] or List[ScriptLine]
+            if isinstance(metadata_diarization, list) and len(metadata_diarization) > 0:
```
wilke0818: we shouldn't recompute even if the list is empty

src/senselab/audio/data_structures/audio.py
ibevers: Reverted

src/senselab/audio/tasks/quality_control/taxonomies/__init__.py
ibevers: Agreed. We could eventually make it a subclass of a more general ontology class, if that would be useful

ibevers: We could also use an existing ontology package to senselab

ibevers: https://github.com/sensein/senselab/issues/413

src/senselab/audio/tasks/quality_control/taxonomies/bioacoustic.py
ibevers: @wilke0818 Yes. I will make a docs and a tutorial

ibevers:
To-dos:
- [ ] docs
- [ ] tutorial

src/senselab/audio/tasks/quality_control/taxonomies/bridge2ai_voice.py
ibevers: yep

src/senselab/audio/tasks/quality_control/checks.py
ibevers: This is most likely to satisfy pydantic

ibevers: Sometimes, if an audio waveform is too short, a metric cannot be computed, so it returns None. This makes the code fault-tolerant for such issues

src/senselab/audio/tasks/quality_control/utils.py
ibevers: Yes (added test)

src/senselab/audio/tasks/quality_control/checks.py
ibevers: Agreed--might make sense to use some form of config file. While these checks are simple thresholds on single metrics, the checks could be more complex and involve multiple metrics

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: Definitely--updated.

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: Yes. It's referred to the cache here, but it's same info that the user is given at the end, so it's both human and machine-readable. I updated the comments to be a bit clearer

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: Done

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: That would be nice. Not sure if it's feasible or is optimal, but I made an issue: https://github.com/sensein/senselab/issues/414

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: Done

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: No strong reason. Changed to .02 window and .01 step size

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: done

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: These are possible with different forced alignment configurations

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: done

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: Started an issue:
https://github.com/sensein/senselab/issues/415

src/senselab/audio/tasks/quality_control/quality_control.py
ibevers: Updated

src/senselab/audio/tasks/quality_control/review.py
ibevers: done

src/senselab/audio/tasks/quality_control/review.py
ibevers: No. We concluded after a fair amount of discussion that having an abstain category is worse than having binary include/exclude, so there is no abstain category. Abstain made sense when we were thinking of actually removing audio files in the QC process because it would be riskier to label a file exclude. Having no abstain and making the QC informational rather than taking away peoples' options by modifying the audio files makes mis-categorized excludes somewhat more acceptable, while ensuring that it's clear for users what the most likely thing they should do with their files is.

ibevers: I guess the code still uses abstain internally, but in practice, it will ever label something as abstain.

ibevers: Updated the comments to make this clear

src/senselab/audio/tasks/quality_control/review.py
ibevers: done

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: Updated to use, if available

src/senselab/audio/tasks/quality_control/review.py
ibevers: Removed this. It's cleaner, and this error check is not really necessary since it effectively checks external library behavior.

src/senselab/audio/tasks/quality_control/taxonomy.py
ibevers: TBD:
https://github.com/sensein/senselab/issues/413

src/senselab/audio/tasks/quality_control/checks.py
ibevers: Made an issue: https://github.com/sensein/senselab/issues/416

Leaving this to future work.

src/senselab/audio/tasks/quality_control/taxonomies/bridge2ai_voice.py
ibevers: Good point. I believe so. Removed the deepcopy

src/senselab/audio/tasks/quality_control/utils.py
ibevers: Yes. However, it's helpful for it to be done earlier before evaluate_audio (which actually loads the files) is run to prevent errors

src/senselab/audio/tasks/quality_control/utils.py
ibevers: Moved to input_output.py

src/senselab/audio/tasks/quality_control/utils.py
wilke0818: well the solution to that might be to be passing Audios in

src/tests/audio/data_structures/audio_test.py
ibevers: removed

src/senselab/audio/tasks/quality_control/evaluate.py
ibevers: A problem with inputting audio paths is that storing large numbers of audio paths in memory is not always feasible, as you know.

ibevers: Perhaps it could take in the metadata as a separate input?

ibevers: I'll leave this to future work. Made an issue here: https://github.com/sensein/senselab/issues/417

src/senselab/audio/tasks/quality_control/review.py
ibevers: done

src/senselab/audio/tasks/quality_control/metrics.py
ibevers: fixed

ibevers:
@wilke0818 Thanks for your review and feedback! Would you mind giving this another look?

src/senselab/audio/tasks/quality_control/utils.py
ibevers: We can't rely on loading audio files into memory right now, and thinking through how the metadata updates (e.g., diarization) would work with diarization and a whole dataset is a bit complicated, so I want to address this in a separate issue/PR (https://github.com/sensein/senselab/issues/417). Let's keep this solution provisionally

src/senselab/audio/tasks/input_output/utils.py
```diff
@@ -73,3 +75,96 @@ def save_audios(
     for audio, out_path in audio_tuples:
         audio.save_to_file(os.fspath(out_path), **params)
     return None
+
+
+def validate_audio_paths(
```
wilke0818: nit: should be something akin to get_valid_audio_paths (imo)

src/senselab/audio/tasks/quality_control/checks.py
```diff
@@ -0,0 +1,860 @@
+"""Contains tests for quality control checks.
```
wilke0818: nit: phrasing feels like this is a test file

wilke0818 (approved):

src/senselab/audio/tasks/input_output/utils.py
ibevers: agree

src/senselab/audio/tasks/quality_control/checks.py
ibevers: good point--the phrasing doesn't make sense

src/senselab/audio/tasks/input_output/utils.py
ibevers: updated

src/senselab/audio/tasks/quality_control/checks.py
ibevers: updated

PR #278, opened 2025-03-14
269 task add storage of qc aqms and check results in a dataframe

ibevers:
## Description

- Stores outputs of AQMs (`metrics`) and `checks` in a data frame. 
- Uses a unified `apply_audio_quality_function` to run each check and metric function and update the dataframe. 
- Removes storage of classified results in an updated taxonomy tree, meaning that the taxonomy tree is no longer modified.
- Adds a metrics field to each class in the taxonomy tree.
- Updates the taxonomy tree to store empty metrics and checks as empty lists rather than `None`.

## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/audio/tasks/bioacoustic_qc/checks.py
+++ b/src/senselab/audio/tasks/bioacoustic_qc/checks.py
@@ -49,11 +49,11 @@ def audio_quality_check(audios: List[Audio], activity_audios: List[Audio], condi
     return {"exclude": exclude, "review": [], "passed": passed}
 
 
-def audio_length_positive_check(audios: List[Audio], activity_audios: List[Audio]) -> Dict:
+def audio_length_positive_check(audio: Audio) -> bool:
     """Checks if an Audio object has a positive length."""
-    return audio_quality_check(audios, activity_audios, lambda audio: audio.waveform.numel() == 0)
+    return audio.waveform.numel() == 0
 
 
-def audio_intensity_positive_check(audios: List[Audio], activity_audios: List[Audio]) -> Dict:
+def audio_intensity_positive_check(audio: Audio) -> bool:
     """Checks if an Audio object has nonzero intensity."""
-    return audio_quality_check(audios, activity_audios, lambda audio: torch.sum(torch.abs(audio.waveform)) == 0)
+    return torch.sum(torch.abs(audio.waveform)) == 0
--- a/src/senselab/audio/tasks/bioacoustic_qc/checks.py
+++ b/src/senselab/audio/tasks/bioacoustic_qc/checks.py
@@ -13,64 +13,47 @@ All checks take a list of Audio objects as input and return:
     - A list of Audio objects that passed the check.
 """
 
-from typing import Dict, List
+from typing import Callable, Dict, List
 
 import torch
 
 from senselab.audio.data_structures import Audio
 
 
-def audio_length_positive_check(audios: List[Audio], activity_audios: List[Audio]) -> Dict:
-    """Checks if an Audio object has a positive length.
-
-    Args:
-        audios (Audio): The Audio object to validate.
-        activity_audios (List[Audio]): The subset of Audio objects to check.
-
-    Returns:
-        Dict[str, List[Audio]]: A tuple containing:
-            - A dictionary with:
-                - "exclude": A list of Audio objects that failed the check.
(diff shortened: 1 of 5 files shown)


src/senselab/audio/tasks/bioacoustic_qc/checks.py
```diff
@@ -13,64 +13,19 @@
...
 import torch
 
 from senselab.audio.data_structures import Audio
 
 
-def audio_length_positive_check(audios: List[Audio], activity_audios: List[Audio]) -> Dict:
-    """Checks if an Audio object has a positive length.
+def audio_length_positive_check(audio: Audio) -> bool:
+    """Checks if an Audio object has a positive length."""
+    return audio.waveform.numel() != 0
```
fabiocat93, line 26: here and later in this same script, you could reuse existing features and avoid duplicate code

src/senselab/audio/tasks/bioacoustic_qc/constants.py
```diff
@@ -5,68 +5,79 @@
 BIOACOUSTIC_ACTIVITY_TAXONOMY = {
```
fabiocat93, line 5: I have the feeling that there is a recursive schema in your `BIOACOUSTIC_ACTIVITY_TAXONOMY` that would be more easily maintainable with a recursive class structure

fabiocat93 (commented):
Hi @ibevers, sorry for the delay in reviewing your PR (I was assisting with other branches).

## Development approach + unasked suggestions
If I were in your shoes, I’d structure the development process as follows:

1. Integrate a decision maker first (the core engine that determines whether an audio sample is noisy or not) and allow users to define their own rules, enabling them to experiment with different configurations and evaluate the feasibility of a rule-based QA framework.
2. Implement and validate individual rules. Once the framework is functional, introduce predefined rules (your duration and energy checks are good basic examples).
3. Organize and structure rules – After validating the approach, categorize rules by task and relevant conditions to ensure clarity and usability. This way, you will provide your users with some validated templates and instructions on how to use the rules that we internally test.

Currently, the development seems to be following the reverse approach—starting with a taxonomy of tasks and rules before testing their practical effectiveness. This could be risky, as it might lead to unnecessary implementation efforts. However, I assume you’ve aligned with @satra on the conceptual framework and development priorities, so I’ll limit my input to leaving comments on the code >>>>>>>>>>>>>>

## Code feedback
Generally, I don't see any major red flags, but I noticed that some functions may be reinventing existing solutions. For example, Snorkel’s `LFApplier` already provides functionality to apply multiple checks, output results in different formats, and even include a progress bar. If we’re considering using Snorkel’s engine, reusing their utilities might be a more efficient approach rather than implementing custom solutions. I’m unsure whether they support parallelization, but neither do we at the moment. What's the advantage of your new implementation then? I have left some more comments and suggestions here and there.

 Cheers 🚀

ibevers:
Thanks for your input @Fabio and patience with my reply (lots to consider above). 

> Integrate a decision maker first

I've implemented AQMs since we want to include those in the upcoming B2AI data release. The framework here is meant to derive AQMs and check results in a decision-making agnostic way since decision-making with a model will be dataset specific for now. Decision-making might be integrated into the structure of the taxonomy through automatic exclusion. Evaluating a LabelModel requires a labeled test set, and the framework for deriving relevant AQMs and checks will facilitate creating such test sets. Also, multiple label functions may use the same AQMs, so I'm prioritizing AQMs over label functions/checks. 

The current taxonomy is provisional and will be refined (https://github.com/sensein/senselab/issues/298). 

That's a good point about using the LFApplier. I plan to explore integrating that.

> I’m unsure whether they support parallelization, but neither do we at the moment.

When I get to parallelization (https://github.com/sensein/senselab/issues/297), I'll consider whether LFApplier is the best approach.

src/senselab/audio/tasks/bioacoustic_qc/constants.py
ibevers: Good point: https://github.com/sensein/senselab/issues/296

src/senselab/audio/tasks/bioacoustic_qc/checks.py
ibevers: Good point--plan to work on here: https://github.com/sensein/senselab/issues/299

PR #284, opened 2025-03-18
Added Audio.from_stream + enhanced Audio.from_filepath

### LINKED ISSUE #226: Support: Memory Efficiency for long recordings
### Issue Description

Had trouble performing feature extraction locally on audio recordings longer than a minute. 

### Environment Details

Ram: 16 Gb
Os: Ubuntu 20.04.3 LTS
Python: 3.10.9

### Attempted Solutions

_No response_

### Reproduction Steps

_No response_

### Additional Notes

_No response_

fabiocat93:
## Description
This PR introduces improvements to `Audio.from_filepath` and presents `Audio.from_stream`, also to enhance support for long audio recordings. These changes lay the groundwork for more robust handling of lengthy audio files in future updates.

## Motivation and Context
Issues have been reported with feature extraction from long recordings ([[issue #226](https://github.com/sensein/senselab/issues/226)](https://github.com/sensein/senselab/issues/226)) and loading long audio files from disk, as discussed in meetings. While this PR does not fully resolve those issues, it provides initial enhancements that allow users to implement partial workarounds.

## Changes Introduced
- **Enhanced `Audio.from_filepath`**: Added support for loading only a portion of an audio file by specifying an `offset` and `duration` in seconds. This provides a flexible way to manage large recordings without fully loading them into memory.
- **Integrated `Audio.from_stream`**: Enabled reading from both files and streams. This implementation relies on `soundfile` instead of `torchaudio`, since `torchaudio`'s stream reading requires `FFMPEG >= 4 and < 7`, which is restrictive for users who may have the latest `FFMPEG` version installed.

## Future Work
Further improvements to facilitate long audio processing may include:
- Making generators serializable so they can be passed as parameters to pydra methods.
- Treating `AudioFile` as a generator of `Audio` objects and allowing all methods to accept both `Audio` and `AudioFile`.

These are larger, breaking changes that require careful consideration, so they are deferred for future updates. Also, I am curious to hear what you all think.

## Testing
- Unit tests.

## Type of Changes
- [x] New feature (non-breaking change that adds functionality)

## Checklist
- [x] My code follows the project’s style guidelines.


CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ Please use the following workflow when contributing:
           ```poetry run ruff check --fix```
       - Spell checking:
         ```poetry run codespell```
-10. **Add repository secrets**: From your github web interface, add the following repository secrets: ```CODECOV_TOKEN``` (CodeCov), ```HF_TOKEN``` (HuggingFace), ```PYPI_TOKEN``` (Pypi).
+10. **Add repository secrets** (maintainers only): From your github web interface, add the following repository secrets: ```CODECOV_TOKEN``` (CodeCov), ```HF_TOKEN``` (HuggingFace), ```PYPI_TOKEN``` (Pypi).
```
wilke0818, line 42: should be 9?

CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ Please use the following workflow when contributing:
...
 11. **Submit a pull request**: Once you are done adding your new amazing functionality, submit a pull request to merge the upstream issue branch into the upstream main.
 12. **Don’t worry much about point 9**: Just joking, there’s nothing there – just making sure you're paying attention!
```
wilke0818, line 44: Is this no longer true? Seems like by adding "for maintainers only" you're suggesting that it is important for them/there is something there now?

src/senselab/audio/data_structures/audio.py
```diff
@@ -67,12 +74,30 @@ def convert_to_tensor(
         return temporary_tensor.to(torch.float32)
 
     @classmethod
-    def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Audio":
+    def from_filepath(
+        cls,
+        filepath: str | os.PathLike,
+        offset_in_sec: float = 0.0,
+        duration_in_sec: float = -1.0,
```
wilke0818: is there a reason for -1 vs. Optional/None?

src/senselab/audio/data_structures/audio.py
```diff
@@ -67,12 +74,30 @@ def convert_to_tensor(
...
+        metadata: Optional[Dict] = None,
+        backend: Optional[str] = None,
+    ) -> "Audio":
         """Creates an Audio instance from an audio file.
 
         Args:
-            filepath: Filepath of the audio file to read from
-            metadata: Additional information associated with the audio file
+            filepath: Filepath of the audio file to read from.
+            offset_in_sec: Offset in seconds to start reading the audio file.
```
wilke0818, line 89: Another form of this that I have seen frequently enough is int|float where int describes spot in seconds/sampling rate vs. float is proportion. I think this is fine for now

src/senselab/audio/data_structures/audio.py
```diff
@@ -67,12 +74,30 @@ def convert_to_tensor(
...
+                Default is 0.0 (start from the beginning).
+            duration_in_sec: Duration in seconds to read from the audio file.
```
wilke0818, line 91: Could put a padding variable such that we pad the audio to a given length if the file would be shorter, default being false

src/senselab/audio/data_structures/audio.py
```diff
@@ -67,12 +74,30 @@ def convert_to_tensor(
...
+                This function may return a shorter audio if there is not enough seconds in the given file.
+                Default is -1.0 (read the entire audio file).
+            metadata: Additional information associated with the audio file.
+                Default is an empty dictionary.
+            backend: I/O backend to use. Valid options include "ffmpeg", "sox", and "soundfile".
+                If None, a backend is automatically selected among available options.
+
+        Returns:
+            Audio: An instance containing the audio data and its corresponding metadata
         """
```
wilke0818, line 108: Update with a Raises section about the different errors that can occur?

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
         if not os.path.exists(filepath):
             raise FileNotFoundError(f"File {filepath} does not exist.")
 
-        array, sampling_rate = torchaudio.load(filepath)
+        if offset_in_sec < 0:
+            raise ValueError("Offset must be a non-negative value.")
+        if duration_in_sec < -1:
```
wilke0818: so this should be something akin to:
`duration_in_sec == -1 or duration_in_sec > 0` (presumably they can't create a 0 second audio

wilke0818: the issue with the current implementation is any value between -1 and 0 is valid

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+        if frame_offset > total_frames:
+            raise ValueError(
+                f"Offset ({offset_in_sec} seconds) exceeds the duration of the audio file ({filepath}): "
+                f"{total_frames / sampling_rate} seconds."
+            )
+
+        num_frames = int(duration_in_sec * sampling_rate) if duration_in_sec > 0 else -1
+
+        # Ensure num_frames are within bounds
+        if num_frames > 0:
```
wilke0818, line 148: Technically this if statement is not necessary though it is useful for readability, if num_frames is not greater than 0 then the min will still return -1

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+
+        return cls(
+            waveform=array, sampling_rate=sampling_rate, orig_path_or_id=filepath, metadata=metadata if metadata else {}
+        )
+
+    @classmethod
+    def from_stream(
+        cls,
+        stream_source: Union[str, os.PathLike, bytes],
+        chunk_size: int = 4096,
```
wilke0818: fine with this but another option is to have chunks be defined in seconds and then internally convert based off of sampling rate

src/tests/audio/data_structures/audio_test.py
```diff
@@ -17,19 +17,124 @@
...
+    assert audio.sampling_rate == 48000
+
+
+@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not installed.")
+def test_audio_creation_with_offset() -> None:
+    """Tests loading audio with a positive offset."""
+    audio = Audio.from_filepath(MONO_AUDIO_PATH, offset_in_sec=1.0)
+    assert audio is not None
+    assert audio.waveform is not None
+    assert audio.waveform.shape[1] > 0
```
wilke0818, line 50: if we want to be certain this worked as expected, we should do a check that the audio is the same as offset=0 where we index `audio.waveform[sampling_rate:,:]` or even just check the shape

src/tests/audio/data_structures/audio_test.py
```diff
@@ -17,19 +17,124 @@
...
+        Audio.from_filepath(MONO_AUDIO_PATH, offset_in_sec=file_duration + 1.0)
+
+
+@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not installed.")
+def test_audio_creation_with_duration() -> None:
+    """Tests loading a specific duration of an audio file."""
+    audio = Audio.from_filepath(MONO_AUDIO_PATH, duration_in_sec=2.0)
+    assert audio is not None
+    assert audio.waveform is not None
+    assert audio.waveform.shape[1] > 0
```
wilke0818, line 71: this one is actually very easy to check that the duration is equal to 2 sampling rates

src/tests/audio/data_structures/audio_test.py
```diff
@@ -17,19 +17,124 @@
...
+def test_audio_creation_negative_duration() -> None:
+    """Tests that a negative duration (except -1) raises an error."""
+    with pytest.raises(ValueError, match="Duration must be -1 .* or a positive value"):
+        Audio.from_filepath(MONO_AUDIO_PATH, duration_in_sec=-2.0)
+
+
+@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not installed.")
+def test_audio_creation_full_duration() -> None:
+    """Tests loading the full audio file with duration=-1."""
+    audio = Audio.from_filepath(MONO_AUDIO_PATH, duration_in_sec=-1)
```
wilke0818, line 104: If -1 is the default then this should already be handled but better to over test than under test

src/tests/audio/data_structures/audio_test.py
```diff
@@ -17,19 +17,124 @@
...
+    assert audio is not None
+    assert audio.waveform is not None
+    assert audio.waveform.shape[1] > 0
```
wilke0818, line 107: again probably best to check that the shape is equal to torchaudio.info().num_frames or whatever that value was

src/tests/audio/data_structures/audio_test.py
```diff
@@ -50,6 +155,20 @@ def test_audio_creation(audio_fixture: str, audio_path: str, request: pytest.Fix
...
 
 
+@pytest.mark.skipif(not SOUNDFILE_AVAILABLE, reason="soundfile is not installed.")
+@pytest.mark.parametrize(
+    "audio_path",
+    [MONO_AUDIO_PATH, STEREO_AUDIO_PATH],
+)
+def test_audio_stream(audio_path: str) -> None:
+    """Tests mono and stereo audio creation from stream."""
+    audio_chunks = Audio.from_stream(audio_path)
```
wilke0818, line 165: could be dumb but we might consider checking that each audio from the stream matches an audio of the same length/time period using from_filepath. This would mostly help also check that using soundfile isn't giving differing results from torchaudio

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+        metadata: Optional[Dict] = None,
+    ) -> Generator["Audio", None, None]:
+        """Yields Audio objects from a live stream (file, stdin, etc.).
+
+        Args:
+            stream_source: Filepath or input stream.
+            chunk_size: Number of audio frames per chunk to read.
+            metadata: Additional metadata associated with the audio.
```
wilke0818: A stream also could theoretically have a starting offset and duration though that might be too much if soundfile isn't natively supporting it

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
 
-        return cls(waveform=array, sampling_rate=sampling_rate, orig_path_or_id=filepath, metadata=metadata)
+        if isinstance(stream_source, os.PathLike) and not os.path.exists(stream_source):
+            raise FileNotFoundError(f"File {stream_source} does not exist.")
+
+        # Open the audio stream
+        with sf.SoundFile(stream_source, "r") as audio_file:
+            sampling_rate = audio_file.samplerate
+
+            while True:
```
wilke0818, line 200: should we offer padding here as well?

wilke0818: by here I mean a user using this function, realistically though padding could be created as part of the Audio constructor

wilke0818 (changes requested):
Looks pretty good. Most things are suggestions though we should make sure CONTRIBUTING.md is correct and that duration can be set (-1,0] even though under the hood I think the way you've programmed it it should be fine

wilke0818:
we should also plan to make similar changes to the video module eventually

fabiocat93:
> we should also plan to make similar changes to the video module eventually

good point. I haven't worked with video data yet recently, hence didn't have the chance to implement anything in that space. but maybe @brukew can contribute on that

CONTRIBUTING.md
fabiocat93: pls, see number 12

fabiocat93: lol

CONTRIBUTING.md
fabiocat93: yup. I realized that you need to setup those variables just if you are the maintainer of the project. If you open a PR vs the main branch, the repo+maintener's variables will be used (or at least, should)

src/senselab/audio/data_structures/audio.py
fabiocat93: it simply follow the same logic as in `torchaudio`

src/senselab/audio/data_structures/audio.py
fabiocat93: What's the advantage of that, though?

src/senselab/audio/data_structures/audio.py
fabiocat93: we are not really doing it much all over the package lol. I will probably release a patch in the future fixing it all (and maybe the entire documentation)

src/senselab/audio/data_structures/audio.py
fabiocat93: this requires you to know what's the sampling rate though. I didn't want to give it for granted

src/senselab/audio/data_structures/audio.py
fabiocat93: true. I didn't think about that... future work?

src/senselab/audio/data_structures/audio.py
fabiocat93: I see what you are saying. I am not sure how useful it is (I see more useful having an audio object respecting the original duration of the audio). people can always do padding using our augmentation suite afterwards

src/tests/audio/data_structures/audio_test.py
fabiocat93: yup, I haven't done any actual checks on the expected values. I honestly don't have the time for that, but I am happy to delegate if anyone wants to contribute

src/tests/audio/data_structures/audio_test.py
fabiocat93: fair. same as above, i don't have time lol but I agree it's a great suggestion and should be done

fabiocat93:
thank you, @wilke0818 . I have addressed your comments but made only some changes. unfortunately, I won't have much time to dedicate to this in the coming future. if you feel this is mergeable, feel free to accept it. otherwise, we can keep it open for now and wait until some good person will want to help :)

src/senselab/audio/data_structures/audio.py
wilke0818: fair point

src/senselab/audio/data_structures/audio.py
wilke0818: most likely they might not be supporting it natively so future work

CONTRIBUTING.md
wilke0818: ah I think I misunderstood that every time I read it lol

src/senselab/audio/data_structures/audio.py
wilke0818: It only requires us the developers to know it (it would be like saying chunk_in_seconds) but also I know it seems there is another function already existing that does a sliding window that can do this

wilke0818 (approved):
Maybe leave it open for a day if @900miles or @ibevers have comments but I also understand that you don't have time for other changes

src/senselab/audio/data_structures/audio.py
fabiocat93: the difference is that in .from_filepath we can programmatically retrieve the sampling rate, but this capability is not available when reading from a stream (from what I know)

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +108,82 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+                "Please install senselab audio dependencies using `pip install senselab['audio']`."
+            )
 
-        return cls(waveform=array, sampling_rate=sampling_rate, orig_path_or_id=filepath, metadata=metadata)
+        if isinstance(stream_source, os.PathLike) and not os.path.exists(stream_source):
+            raise FileNotFoundError(f"File {stream_source} does not exist.")
+
+        # Open the audio stream
+        with sf.SoundFile(stream_source, "r") as audio_file:
+            sampling_rate = audio_file.samplerate
```
ibevers, line 196: @fabiocat93 Aren't you retrieving the sample rate of a stream programmatically here?

ibevers (approved):
Overall, looks useful:)

src/senselab/audio/data_structures/audio.py
fabiocat93: you are right!!! I got confused with an older torchaudio implementation which allowed streams from URLs (not supported in this version), but didn't have access to the sample rate. thank you @ibevers !

ibevers: No problem!

wilke0818:
think this is ready for merging

PR #279, opened 2025-03-14
Add 20 AQMs to bioacoustic qc

ibevers:
## Description

Depends on: 
https://github.com/sensein/senselab/pull/278

AQMs:
- [x] proportion of silence
- [x] silence at beginning
- [x] silence at end
- [x] amplitude headroom
- [x] amplitude toeroom
- [ ] peak SNR
- [ ] percentage clipping
- [ ] clipping

## Types of changes
- [ ] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My code follows the code style of this project.

PR #287, opened 2025-03-20
286 Task Add 20 Audio Quality Metrics

### LINKED ISSUE #286: Task: Add 20 AQMs
### Description

Adds 20 AQMs to main that are independent of the Bioacoustic QC feature branch as requested by @satra in Slack. Uses cherry-picked commits from: https://github.com/sensein/senselab/pull/279 

### Tasks

AQMs:
- [x] proportion of silence
- [x] silence at beginning
- [x] silence at end
- [x] amplitude headroom
- [x] spectral grating SNR
- [x] peak SNR (not sure about implementation)
- [x] proportion clipped
- [x] clipping present
- [x] amplitude modulation depth
- [x] root mean square energy
- [x] zero crossing rate
- [x] signal variance
- [x] dynamic range
- [x] mean absolute deviation
- [x] shannon entropy of amplitude
- [x] crest factor
- [x] amplitude skew
- [x] amplitude kurtosis
- [x] amplitude IQR
- [x] phase correlation

### Freeform Notes

_No response_

ibevers:
## Description
Depends on:
https://github.com/sensein/senselab/pull/353

Adds 20 AQMs to main that are independent of the Bioacoustic QC feature branch as requested by @satra in Slack. Uses cherry-picked commits from: https://github.com/sensein/senselab/pull/279 

AQMs:
- [x] proportion of silence
- [x] silence at beginning
- [x] silence at end
- [x] amplitude headroom
- [x] spectral grating SNR
- [x] peak SNR (not sure about implementation)
- [x] proportion clipped
- [x] clipping present
- [x] amplitude modulation depth
- [x] root mean square energy
- [x] zero crossing rate
- [x] signal variance
- [x] dynamic range
- [x] mean absolute deviation
- [x] shannon entropy of amplitude
- [x] crest factor
- [x] amplitude skew
- [x] amplitude kurtosis
- [x] amplitude IQR
- [x] phase correlation

## Types of changes

## Checklist:


ibevers:
@fabiocat93 @satra Can we merge this?

ibevers:
Planning to use for the Bridge2AI feature extraction and would rather not use the task branch

ibevers:
@fabiocat93 @satra Do you have any feedback on this? Is it mergable?

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+
+    Args:
+        audio (Audio): The SenseLab Audio object.
+        silence_threshold (float): Amplitude below which a sample is silent.
+
+    Returns:
+        float: Proportion of silence at the start.
+    """
+    waveform = audio.waveform
+    assert waveform.ndim == 2, "Expected waveform shape (num_channels, num_samples)"
```
satra: is this assertion necessary? i.e. shouldn't this be a function of a valid audio object?

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+
+    silent_samples = (waveform.abs() < silence_threshold).sum().item()
+    return silent_samples / waveform.numel()
+
+
+def proportion_silence_at_beginning_metric(audio: Audio, silence_threshold: float = 0.01) -> float:
+    """Calculates the proportion of silence at the start.
```
satra, line 29: is there a reason why this proportion is a useful metric?

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    total_samples = waveform.size(-1)
+
+    non_silent_indices = torch.where(~all_channels_silent)[0]
+    if len(non_silent_indices) == 0:
+        return 1.0  # Entire audio is silent
+
+    return non_silent_indices[0].item() / total_samples
+
+
+def proportion_silence_at_end_metric(audio: Audio, silence_threshold: float = 0.01) -> float:
```
satra, line 50: same question here about utility of the metric?

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    max_amp = audio.waveform.max().item()
+    min_amp = audio.waveform.min().item()
+
+    if max_amp > 1.0:
+        raise ValueError(f"Audio contains samples over 1.0. Max amplitude = {max_amp:.4f}")
+    if min_amp < -1.0:
+        raise ValueError(f"Audio contains samples under -1.0. Min amplitude = {min_amp:.4f}")
+
+    pos_headroom = 1.0 - max_amp
+    neg_headroom = 1.0 + min_amp
```
satra: all of these assume that the representation is in floating point and limited between +1 and -1. it may be useful to ensure that in the audio object. most audio loaders do this, but sometimes it's much more efficient to store a int/unit sequence, which is how audio is often acquired. 

for purpose of this function, we should assume that the audio object comes with some property or guarantee of this.  fine for now, but some of these checks should be moved to audio object, and it should have a computed property for doing such normalization or expansion. 

in addition, i would suggest adding max/min and other operations to the audio object, as it can leverage streaming when available for large audio files.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+        float: Estimated segmental SNR in dB.
+    """
+    waveform: np.ndarray | torch.Tensor = audio.waveform
+    if isinstance(waveform, torch.Tensor):
+        waveform = waveform.numpy()
+
+    if waveform.shape[0] > 1:
+        waveform = np.mean(waveform, axis=0)
+
+    stft: np.ndarray = np.abs(librosa.stft(waveform, n_fft=frame_length, hop_length=hop_length))
```
satra: can we be consistent about which fft routines we use throughout senselab? i believe we use torchaudio, and that senselab function should be used here instead of reimplementing it.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    pos_headroom = 1.0 - max_amp
+    neg_headroom = 1.0 + min_amp
+
+    return min(pos_headroom, neg_headroom)
+
+
+def spectral_gating_snr_metric(
+    audio: Audio, frame_length: int = 2048, hop_length: int = 512, percentile: int = 10
+) -> float:
+    """Computes segmental SNR using the spectral gating approach.
```
satra, line 110: please explain what this is and put a pointer to a relevant paper.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    else:
+        waveform = audio.waveform
+        assert waveform.ndim == 2, "Expected waveform shape (num_channels, num_samples)"
+        waveform = waveform.abs()
+
+        clipped_proportion_by_channel = []
+        for channel in waveform:
+            max_val = torch.max(channel)
+            clipped_samples = torch.isclose(channel, max_val).sum().item()
+            clipped_proportion_by_channel.append(clipped_samples / channel.numel())
```
satra: this can be done without the for loop

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+
+        return float(np.mean(clipped_proportion_by_channel))
+
+
+def clipping_present_metric(audio: Audio, plateau_length: int = 5) -> bool:
+    """Detects clipping by looking for flat plateaus of high-amplitude samples.
```
satra: what's the rationale for 5 as default? for speech, clipping is often not plateaus but spikes of the large volume velocity of the initial glottal pulse. i would suggest default being 1.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+                count += 1
+            else:
+                count = 0
+            if count >= plateau_length:
+                return True
+    return False
+
+
+def amplitude_modulation_depth_metric(audio: Audio) -> float:
+    """Calculates the amplitude modulation depth of an audio signal.
```
satra:
```suggestion
    """Calculates the amplitude modulation depth of an audio signal using a Hilbert transform.
```

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+        audio (Audio): The SenseLab Audio object.
+
+    Returns:
+        float: RMS energy averaged across channels.
+    """
+    waveform = audio.waveform
+    assert waveform.ndim == 2, "Expected waveform shape (num_channels, num_samples)"
+
+    rms_per_channel = torch.sqrt(torch.mean(waveform**2, dim=1))
+    return float(torch.mean(rms_per_channel))
```
satra: if there are multiple channels, we should leave the output with shape `(n_channels,)`. this is a general comment for all routines.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
```diff
@@ -0,0 +1,511 @@
...
+    waveform = audio.waveform
+    assert waveform.ndim == 2, "Expected waveform shape (num_channels, num_samples)"
+
+    # Compute the mean absolute amplitude across all samples and channels
+    mean_abs = torch.mean(waveform.abs())
+    return float(mean_abs)
+
+
+def mean_absolute_deviation_metric(audio: Audio) -> float:
+    """Calculates the mean absolute deviation (MAD) of the audio signal.
```
satra: perhaps make this an absolute deviation metric with `mean` or `median` as a parameter?

satra (commented):
Overall this looks like a great collection. I left some specific comments. Should be addressable quickly to update the PR relative to those and then merge.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: It could be a good indication of whether a participant was focused on completing the task after hitting the record button. If it is large, they might have got distracted after hitting record. If it is zero, it might indicate they started the task before hitting record.

ibevers: Absolute duration might be better than proportion

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: >is this assertion necessary?

Nope! Removed.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: Similar to the first one, it could indicate that the participant got distracted part way through recording or wasn't able to fully complete the task for some reason.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: >all of these assume that the representation is in floating point and limited between +1 and -1. it may be useful to ensure that in the audio object. 

Created a PR: https://github.com/sensein/senselab/pull/324

>but some of these checks should be moved to audio object, 

Removing this because it is already guaranteed by the Audio object:
```
    if audio.waveform.dtype != torch.float32:
        raise TypeError(f"Expected waveform dtype torch.float32, but got {audio.waveform.dtype}")
```

ibevers: >  i would suggest adding max/min and other operations to the audio object, as it can leverage streaming when available for large audio files.

Created an issue: https://github.com/sensein/senselab/issues/325

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: Updated

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: Due to the way this is calculated, if the plateau_length is set to 1, it will always return True. It will still detect clipping of a single sample if that sample is 1 or -1.

ibevers: I changed the variable naming of plateau_length to max_value_count since it is looking at the number of maximum valued samples regardless of whether they are consecutive or not.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: Certainly! Fixed

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: Updated!

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: To make sure the naming is clear and descriptive, I added median as a separate function. This also is consistent with the API and naming convention throughout the module.

src/senselab/audio/tasks/bioacoustic_qc/metrics.py
ibevers: @satra I updated the docstring with more explanation and a reference:
https://arxiv.org/abs/2412.17851

ibevers:
@satra I've updated based on your feedback! Thank you for your review!

Issue #261, opened 2025-03-04
Task: senselab v1


fabiocat93:
### Description

Restructuring and improving utility, robustness and performance of senselab before releasing v1.0.0

### Tasks

- [x] Restructure senselab into audio, text, and video submodules (there is already an open PR for this: https://github.com/sensein/senselab/pull/258)
- [x] Implement lazy-loading for audio and video 
- [x] Improve performance with pydra
- [ ] Extend audio feature support (optional) 
- [ ] Add deep face functionalities (optional) 
- [ ] Implement a CLI for audio feature extraction (optional)

fabiocat93:
Closing this issue as priorities and schedule have changed. Open points are still valid, but won't be necessarily part of v1 release

PR #288, opened 2025-03-20
Lazy load audios

wilke0818:
## Description
Lazy loads audios when their waveform is attempted to be accessed

## Related Issue(s)
#261 

## Motivation and Context
Even with Fabio's changes, there is still inefficiency when creating many Audios at once prior to using them such that they are all loaded into memory at once. This defers their loading until the `waveform` attribute needs to be accessed.

## How Has This Been Tested?
- New unit tests
- Existing unit tests work showing that functionality is backwards compatible

## Types of changes

## Checklist:


wilke0818:
TODO: update documentation to match new strategy

src/senselab/audio/data_structures/audio.py
```diff
@@ -66,13 +120,46 @@ def convert_to_tensor(
...
+        # Load the specified portion of the audio
+        array, _ = torchaudio.load(
+            filepath, frame_offset=self._frame_offset, num_frames=self._num_frames, backend=self._backend
+        )
+
+        return array
+
     @classmethod
-    def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Audio":
+    def from_filepath(
```
fabiocat93: @wilke0818 at this point, do we still need from_filepath?

src/senselab/audio/data_structures/audio.py
```diff
@@ -83,9 +170,86 @@ def from_filepath(cls, filepath: str | os.PathLike, metadata: Dict = {}) -> "Aud
...
+            orig_path_or_id=filepath,
+            metadata=metadata if metadata else {},
+            filepath=filepath,
+            num_frames=num_frames,
+            frame_offset=frame_offset,
+            backend=backend,
+        )
+
+    @classmethod
+    def from_stream(
```
fabiocat93: do we want to remove from_stream and handle it all in the class constructor?

src/senselab/audio/data_structures/audio.py
```diff
@@ -40,24 +47,71 @@ class Audio(BaseModel):
...
-        cls, v: Union[List[float], List[List[float]], np.ndarray, torch.Tensor], _: ValidationInfo
-    ) -> torch.Tensor:
+    def __init__(
+        self,
+        **data: Any,  # noqa: ANN401
+    ) -> None:
+        """Custom init to handle waveform as an optional user-provided input."""
+        waveform = data.pop("waveform", None)  # Extract waveform if provided
+        filepath = data.pop("filepath", None)
+        frame_offset = data.pop("frame_offset", None)
```
fabiocat93: from previous convs, I guess we will need to decide if we want the users to reason in terms of frames or seconds

fabiocat93 (commented):
thank you @wilke0818 ! I think this represent a good enhancement. I have left a couple of questions to discuss

src/senselab/audio/data_structures/audio.py
wilke0818: Yeah makes sense. These specific parameters are more there for maintaining the logic you implemented in a way that still allows to lazy-load

src/senselab/audio/data_structures/audio.py
wilke0818: hmmm we could. another option might be to subclass Audio into AudioStream which would allow it to be passed into current functionality but might help from over-cluttering Audio. If we go with the constructor, we would need additional arguments which might get messy but still manageable. To be honest with lazy-loading, no matter what we do, the code will likely become a bit messier under the hood

src/senselab/audio/data_structures/audio.py
wilke0818: we could remove it but we would then be moving further away from the way Pydantic is meant to operate I think (my current implementation certainly already is). I guess the way I was seeing it is that Audio should be able to be very generally defined as the waveform and the sampling rate associated with it which was what I was trying to maintain. Then getting audio from a file is somewhat secondary (even though it is our primary use case). In the end whatever method we choose is fine, though we should probably restructure some stuff if we do want to lean into just using the constructor

src/senselab/audio/data_structures/audio.py
fabiocat93: Both options are ok with me. My vote would probably go to one unique Audio class, to make it easier for the user, despite the code may be messier under the hood

fabiocat93:
@wilke0818 let me know what you think. Thanks!

src/senselab/audio/tasks/voice_cloning/coqui.py
```diff
@@ -96,15 +96,15 @@ def clone_voices(
...
+                    f"Source ({source_audio.generate_id()}) shape: {source_audio.waveform.shape}, "
+                    f"Target ({target_audio.generate_id()}) shape: {target_audio.waveform.shape}"
                 )
 
             if source_audio.sampling_rate != expected_sample_rate or target_audio.sampling_rate != expected_sample_rate:
                 raise ValueError(
                     f"Expected input sample rate {expected_sample_rate}, but got "
-                    f"{source_audio.sampling_rate} (source = {source_audio.orig_path_or_id}) "
-                    f"and {target_audio.sampling_rate} (target = {target_audio.orig_path_or_id})."
+                    f"{source_audio.sampling_rate} (source = {source_audio.generate_id()}) "
```
wilke0818: I don't know how informative this is now since the genereate_id will just be a UUID even if the data is from a file.

src/senselab/utils/data_structures/dataset.py
```diff
@@ -295,22 +293,19 @@ def _get_dict_representation(self) -> Dict:
...
             video_fps_data.append(video.frame_rate)
-            video_path_data.append(video.generate_path())
             video_metadata.append(video.metadata.copy())
             video_audio_data.append(
                 None
                 if not video.audio
                 else {
                     "array": video.audio.waveform.T.to(torch.float32).numpy(),
                     "sampling_rate": video.audio.sampling_rate,
-                    "path": video.audio.generate_path(),
```
wilke0818, line 306: presuming this change is fine if the test passed but I did think that HuggingFace's Audio class needed a path variable?

src/senselab/audio/tasks/voice_cloning/coqui.py
fabiocat93: it gives a way to find the audio though. generate_id can be changed if we want it to point to the audio path though

src/senselab/utils/data_structures/dataset.py
fabiocat93: they have a path field in the doc, but tests pass even without it. idk

src/senselab/audio/tasks/voice_cloning/coqui.py
wilke0818: Hmmm. I mean I just don't see it being as useful for finding the audio, especially say if you're running this function over many audios and one source audio fails, this doesn't tell you which one it might be. That's my thinking at least. Do you think if you load an audio from a file, that the filepath can make a unique identifier? If not then we should keep this, but if so then we can just use the filename that is passed in.

src/senselab/utils/data_structures/dataset.py
wilke0818: Sounds good if it passes.

src/senselab/audio/tasks/voice_cloning/coqui.py
fabiocat93: I understand. But then what do you do when for example you create a sliding window on the audio? or when you augment an audio? do you keep the original filepath as id or do you change it? As a quick solution here, how about adding the index of the audio as part of the error message?

wilke0818: I think the index makes sense. Sliding windows I think would be fine if it's creating a new Audio object since we're likely creating it through waveform. Augmentation is a weirder one where I would say you don't keep the original filename or if you do we should put it into the metadata. We can make such choices later.

fabiocat93: agree

fabiocat93: I have included the filepath to the error messages (if present) and merged

PR #290, opened 2025-03-24
Added filter by wordcount function

### LINKED ISSUE #289: Task: Filtering audio samples by whether or not the number of words falls within a certain range
### Description

I think that people with certain mental health conditions might talk too little in the audio sample for us to be able to generate statistically meaningful data about some features, including mean length of utterance. So, I propose that we include a function to count how many words the speaker says and then exclude samples that fall outside of a range specified as a function argument.

### Tasks

- [ ] Task 1

### Freeform Notes

_No response_

Veronika271:
## Description
Wrote a method that counts the number of words in a speech sample and then returns True if the number of words falls within a specified range, and False otherwise.
## Related Issue(s)
#289 

## Motivation and Context
Sometimes people talk very little, especially when they are struggling with certain mood disorders. I'm worried that if we don't get a sufficiently long voice sample, it will be hard to draw any conclusions due to small sample size.

## How Has This Been Tested?
I tested it locally on a toy example and showed it to Dr. Catania!


ibevers:
@Veronika271 can you add a unit test for this?

src/senselab/audio/tasks/filter_by_word_count.py
```diff
@@ -0,0 +1,28 @@
...
+    audio (Audio): the audio sample
+    range (tuple of two ints): the possible range of wordcount, from minimum to maximum
+    model: SenselabModel
+
+    Returns:
+    True if the number of words in the audio falls within the range, and false otherwise (Boolean)
+    """
+    transcript = transcribe_audios([audio], model)[0]
+    num_words = len(transcript.text.split())
+    if num_words < range[0] or num_words > range[-1]:
```
ibevers: You could make this more concise by returning an evaluated boolean directly

src/senselab/audio/tasks/filter_by_word_count.py
```diff
@@ -0,0 +1,28 @@
+from typing import Any, List, Optional
+
+import os
+
+from senselab.audio.data_structures import Audio
+from senselab.audio.tasks.speech_to_text.huggingface import HuggingFaceASR
```
ibevers: Unused import

ibevers: Make sure to remove other unused imports

src/senselab/audio/tasks/filter_by_word_count.py
```diff
@@ -0,0 +1,28 @@
...
+from senselab.audio.tasks.speech_to_text.api import transcribe_audios
+from senselab.utils.data_structures import DeviceType, HFModel, Language, ScriptLine, SenselabModel
+import torchaudio
+from senselab.audio.tasks.preprocessing import resample_audios, downmix_audios_to_mono
+
+def is_num_words_in_range(audio, range, model):
```
ibevers: Needs type annotations. Make sure you install pre-commit and this passes. See CONTRIBUTING.md for details

ibevers:
> I tested it locally on a toy example and showed it to Dr. Catania!

Nice!:) @Veronika271 feel free to call him Fabio--he prefers that as far as I know:)

ibevers (changes requested):
Overall, this looks like a great first addition! There's a few little things to do, but I like it overall🙂🌱

ibevers:
@Veronika271 https://github.com/sensein/senselab/blob/a41ffc0dbbda625dc05f7c25c01e66bac9fd9066/src/tests/audio/tasks/speech_enhancement_test.py#L71

In your unit test, you can pass an audio sample like this--we have it configured this way for convenience

ibevers:
@Veronika271 see all the audio samples you can use here:
[`src/tests/audio/conftest.py`](https://github.com/sensein/senselab/blob/a41ffc0dbbda625dc05f7c25c01e66bac9fd9066/src/tests/audio/conftest.py#L4)

src/senselab/audio/tasks/filter_by_word_count.py
```diff
@@ -0,0 +1,25 @@
...
+    Args:
+    audio (Audio): the audio sample
+    range (tuple of two ints): the possible range of wordcount, from minimum to maximum
+    model: SenselabModel
+
+    Returns:
+    True or False (bool): True if the number of words in the audio falls within the range, and False otherwise
+    """
+    transcript = transcribe_audios([audio], model)[0]
+    if transcript and transcript.text:
```
ibevers, line 24: I would expect it to return False iff the number of words in the audio falls outside the range. This could raise an error if there is no text in the transcript

ibevers (changes requested):
Thanks for adding unit tests:)

src/senselab/audio/tasks/filter_by_word_count.py
Veronika271: Thank you for explaining that! I just changed it to raise an error if the transcript is empty.

ibevers:
@Veronika271 I think this functionality would fit better within the `quality_control` module now that it exists. Would you mind doing these steps?
- [ ] Change base to `276-feature-implement-bioacoustic-quality-control-framework`
- [ ] Create a metric function that counts the number of words here: `src/senselab/audio/tasks/bioacoustic_qc/metrics.py`
- [ ] Create a check function that determines whether the number of words are in a range here: `src/senselab/audio/tasks/bioacoustic_qc/checks.py`

satra:
Closing — stale. Tracked by issue #289 under meta issue #474 (Audio Data Model).

PR #291, opened 2025-03-25
Fixed but where n_fft defaults to 1024

Evan8456:
## Description

Fixed a bug with spectrogram dimensions being generated with n_fft = 1024 instead of specified value.
## Related Issue(s)

https://github.com/sensein/b2aiprep/issues/135
## How Has This Been Tested?

Added test case where we input n_fft to a 400 and confirm the dimensions match.
## Types of changes

## Checklist:

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -16,7 +16,7 @@ repos:
   rev: v1.10.0
   hooks:
   - id: mypy
-    args: [--ignore-missing-imports]
+    args: [--ignore-missing-imports, --extra-checks]
     additional_dependencies:
     - pydantic>=1.10.4
     - types-requests
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -131,6 +131,7 @@ ignore_missing_imports = true
 plugins = [
   "pydantic.mypy"
 ]
+extra_checks = true
 
 [tool.ruff]
 exclude = [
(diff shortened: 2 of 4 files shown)


fabiocat93 (approved):
nice catch! Looks good to me, if tests pass

satra:
nice tracking @Evan8456.

@fabiocat93 - perhaps we can turn mypy into extra checks or stricter mode to see if this could be caught that way (https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-extra-checks)

fabiocat93:
> nice tracking @Evan8456.
> 
> @fabiocat93 - perhaps we can turn mypy into extra checks or stricter mode to see if this could be caught that way (https://mypy.readthedocs.io/en/stable/command_line.html#cmdoption-mypy-extra-checks)

I can do that. But can I ask you why do we want that?

satra:
> But can I ask you why do we want that?

this really seems like the kind of thing that should be caught by a static checker and it didn't. i didn't check if strict or any of the extra args would catch this situation. in general, stricter will result in more compliant code, but i don't know what constraints it would create for us.

ibevers:
Good catch! @Evan8456

ibevers:
> i don't know what constraints it would create for us.

Perhaps we should identify what these are first and weight the pros and cons

fabiocat93:
I have fixed the style issue that was preventing pre-commit hooks from passing, and I have added the `-- extra-checks` flag to mypy configs. This is expected to help us maintain cleaner code in general, but from some experiments, it doesn't seem that this could have helped with catching the issue

Issue #292, opened 2025-03-25
Bug: Issues running feature extractions on latest version of senselab


Evan8456:
### Description

Issues running feature extractions with the latest version of senselab. 

### Steps to Reproduce

Running the b2aiprep command prepare-bids causes a runtime error from senselab. Please see https://github.com/sensein/b2aiprep/issues/140 for context.

### Expected Results

feature extractions returned.

### Actual Results

```
Traceback (most recent call last):
  File "/home/evanng/mambaforge/bin/b2aiprep-cli", line 8, in <module>
    sys.exit(cli())
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/click/core.py", line 1157, in __call__
    return self.main(*args, **kwargs)
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/click/core.py", line 1078, in main
    rv = self.invoke(ctx)
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/click/core.py", line 1688, in invoke
    return _process_result(sub_ctx.command.invoke(sub_ctx))
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/click/core.py", line 1434, in invoke
    return ctx.invoke(self.callback, **ctx.params)
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/click/core.py", line 783, in invoke
    return __callback(*args, **kwargs)
  File "/home/evanng/Documents/b2aiprep_fork/b2aiprep/src/b2aiprep/commands.py", line 172, in prepare_bids
    extract_features_sequentially(
  File "/home/evanng/Documents/b2aiprep_fork/b2aiprep/src/b2aiprep/prepare/prepare.py", line 232, in extract_features_sequentially
    extract_single(audio_file["path"], transcription_model_size, with_sensitive)
  File "/home/evanng/Documents/b2aiprep_fork/b2aiprep/src/b2aiprep/prepare/prepare.py", line 127, in extract_single
    features = extract_features_from_audios(
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/senselab/audio/tasks/features_extraction/api.py", line 348, in extract_features_from_audios
    parselmouth_features = extract_praat_parselmouth_features_from_audios(audios=audios, **my_parselmouth)  # type: ignore
  File "/home/evanng/mambaforge/lib/python3.10/site-packages/senselab/audio/tasks/features_extraction/praat_parselmouth.py", line 1397, in extract_praat_parselmouth_features_from_audios
[...]

PR #293, opened 2025-03-25
fixed bug with praat

Evan8456:
## Description

Fixed issue with https://github.com/sensein/senselab/issues/292
## Related Issue(s)

https://github.com/sensein/b2aiprep/issues/140
## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
+++ b/src/senselab/audio/tasks/features_extraction/praat_parselmouth.py
@@ -24,7 +24,7 @@ except ModuleNotFoundError:
     PARSELMOUTH_AVAILABLE = False
 
 
-def get_sound(audio: Union[Path, Audio], sampling_rate: int = 16000) -> "parselmouth.Sound":
+def get_sound(audio: Union[Path, Audio], sampling_rate: int = 16000) -> parselmouth.Sound:
     """Get a sound object from a given audio file or Audio object.
 
     Args:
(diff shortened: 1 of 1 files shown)


fabiocat93:
@Evan8456 unfortunately we cannot assume `parselmouth` is installed.

Evan8456:
@fabiocat93 do you have any suggestions on how to resolve this issue if we can not assume `parcelmouth` is installed?

fabiocat93:
> @fabiocat93 do you have any suggestions on how to resolve this issue if we can not assume `parcelmouth` is installed?

One way may be to annotate the types of inputs and outputs with pydra. Another way is to define functions only if parcelmouth is available. I suspect this same problem would come up with other dependencies, and the solution should be applied everywhere in the code. One question, though: why doesn't this come up from the unit tests?

Evan8456:
Aren't certain unit tests skipped if parcelmouth isn't installed? I believe this is the case for others such as torch audio

fabiocat93:
> Aren't certain unit tests skipped if parcelmouth isn't installed? I believe this is the case for others such as torch audio

Some tests are skipped if parselmouth isn't installed. But when parselmouth is installed, all tests pass... We may need to add tests for `extract_praat_parselmouth_features_from_audios` I guess!

fabiocat93:
this may  be fixed as part of the migration to pydra v1 (https://github.com/sensein/senselab/issues/301)

fabiocat93:
this may  be fixed as part of the migration to pydra v1 (https://github.com/sensein/senselab/issues/301). I will close the PR for now since it doesn't provide a feasible solution to the problem. we can keep the discussion open in the related issue. thank you @Evan8456 !

PR #294, opened 2025-03-25
Enhanced stability for cca and cka

fabiocat93:
## Description
Attempt to improve cca and cka implementations. To compute CCA,  I added features' standardization plus QR decomposition. As to copute CKA, I added standardization and boosted numerical stability with regularization and clamping. The primary motivation for this PR is that with the previous implementation, we could obtain values > 1 (not theoretically grounded). 

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/utils/tasks/cca_cka.py
+++ b/src/senselab/utils/tasks/cca_cka.py
@@ -1,7 +1,6 @@
-"""This module is for computing CCA and CKA."""
+"""This module computes CCA and CKA."""
 
 from enum import Enum
-
 import torch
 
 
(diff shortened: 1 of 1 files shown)


src/senselab/utils/tasks/cca_cka.py
```diff
@@ -15,10 +15,15 @@ def compute_cca(features_x: torch.Tensor, features_y: torch.Tensor) -> float:
     Returns:
         float: The mean squared CCA correlations between X and Y.
     """
-    qx, _ = torch.linalg.qr(features_x)
-    qy, _ = torch.linalg.qr(features_y)
+    # Standardize inputs to unit variance (improves numerical stability)
+    features_x = (features_x - features_x.mean(0)) / (features_x.std(0) + 1e-8)
+    features_y = (features_y - features_y.mean(0)) / (features_y.std(0) + 1e-8)
```
satra: this will modify the input, which should be avoided. also may want a flag whether to do this or not.

src/senselab/utils/tasks/cca_cka.py
```diff
@@ -33,101 +38,76 @@ def compute_cka(
...
     Returns:
-        float: The value of CKA between X and Y.
+        float: CKA similarity ∈ [0,1] where 1 = identical representations.
+
+    Raises:
+        ValueError: If an unsupported kernel type is provided.
     """
+    # Standardize inputs to unit variance (critical for stability)
+    features_x = (features_x - features_x.mean(0)) / (features_x.std(0) + 1e-8)
+    features_y = (features_y - features_y.mean(0)) / (features_y.std(0) + 1e-8)
```
satra: same comment here.

src/senselab/utils/tasks/cca_cka.py
fabiocat93: and is it necessarily bad? from my tests, without standardization, this method is very unstable and breaks 9/10 times

fabiocat93: I am ok with having the flag. my point is: should we default to normalization=True?

satra: my point was more about modifying inputs in place (which is bad and should be avoided).
regarding the flag, one may have standardized it, so setting it to true as default is ok, but having the option to pass standardized features could be helpful.

satra (approved):

PR #308, opened 2025-04-11
removing timestamped_transcript workflow

fabiocat93:
I have removed the `timestamped_transcript` workflow since it doesn't pass the tests. Once the issues are fixed, we will integrate it again. We just should avoid having non-functioning code in the main branch


ibevers:
@fabiocat93 Makes sense! No problem!

ibevers (approved):

wilke0818 (approved):

PR #312, opened 2025-04-18
fixing install command to make it compatible with zsh, bash, and fish

### LINKED ISSUE #311: Bug: senselab['audio'] should be 'senselab[audio]' when installing with zsh
### Description

```
(senselab) isaacbevers@MacBook-Pro-87 ~ % pip install senselab['audio']

zsh: no matches found: senselab[audio]
```
Fixed by changing to `pip install 'senselab[audio]'`

To-do:
- [x] Update documentation to account for the above

### Steps to Reproduce

Follow pip install instructions in zsh

### Expected Results

Installs correctly

### Actual Results

Doesn't recognize command

### Additional Notes

_No response_

fabiocat93:
## Description
Update documentation to account installation compatibility across shells (at least zsh, bash, and fish)

## Related issues
https://github.com/sensein/senselab/issues/311

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,7 +46,7 @@ Welcome to ```senselab```! This is a Python package for streamlining the process
 Install this package via:
 
 ```sh
-pip install senselab['all']
+pip install 'senselab[all]'
 ```
 
 Or get the newest development version via:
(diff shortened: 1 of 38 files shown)


fabiocat93:
@ibevers ready for review

fabiocat93:
self-reviewed :( . I will merge this one so that we have this fix before the b2ai symposium

ibevers:
> self-reviewed :( . I will merge this one so that we have this fix before the b2ai symposium

Makes sense! In the future, if you can give me 2 working days of lead time, I will do my best to review in that timeframe

PR #322, opened 2025-04-30
297 Task Parallelize Bioacoustic QC

### LINKED ISSUE #297: Task: parallelize bioacoustic qc
### Description

Parallelize deriving AQMs and running checks probably using Pydra and LFApplier, as mentioned here: https://github.com/sensein/senselab/pull/278

### Tasks

- [ ] Design approach
- [ ] Implement
- [ ] Test

### Freeform Notes

_No response_

ibevers:
## Description
Depended on by: 
https://github.com/sensein/senselab/tree/326-task-calculate-metrics-over-windows

Updates bioacoustic QC to run nested evaluations efficiently for large datasets. Specifically:
- Streamlines the design by collecting a subtree's evaluations in a list and running them at once rather than recursively according to the structure of the taxonomy tree.
- Parallelizes execution with Pydra.
- Caches evaluations and avoids recomputing them if the evaluation process is interrupted.
- Uses a list of audio paths as input and lazily loads each audio file only once.
- Aggregates calculated evaluations into a readable CSV after the evaluations are run.

## Types of changes

## Checklist:


ibevers:
@satra @fabiocat93 Does this adequately address your design concerns from the first iteration?

ibevers:
@Rahul-Brito any thoughts on this?

ibevers:
@satra @fabiocat93 @Rahul-Brito Do you have any thoughts on this?

src/senselab/audio/tasks/bioacoustic_qc/__init__.py
```diff
@@ -1,21 +1,23 @@
...
-    run_taxonomy_subtree_checks_recursively,
-    taxonomy_subtree_to_pydra_workflow,
+    create_activity_to_evaluations,
+    evaluate_audio,
+    evaluate_batch,
+    run_evaluations,
+    subtree_to_evaluations,
 )
 
 __all__ = [
```
fabiocat93: I would recommend not using __all__ = [...] but the same approach as in the other submodules of the package. This is because pdoc (the tool we use for automatically generating the documentation from docstrings) doesn't work as expected otherwise

src/tests/audio/tasks/checks_test.py
fabiocat93, line 1: you test `get_metric` with and without a DataFrame, which is great. But I see no test for the case when the file name is not found in the df, which may occur if there's a mismatch or missing entry. I would consider adding that

fabiocat93, line 1: At line 83, what if the `filepath` is None?

src/senselab/audio/tasks/bioacoustic_qc/checks.py
fabiocat93, line 1: If I undertsand it right, the  `get_metric` function introduces logic to avoid recomputing metrics already present in the df. I like this in concept, but why not manage this using the cache mechanism already built in in Pydra as we do for the other audio features?

fabiocat93: The `extract_features_from_audios` function **tries to** cleanly separate feature extractor configuration, execution, and result aggregation. Consider aligning `get_metric` with the structure of `extract_features_from_audios`. You could define a `extract_metrics_from_audios` function that:

- Accepts a list of audios
- Accepts a list of metrics
- Returns a list of dicts
- Uses the same plugin/config pattern as the rest of the package

Having different logics inside of the same package may confuse future developers or users

fabiocat93: As a minor thing: `get_metric` uses `metric_function.__name__` as the key. This assumes that all metric functions have unique names but if two different metrics are implemented with the same function name (e.g., via wrapper functions), this could siilently overwrite or misread cached values. You may want to use a namespacing approach like `f"{metric_function.__module__}.{metric_function.__name__}"`

fabiocat93: Please treat filepaths as Path-like objects instead of strings, unless you do that for a reason

fabiocat93: The current setup works one metric per file. If I want 10 metrics across 5000 files, that’s 50,000 function calls... Integrating `pydra` would help optimize execution time

fabiocat93 (commented):
Hi @ibevers , nice work! For now, I have looked at the general structure and I haven't gone through every single metric, which is something that may be worth doing during a coming biometrics meeting or even in a bunch of 1on1 meetings we can schedule. I have left major and minor comments here and there

ibevers:
> Hi @ibevers , nice work! For now, I have looked at the general structure and I haven't gone through every single metric, which is something that may be worth doing during a coming biometrics meeting or even in a bunch of 1on1 meetings we can schedule. I have left major and minor comments here and there

Hi @fabiocat93, thank you for your detailed feedback! I will work through this, and I'm definitely happy to workshop the metrics. The metrics have some updates in this PR (as well as the series of PRs that build on top of this one):
https://github.com/sensein/senselab/pull/287

src/senselab/audio/tasks/bioacoustic_qc/__init__.py
ibevers: Removed✅

src/tests/audio/tasks/checks_test.py
ibevers: Added:)

ibevers: Updated get_metric to be robust to this

ibevers:
@fabiocat93 In the chain of QC PRs, some things in the earlier PRs got changed in the later ones. I don't want to waste your time on reviewing things that got changed. Also, it will create merge conflicts when merging up the chain. However, if you only review the final changes, it will be a lot to review. Do you have a preference for which approach to take? 

I'm thinking that only reviewing the latest one will be more efficient overall, even though it's still cumbersome.

fabiocat93:
> I'm thinking that only reviewing the latest one will be more efficient overall, even though it's still cumbersome.

sounds good. Maybe I will need some guidance during the process at that stage and will ask your help

ibevers:
>sounds good. Maybe I will need some guidance during the process at that stage and will ask your help

@fabiocat93 Okay! I'm happy to walk through it with then. That should give some ideas about what would be helpful to include in a tutorial too

PR #333, opened 2025-05-14
Change pyav to av

ibevers:
## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [x] My code follows the code style of this project.

Issue #321, opened 2025-04-30
Bug: common ci bug: issues with torbi (1.1.0)


ibevers:
### Description

For at least two PRs, these tests failed in recent CI runs with the same error:

<img width="740" alt="Image" src="https://github.com/user-attachments/assets/d5685817-e745-4a09-8c9d-8aa16b9aee51" />

CI runs that failed: 
https://github.com/sensein/senselab/actions/runs/14721752623/job/41317052710
https://github.com/sensein/senselab/actions/runs/14716179271/job/41300493162

Shared error message:
```
...
Note: This error originates from the build backend, and is likely not a problem with poetry but one of the following issues with torbi (1.1.0)

  - not supporting PEP 517 builds
  - not specifying PEP 517 build requirements correctly
  - the build requirements are incompatible with your operating system or Python version
  - the build requirements are missing system dependencies (eg: compilers, libraries, headers).

You can verify this by running pip wheel --no-cache-dir --use-pep517 "torbi (==1.1.0)".
```

In https://github.com/sensein/senselab/pull/215 the CI failure began after a commit that only updated some links in  a docs.md, which shouldn't result in a CI failure.

### Steps to Reproduce

Given that the two PRs that show the error are touching different parts of the code, I'm guessing this same CI failure will happen for any new PRs, although I don't know for sure.

### Expected Results

Previously passing CI tests pass after change to doc.md.

### Actual Results

The CI runs in the description fail with the error message in the description. 

### Additional Notes

This may have been caused by an adjustment to the CI configuration that was not included in the code.

fabiocat93:
hi @ibevers , what do you think may be a solution to this? have you tried installing `torbi` by any chance?

ibevers:
> hi [@ibevers](https://github.com/ibevers) , what do you think may be a solution to this? have you tried installing `torbi` by any chance?

 - It looks like torbi depends on pytorch:
>dependencies = [
    "numpy", "torch", "torchutil", "yapecs", "tqdm"
]
See: https://github.com/maxrmorrison/torbi/blob/main/pyproject.toml


- I see `torch==2.7` was released 7 days ago:
https://pytorch.org/blog/pytorch-2-7/


- In the failing CI run, `torch==2.7`  was installed:
<img width="358" alt="Image" src="https://github.com/user-attachments/assets/856a607b-6152-4667-8e2c-64c1fd0472b2" />

See: https://github.com/sensein/senselab/actions/runs/14716179271/job/41300493162


[...]

PR #335, opened 2025-05-14
334 Task Store Filepath if Provided with Waveform

### LINKED ISSUE #334: Task: store filepath if provided with waveform
### Description

Currently, Audio stores None for `filepath` if it is provided with a waveform. However, I think both should be stored. 

### Tasks

- [x] Update Audio to store filepath if provided with waveform

### Freeform Notes

_No response_

ibevers:
## Description

Ensures that a filepath is stored in an `Audio` object if both `filepath` and `waveform` arguments are passed when an `Audio` is initialized.

## Types of changes

## Checklist:

### DIFF
--- a/src/senselab/audio/data_structures/audio.py
+++ b/src/senselab/audio/data_structures/audio.py
@@ -80,6 +80,8 @@ class Audio(BaseModel):
                 raise ValueError("When a waveform is provided, a sampling_rate must also be supplied.")
             self._waveform = self.convert_to_tensor(waveform)
             self._sampling_rate = provided_sr
+            if filepath:
+                self._file_path = filepath
         else:
             # otherwise, a valid filepath is required for lazy loading.
             if not filepath:
@@ -311,7 +313,7 @@ class Audio(BaseModel):
             if not os.path.exists(output_dir):
                 os.makedirs(output_dir)
             torchaudio.save(
-                uri=file_path,
+                uri=str(file_path),
                 src=self.waveform,
                 sample_rate=self.sampling_rate,
                 channels_first=True,
--- a/src/tests/audio/data_structures/audio_test.py
+++ b/src/tests/audio/data_structures/audio_test.py
@@ -228,7 +228,7 @@ def test_audio_save_to_file(audio_fixture: str, request: pytest.FixtureRequest) 
         assert temp_file_path.exists(), "The audio file was not saved."
 
         # Load the saved file and verify its content
-        loaded_waveform, loaded_sampling_rate = torchaudio.load(temp_file_path)
+        loaded_waveform, loaded_sampling_rate = torchaudio.load(str(temp_file_path))
         assert torch.allclose(audio_sample.waveform, loaded_waveform, atol=1e-5), "Waveform data does not match."
         assert audio_sample.sampling_rate == loaded_sampling_rate, "Sampling rate does not match."
 
--- a/src/tests/audio/data_structures/audio_test.py
+++ b/src/tests/audio/data_structures/audio_test.py
@@ -401,3 +401,15 @@ def test_window_generator_step_greater_than_audio(audio_fixture: str, request: p
     expected_windows = (audio_length - window_size) // step_size + 1  # This is always 1
     assert len(windowed_audios) == expected_windows, f"Should yield {expected_windows} \
         windows when step size is greater than audio length. Yielded {len(windowed_audios)}."
+
+
+@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not installed.")
+def test_audio_filepath_is_stored_with_waveform() -> None:
+    """Tests that the filepath is stored even when a waveform is provided."""
+    dummy_path = "/tmp/dummy_audio.wav"
+    dummy_waveform = torch.randn(1, 16000)
+    dummy_sr = 16000
+
+    audio = Audio(waveform=dummy_waveform, sampling_rate=dummy_sr, filepath=dummy_path)
+
+    assert audio.filepath() == dummy_path, "Filepath should be preserved even when waveform is provided."


ibevers:
@fabiocat93 what do you think?

fabiocat93 (approved):
makes sense. i would still encourage users to load the audio from file using our lazy loading instead of loading it with other tools and then initializing out Audio object.

ibevers:
> i would still encourage users to load the audio from file using our lazy loading instead of loading it with other tools and then initializing out Audio object.

Makes sense!

ibevers:
Screenshot 2025-05-14 at 3 50 15 PM
@fabiocat93 I can't merge it^^

fabiocat93:
> Screenshot 2025-05-14 at 3 50 15 PM @fabiocat93 I can't merge it^^

it's because the tests need to pass first

ibevers:
@fabiocat93 Ah, this needs to be finished first:
https://github.com/sensein/senselab/pull/333

Also possibly this one: https://github.com/sensein/senselab/issues/321

fabiocat93:
@wilke0818 since you also worked on the Audio interface, can you please double-check that this change doesn't mess up with other pieces of code in the package? I believe it's ready to be merged

src/senselab/audio/data_structures/audio.py
```diff
@@ -311,7 +313,7 @@ def save_to_file(
             if not os.path.exists(output_dir):
                 os.makedirs(output_dir)
             torchaudio.save(
-                uri=file_path,
+                uri=str(file_path),
```
wilke0818, line 316: there a reason for this conversion? file_path should already be PathLike (or string)

src/senselab/audio/data_structures/audio.py
```diff
@@ -80,6 +80,8 @@ def __init__(self, **data: Any) -> None:  # noqa: ANN401,D417
                 raise ValueError("When a waveform is provided, a sampling_rate must also be supplied.")
             self._waveform = self.convert_to_tensor(waveform)
             self._sampling_rate = provided_sr
+            if filepath:
```
wilke0818, line 83: what happens if the user provides an incorrect filepath compared to the waveform they provided? Feels like to cover this edge case of providing both we're generating other possible issues.

wilke0818: Or are you specifically thinking of a filepath that the user might want to call/save this audio as? If that's the case then I don't think this is the right approach.

wilke0818: Personally I am in favor of forcing the user to choose in a mutually exclusive manner rather than this

src/tests/audio/data_structures/audio_test.py
```diff
@@ -228,7 +228,7 @@ def test_audio_save_to_file(audio_fixture: str, request: pytest.FixtureRequest)
         assert temp_file_path.exists(), "The audio file was not saved."
 
         # Load the saved file and verify its content
-        loaded_waveform, loaded_sampling_rate = torchaudio.load(temp_file_path)
+        loaded_waveform, loaded_sampling_rate = torchaudio.load(str(temp_file_path))
```
wilke0818, line 231: again I don't understand the need for this

wilke0818 (commented):
Can you elaborate on the need you're thinking of with this change?

fabiocat93:
> Can you elaborate on the need you're thinking of with this change?

@ibevers is there a use case scenario where you found helpful to explicitly specify both the audio waveform and the filepath? If not, we can close this I guess. Especially since we all seem to agree that this change doesn't seem to promote a best practice and may lead to future issues that we are not controlling (e.g., the audio filepath doesn't actually point to the audio in the waveform)

src/senselab/audio/data_structures/audio.py
ibevers: I believe it was because I got an error with the sox backend when using .save similar to this:
https://github.com/pytorch/audio/issues/3775

src/senselab/audio/data_structures/audio.py
ibevers: >Or are you specifically thinking of a filepath that the user might want to call/save this audio as? If that's the case then I don't think this is the right approach.

@wilke0818 This is along the lines of what I was thinking. Would you suggest storing the file path in the metadata instead in this case?

src/tests/audio/data_structures/audio_test.py
ibevers: Same as before: https://github.com/pytorch/audio/issues/3775

ibevers:
>is there a use case scenario where you found helpful to explicitly specify both the audio waveform and the filepath? 

@fabiocat93 Yes, I wanted to have the original path so that I could store related files in the same path hierarchy. The path could be saved in the metadata instead of `filepath`. Is this the approach you would recommend?

fabiocat93:
> > is there a use case scenario where you found helpful to explicitly specify both the audio waveform and the filepath?
> 
> @fabiocat93 Yes, I wanted to have the original path so that I could store related files in the same path hierarchy. The path could be saved in the metadata instead of `filepath`. Is this the approach you would recommend?

I would recommend using `Audio(filepath=<your_file_path>)`. Why doesn't that work for you?

ibevers:
>Why doesn't that work for you?

@fabiocat93 `filepath=None` after modifying the Audio object. For example:
```
>>> from senselab.audio.tasks.preprocessing import resample_audios
>>> a1 = Audio(filepath="/Users/isaacbevers/sensein/senselab-wrapper/senselab/src/tests/data_for_testing/audio_48khz_mono_16bits.wav")
>>> print(a1.filepath())
/Users/isaacbevers/sensein/senselab-wrapper/senselab/src/tests/data_for_testing/audio_48khz_mono_16bits.wav
>>> a1 = resample_audios([a1], resample_rate=4000)[0]
>>> print(a1.filepath())
None
```

fabiocat93:
> resample_audios

oh. that's just because it's never attached back to the new audio object: https://github.com/sensein/senselab/blob/main/src/senselab/audio/tasks/preprocessing/preprocessing.py#L57
maybe attaching it as metadata is a good solution. we discussed about this some time ago and we decided that for now metadata handling is on the user and not on us

ibevers:
>maybe attaching it as metadata is a good solution

@fabiocat93 okay, if I want to store the source path after loading going forward, I will store it in the metadata

ibevers:
> > maybe attaching it as metadata is a good solution
> 
> @fabiocat93 okay, if I want to store the source path after loading going forward, I will store it in the metadata

Closing now since it appears that we don't need these changes

Issue #331, opened 2025-05-13
Bug: Issue when installing torbi


Rahul-Brito:
### Description

Following the contributing doc, when I run 

`poetry install --extras "audio text video" --with dev,docs`

I get errors on the torbi install:

![Image](https://github.com/user-attachments/assets/23dd0150-28e2-4b82-9c94-ba4fb724bcf0)

### Steps to Reproduce

run that above command after following the previous steps.

I also switched my python to 3.11

### Expected Results

smooth install with not errors

### Actual Results

and error

fabiocat93:
related to https://github.com/sensein/senselab/issues/321

Rahul-Brito:
Tried to force torch 2.6 with `pip install torch==2.6.0` and got the same error

satra:
@Rahul-Brito - what system are you installing this on? also is this into a fresh environment?

Rahul-Brito:
@satra on engaging, in a fresh conda environment.

@ibevers  - forcing torch 2.6 on my pyproject.toml did not solve this, got the same error

ibevers:
@Rahul-Brito have you tried downgrading `torbi`?

Rahul-Brito:
@ibevers not sure how, I don't have it installed and its not listed in the pyproject.toml file (i checked `conda list torbi` and `poetry show torbi` and its not in my conda env

fabiocat93:
@satra's note from another PR: it would be good to fix this underlying issue (consistency between pytorch+cuda and cuda development environment) as it will keep coming up. i think we may need to start thinking about container execution for certain parts of senselab.

noting some alternatives, but shouldn't prevent the merge.

torbi uses nvcc to compile hence the issue. can we break this into separate problems:

- on colab there is a mismatch between driver, cuda version, and pytorch version. torbi can be installed with no deps prior to senselab (one can technically install any cuda driver there, but installing torbi with no-deps is the fastest and will work).
- on our github workflows we don't need to have mismatch - i.e. we can create environments that match pytorch cuda versions/drivers.
- also, in most places the cuda version and compiler version can be matched to pytorch compiled cuda version
- move torbi/ppg/sparc (they all use it) into a sub install for now, instead of removing it (_**note**: this is how it has been done_)

fabiocat93:
since in most places the cuda version and compiler version can be matched to pytorch compiled cuda version and since we specified that as part of our doc, I will close this issue

PR #328, opened 2025-05-12
Fix torbi issue

ibevers:
## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

## Checklist:
- [ ] I have added tests to cover my changes.
- [ ] All new and existing tests passed.
- [ ] My code follows the code style of this project.

PR #337, opened 2025-05-20
solving different blockers

fabiocat93:
## Description
- After the release of PyTorch 2.7, users installing torbi with GPU support encounter build failures (see maxrmorrison/torbi#4). Because torbi is pulled in indirectly via sparc → penn, this change temporarily removes sparc from our dependency tree until the underlying issue is resolved upstream.
- This PR also updates the video backend import to use the av package (the old pyav alias has been removed)
- And it also applies a fix for Hugging Face Transformers’ word-level timestamping bug (huggingface/transformers#36093), restoring correct transcript alignment.

Note that this represent just a temporary patch, because as a consequence of this change the SPARC models are not usable

## Related Issue(s)
https://github.com/maxrmorrison/torbi/issues/4
https://github.com/sensein/senselab/issues/331
https://github.com/huggingface/transformers/issues/36093

## How Has This Been Tested?
Unit tests as usual

## Types of changes

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,7 +66,7 @@ pylangacq = {version = "~=0.19", optional = true}
 mediapipe = {version = "~=0.10", optional = true}
 opencv-python = {version = "~=4.10", optional = true}
 ultralytics = {version = "~=8.3", optional = true}
-pyav = {version = "~=14.2", optional = true}
+av = {version = "~=14.2", optional = true}
 
 [tool.poetry.extras]
 audio = [
@@ -80,7 +80,9 @@ audio = [
   "vocos",
   "jiwer",
   "nltk",
-  "coqui-tts",
+  "coqui-tts"
+]
+articulatory = [
   "speech-articulatory-coding"
 ]
 text = [
@@ -88,7 +90,7 @@ text = [
   "pylangacq"
 ]
 video = [
-  "pyav",
+  "av",
   "mediapipe",
   "opencv-python",
   "ultralytics"
--- a/src/senselab/audio/tasks/speech_to_text/huggingface.py
+++ b/src/senselab/audio/tasks/speech_to_text/huggingface.py
@@ -160,7 +160,8 @@ class HuggingFaceASR:
         start_time_transcription = time.time()
         # Run the pipeline
         transcriptions = pipe(
-            formatted_audios, generate_kwargs={"language": f"{language.name.lower()}"} if language else {}
+            formatted_audios,
+            generate_kwargs={"language": f"{language.name.lower()}", "num_beams": 1} if language else {"num_beams": 1},
         )
 
         # Take the end time of the transcription
--- a/src/senselab/video/data_structures/video.py
+++ b/src/senselab/video/data_structures/video.py
@@ -16,9 +16,9 @@ from senselab.utils.constants import SENSELAB_NAMESPACE
 try:
     import av  # noqa: F401
 
-    PYAV_AVAILABLE = True
+    AV_AVAILABLE = True
 except ModuleNotFoundError:
-    PYAV_AVAILABLE = False
+    AV_AVAILABLE = False
 
 
 class Video(BaseModel):
@@ -155,15 +155,15 @@ class Video(BaseModel):
         Raises:
             ValueError: If no file path is available for lazy loading.
             FileNotFoundError: If the file does not exist.
-            ModuleNotFoundError: If PyAV is not available.
+            ModuleNotFoundError: If AV is not available.
         """
         if not self._file_path:
             raise ValueError("No file path available for lazy loading.")
         if not os.path.exists(self._file_path):
             raise FileNotFoundError(f"File {self._file_path} does not exist.")
-        if not PYAV_AVAILABLE:
+        if not AV_AVAILABLE:
             raise ModuleNotFoundError(
-                "`pyav` is not installed. "
+                "`av` is not installed. "
                 "Please install senselab video dependencies using `pip install 'senselab[video]'`."
             )
         # Load video frames, audio frames, and metadata.
--- a/src/tests/audio/tasks/speech_to_text_test.py
+++ b/src/tests/audio/tasks/speech_to_text_test.py
@@ -77,7 +77,7 @@ def test_hf_asr_pipeline_factory(hf_model: HFModel, device: DeviceType, is_devic
 
 
 @pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not available")
-@pytest.mark.skipif(not torch.cuda.is_available(), reason="GPU is not available")
+# @pytest.mark.skipif(not torch.cuda.is_available(), reason="GPU is not available")
 @pytest.mark.parametrize("hf_model", ["hf_model", "hf_model2"], indirect=True)
 def test_transcribe_audios(
     resampled_mono_audio_sample: Audio, resampled_mono_audio_sample_x2: Audio, hf_model: HFModel
@@ -88,10 +88,7 @@ def test_transcribe_audios(
     )
     assert len(transcripts) == 2
     assert isinstance(transcripts[0], ScriptLine)
-    assert (
-        transcripts[0].text
-        == "This is Peter. This is Johnny. Kenny. And Joe. We just wanted to take a minute to thank you."
-    )
+    assert transcripts[0].text is not None and "This is Peter. This is Johnny. Kenny." in transcripts[0].text
 
 
 @pytest.mark.skipif(not TORCHAUDIO_AVAILABLE, reason="torchaudio is not available")
--- a/src/tests/utils/data_structures/dataset_test.py
+++ b/src/tests/utils/data_structures/dataset_test.py
@@ -28,9 +28,9 @@ except ModuleNotFoundError:
 try:
     import av
 
-    PYAV_AVAILABLE = True
+    AV_AVAILABLE = True
 except ModuleNotFoundError:
-    PYAV_AVAILABLE = False
+    AV_AVAILABLE = False
 
 
 def test_create_participant() -> None:
@@ -176,7 +176,7 @@ def test_audio_dataset_splits() -> None:
     ], "Excess GPU split should generate a list with one list of all of the audios, unpadded"
 
 
-@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE or not PYAV_AVAILABLE, reason="torchaudio or pyav are not installed")
+@pytest.mark.skipif(not TORCHAUDIO_AVAILABLE or not AV_AVAILABLE, reason="torchaudio or av are not installed")
 def test_convert_senselab_dataset_to_hf_datasets() -> None:
     """Tests the conversion of Senselab dataset to HuggingFace."""
     dataset = SenselabDataset(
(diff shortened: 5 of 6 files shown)


satra:
@fabiocat93 - this seems ok to me temporarily.

however, did you test by installing torch 2.7.0 and CUDA 12.6. the issue with google colab was related to a mismatch in cuda version with the cuda version that torch was built with according to the discussion on the issue i filed at torbi.

satra:
when i look at the nvidia smi output on the failed test cuda version is:  
`NVIDIA-SMI 535.183.01             Driver Version: 535.183.01   CUDA Version: 12.2  `

can that be updated to CUDA version `12.6` as that would also cause a mismatch with pytorch 2.7?

fabiocat93:
> @fabiocat93 - this seems ok to me temporarily.
> 
> however, did you test by installing torch 2.7.0 and CUDA 12.6. the issue with google colab was related to a mismatch in cuda version with the cuda version that torch was built with according to the discussion on the issue i filed at torbi.

I am not sure it's just due to a mismatch of cuda version. I tried to force torch 2.6 with CUDA 12.2 and CUDA 12.4, but didn't solve the issue (both on engaging and AWS). Btw, Engaging doesn't have CUDA 12.6 modules available, yet

satra:
> I am not sure it's just due to a mismatch of cuda version. I tried to force torch 2.6 with CUDA 12.2 and CUDA 12.4, but didn't solve the issue (both on engaging and AWS). 

i'll post on the issue at torbi. 

on engaging, cuda modules i think can be installed personally (in conda) as long as the drivers support it.

fabiocat93:
> on engaging, cuda modules i think can be installed personally (in conda) as long as the drivers support it.

it requires sudo privileges, though

satra:
on engaging, this worked fine. i used 3.10 only because it seemed the runners were using it.

```
conda create -n torbi python=3.10
conda activate torbi
conda config --add channels nvidia
conda install nvidia/label/cuda-12.6.3::cuda-libraries-dev
pip install torch torchvision torchaudio
pip install torbi
```

satra:
however the above didn't install `penn` and nvidia-smi still showed 12.4 on engaging. just that it seems torbi install fine with pytorch 2.7

```
>>> import torch
>>> torch.__version__
'2.7.0+cu126'
```

Rahul-Brito:
I checked out the `torbi` branch and ran `poetry install --extras "audio text video" --with dev,docs` per the contributing.md doc. It seemed to kinda work (and not error on the torbi part!) It gave me the bellow error/warning log, seemed like it had an issue with the nvidia cuda runtime `(nvidia-cuda-runtime-cu12 (12.4.127 -> 12.6.77)`

It was able to do this update: `nvidia-nvtx-cu12 (12.4.127 -> 12.6.77)` but not sure if that is the same thing or not

Full log below:

```
Updating dependencies
Resolving dependencies... (35.7s)

Package operations: 82 installs, 8 updates, 0 removals

  - Updating nvidia-cuda-runtime-cu12 (12.4.127 -> 12.6.77): Failed

  CalledProcessError

  Command '['/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/bin/python', '-m', 'pip', 'uninstall', 'nvidia-cuda-runtime-cu12', '-y']' returned non-zero exit status 2.

  at /orcd/home/002/rfbrito/miniconda3/envs/senselab/lib/python3.13/subprocess.py:577 in run
       573│             # We don't call process.wait() as .__exit__ does that for us.
       574│             raise
       575│         retcode = process.poll()
       576│         if check and retcode:
    →  577│             raise CalledProcessError(retcode, process.args,
       578│                                      output=stdout, stderr=stderr)
       579│     return CompletedProcess(process.args, retcode, stdout, stderr)
       580│ 
       581│ 

The following error occurred when trying to handle this error:


  EnvCommandError

  Command ['/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/bin/python', '-m', 'pip', 'uninstall', 'nvidia-cuda-runtime-cu12', '-y'] errored with the following return code 2
  
  Output:
  Found existing installation: nvidia-cuda-runtime-cu12 12.4.127
  Uninstalling nvidia-cuda-runtime-cu12-12.4.127:
  ERROR: Exception:
  Traceback (most recent call last):
    File "/usr/lib64/python3.11/shutil.py", line 853, in move
      os.rename(src, real_dst)
  OSError: [Errno 18] Invalid cross-device link: '/orcd/home/002/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/nvidia/__init__.py' -> '/tmp/pip-uninstall-233dkzzh/__init__.py'
  
  During handling of the above exception, another exception occurred:
  
  Traceback (most recent call last):
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 105, in _run_wrapper
      status = _inner_run()
               ^^^^^^^^^^^^
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/cli/base_command.py", line 96, in _inner_run
      return self.run(options, args)
             ^^^^^^^^^^^^^^^^^^^^^^^
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/commands/uninstall.py", line 106, in run
      uninstall_pathset = req.uninstall(
                          ^^^^^^^^^^^^^^
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/req/req_install.py", line 723, in uninstall
      uninstalled_pathset.remove(auto_confirm, verbose)
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/req/req_uninstall.py", line 370, in remove
      moved.stash(path)
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/req/req_uninstall.py", line 261, in stash
      renames(path, new_path)
    File "/home/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/pip/_internal/utils/misc.py", line 350, in renames
      shutil.move(old, new)
    File "/usr/lib64/python3.11/shutil.py", line 873, in move
      copy_function(src, real_dst)
    File "/usr/lib64/python3.11/shutil.py", line 448, in copy2
      copyfile(src, dst, follow_symlinks=follow_symlinks)
    File "/usr/lib64/python3.11/shutil.py", line 256, in copyfile
      with open(src, 'rb') as fsrc:
           ^^^^^^^^^^^^^^^
  FileNotFoundError: [Errno 2] No such file or directory: '/orcd/home/002/rfbrito/.cache/pypoetry/virtualenvs/senselab-9q3YbY4j-py3.11/lib/python3.11/site-packages/nvidia/__init__.py'
  

  at /orcd/home/002/rfbrito/.local/share/pipx/venvs/poetry/lib/python3.13/site-packages/poetry/utils/env/base_env.py:434 in _run
      430│                 output = subprocess.check_output(
      431│                     cmd, stderr=stderr, env=env, text=True, encoding=encoding, **kwargs
      432│                 )
      433│         except CalledProcessError as e:
    → 434│             raise EnvCommandError(e)
      435│ 
      436│         return output
      437│ 
      438│     def execute(self, bin: str, *args: str, **kwargs: Any) -> int:

Cannot install nvidia-cuda-runtime-cu12.

  - Updating nvidia-nvtx-cu12 (12.4.127 -> 12.6.77)
  ```

File: pyproject.toml [hunk 1]
```diff
@@ -66,7 +66,7 @@ pylangacq = {version = "~=0.19", optional = true}
 mediapipe = {version = "~=0.10", optional = true}
 opencv-python = {version = "~=4.10", optional = true}
 ultralytics = {version = "~=8.3", optional = true}
-pyav = {version = "~=14.2", optional = true}
+av = {version = "~=14.2", optional = true}
```
ibevers, line 69: Maybe we want to adopt the practice of specifying an exact version instead of `~=14.2` to prevent nasty surprises like we encountered with `torbi` and `torch==2.7`?

ibevers (commented):
@fabiocat93 Why did you make a new PR instead of using these existing ones?
https://github.com/sensein/senselab/pull/333
https://github.com/sensein/senselab/pull/328

Also:
- This PR doesn't follow the contributing guidelines (no corresponding issue).
- The scope of this PR is ambiguous and aggregates multiple atomic issues. I think it is better practice to have separate PRs for each since it facilitates faster integration.

fabiocat93:
> @fabiocat93 Why did you make a new PR instead of using these existing ones? #333 #328
> 
> Also:
> 
> * This PR doesn't follow the contributing guidelines (no corresponding issue).
> * The scope of this PR is ambiguous and aggregates multiple atomic issues. I think it is better practice to have separate PRs for each since it facilitates faster integration.

Not at all @ibevers  The 2 problems need to be solved in the same PR, otherwise cannot be tested. Treating them as 2 separate issues leads to a deadlock. But thanks for your comment

pyproject.toml
satra: not a very good practice of specifying exact version in general. it causes too many releases and a lot of work. so willing to live with a few surprises, but we should enable dependabot updates perhaps if not already in place.

fabiocat93: Already in place

satra:
based on my experiments on engaging, it does seem that torbi is fine with pytorch 2.7, so i think this patch is not fixing the underlying issue of cuda versions. we should be able to adjust the github workflows to make sure cuda 12.6 versions.

fabiocat93:
> based on my experiments on engaging, it does seem that torbi is fine with pytorch 2.7, so i think this patch is not fixing the underlying issue of cuda versions. we should be able to adjust the github workflows to make sure cuda 12.6 versions.

indeed this PR isn't for fixing the problem, it's for at least enabling the installation of senselab (which btw hasn't been installable for around the last 20 days...) while someone implements an actual solution (ensuring cuda 12.6 for the CI is easy, it's more challenging to make senselab work on Colab and other machines missing the latest drivers and the possibility to install them - e.g., there is no conda option on Colab)

satra:
for now it would be fine to merge, but it would be good to fix this underlying issue (consistency between pytorch+cuda and cuda development environment) as it will keep coming up. i think we may need to start thinking about container execution for certain parts of senselab.

noting some alternatives, but shouldn't prevent the merge.

torbi uses nvcc to compile hence the issue.  can we break this into separate problems:

1. on colab there is a mismatch between driver, cuda version, and pytorch version. torbi can be installed with no deps prior to senselab (one can technically install any cuda driver there, but installing torbi with no-deps is the fastest and will work).
2. on our github workflows we don't need to have mismatch - i.e. we can create environments that match pytorch cuda versions/drivers.
3. also, in most places the cuda version and compiler version can be matched to pytorch compiled cuda version
4. move torbi/ppg/sparc (they all use it) into a sub install for now, instead of removing it.

PR #342, opened 2025-05-30
Fix nest_asyncio crash on uvloop: apply only when safe

### LINKED ISSUE #341: Make nest_asyncio application conditional to avoid conflict with uvloop
Currently, `senselab` unconditionally applies nest_asyncio in __init__.py, which leads to a ValueError when the default event loop is replaced with uvloop (e.g., in performance-critical applications like the interactive avatar I am implementing).

To avoid this issue and make the library more robust across different environments (e.g., Jupyter notebooks, scripts, production servers using uvloop), I propose replacing the current nest_asyncio.apply() call with a safer conditional version:

```python
import asyncio
import nest_asyncio

def safe_apply_nest_asyncio():
    try:
        loop = asyncio.get_event_loop()
        # Only apply if it's a standard asyncio event loop
        if "uvloop" not in str(type(loop)):
            nest_asyncio.apply()
    except Exception as e:
        # Log if needed
        print(f"nest_asyncio not applied: {e}")

safe_apply_nest_asyncio()
```

This approach should ensure:
- Compatibility with notebooks and other nested event loop scenarios.
- No runtime crashes when uvloop is active.
- No behavior changes in typical asyncio use cases.

fabiocat93:
## Description
This PR updates the initialization logic to apply nest_asyncio only if the current event loop is not managed by uvloop, which avoids a known runtime ValueError when both are used together. 

## Related Issue(s)
https://github.com/sensein/senselab/issues/341

## Motivation and Context
This makes the library more robust across different environments, including Jupyter notebooks, scripts, and production servers using uvloop.

## How Has This Been Tested?
Unit tests

## Types of changes

## Checklist:

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -110,6 +110,7 @@ codespell = "~=2.3"
 jupyter = "~=1.0"
 ipywidgets = "~=8.1"
 ipykernel = "*"
+uvloop = "~=0.21"
 
 [tool.poetry.group.docs]
 optional = true
--- a/src/senselab/__init__.py
+++ b/src/senselab/__init__.py
@@ -1,23 +1,32 @@
 """.. include:: ../../README.md"""  # noqa: D415
 
 import platform
+import asyncio
+import nest_asyncio
 from multiprocessing import set_start_method
 
+# Raise error on incompatible macOS architecture
 if platform.system() == "Darwin" and platform.machine() != "arm64":
     raise RuntimeError(
         "Error: This package requires an ARM64 architecture on macOS "
         "since PyTorch 2.2.2+ does not support x86-64 on macOS."
     )
 
+# Conditionally apply nest_asyncio to avoid uvloop conflict
+def safe_apply_nest_asyncio():
+    try:
+        loop = asyncio.get_event_loop()
+        if "uvloop" not in str(type(loop)):
+            nest_asyncio.apply()
+    except Exception as e:
+        print(f"nest_asyncio not applied: {e}")
 
-import nest_asyncio
-
-nest_asyncio.apply()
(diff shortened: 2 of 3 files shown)


satra (approved):
overall looks fine to me. if there was a way to create a test that fails before this fix, but passes after would be good. but i suspect uvloop would become a requirement.

fabiocat93:
> overall looks fine to me. if there was a way to create a test that fails before this fix, but passes after would be good. but i suspect uvloop would become a requirement.

It shouldn't be a problem to add `uvloop` as a dev dependency. Thank you for the hint. I will try to be more rigorous with the method and do test-based development

ibevers:
@fabiocat93 it's interesting that the tests passed before merging but not when the PR was merged. Is not running the tests here (associated with error code 5) the intended behavior? Asking out of curiosity mainly.
Screenshot 2025-06-04 at 11 26 30 AM

fabiocat93:
> @fabiocat93 it's interesting that the tests passed before merging but not when the PR was merged. Is not running the tests here (associated with error code 5) the intended behavior? Asking out of curiosity mainly. Screenshot 2025-06-04 at 11 26 30 AM

That's mostly a problem with pytest-xdist. It's way faster than the simple pytest, but much more unstable and sometimes it failes just like that. Launching the same test without touching it. Using fewer threads helps reduce instability, we may want to do that (this will have an impact on performance though)

ibevers:
>That's mostly a problem with pytest-xdist. It's way faster than the simple pytest, but much more unstable and sometimes it failes just like that

@fabiocat93 Oh, I remember running into similar issues before. Thanks for explaining!

PR #366, opened 2025-08-25
Add HeAR Embedding Extraction 

### LINKED ISSUE #364: Task: Add Hear Embedding Extraction
### Description

Use Hear to extract embeddings for voice health applications.

https://huggingface.co/google/hear

### Tasks

- [ ] Implement functionality
- [ ] Create tutorial
- [ ] Add tests

### Freeform Notes

_No response_

ibevers:
## Types of changes

## Checklist:


ibevers:
Added some rough AI-generated quick-and-dirty code that works AFAIK (currently running on Bridge2AI Voice).

ibevers:
@wilke0818 FYI

ibevers:
For about 20k audio files, tqdm's ETA for this script with MPS acceleration is about five hours.

ibevers:
@Evan8456 We probably want to add this to the feature extraction in b2aiprep for Bridge2AI once it is added here.

satra:
Superseded by #568, which lands HeAR as a full senselab task. Closing this one — thank you for the groundwork, several pieces of it carried over.

**What #568 reuses from here:** the 50% window overlap default, window mean-pooling (now an explicit `pooled()` on the result type), and resampling through senselab rather than librosa.

**Why a rewrite rather than a rebase**, in three parts:

*Placement.* This PR puts HeAR under `speaker_embeddings/`. HeAR describes what a sound is, not who made it — it is a health-acoustics model whose labels are Cough, Snore, Baby Cough, Breathe, Sneeze, Throat Clear, Laugh and Speech. #568 creates `audio/tasks/health_acoustics/` and reaches the detector from `classification/` through a small dispatch alias, so the yamnet precedent is honoured without splitting one model family across two packages.

*Shape.* This is a research script rather than a task: `transformers.AutoModel`, `pandas`, `tqdm`, `notebook_login`, directory scanning, a bare `from utils import get_audio_files_from_directory` that cannot resolve as a package module, and a comment reading "Try to import senselab for future use". There is no `api.py` integration, no registry entry and no tests. Rebasing it onto a year of task-convention changes would have meant rewriting all of that anyway.

*The transformers path turns out not to work.* Worth recording, since it is the crux of the approach here: `google/hear-pytorch` is separately gated and this account is not authorized for it, so equivalence to the SavedModel cannot even be verified; it is a spectrogram-input ViT with the PCEN frontend absent (this PR imports that from a `sys.path.append`ed clone of `google-health/hear`); and it carries no event detector at all. #568 therefore uses the TF SavedModels in an isolated subprocess venv, which is the established pattern here for foreign frameworks.

*What we learned in between.* A week of measurement on real audio produced constraints the API now enforces rather than documents: the detector's window is hard-fixed at 32000 samples — 0.5, 1.0, 1.5, 3.0 and 4.0 s all raise `InvalidArgumentError` — while the encoder *silently accepts* any length from 160 to 64000 samples and returns a plausible-looking vector; padding a 0.3 s clip out to 2 s moves its embedding about as far as substituting unrelated audio, so no code path in #568 can pad; and raw cosine similarity is uninformative (0.977 within-class against 0.918 between) so the API centres by construction.

PR #507, opened 2026-05-04
Add scalene profiling wrapper as opt-in dependency group

satra:
## Summary

Adds a Scalene-based profiling tool to senselab as an opt-in developer dependency. A single CLI wrapper (`scripts/profile_with_scalene.py`) accepts any `.py` script or `.ipynb` notebook, transparently converts notebooks via `nbconvert` (with an injected IPython stub so magic calls survive), and produces self-contained HTML or JSON reports under `artifacts/scalene/`.

- Scalene + nbconvert live in a new `[dependency-groups.profiling]` entry in `pyproject.toml`. Default `uv sync` is unchanged — production installs do not pay the profiler cost.
- The wrapper invokes Scalene 2.2 in two steps: `scalene run` produces JSON, `scalene view --standalone` produces a single self-contained HTML.
- Supports path-substring scoping via `--scope`, `--no-thirdparty`, `--exclude`, plus `--cpu-only`, `--gpu`, `--keep-intermediate`, `--format {html,json}`.
- 5 pytest smoke tests under `src/tests/scripts/`, each guarded by `@pytest.mark.skipif(not scalene_available)` so the existing CI suite is unaffected when the `profiling` group is not installed.

To use:

```bash
uv sync --group profiling
uv run python scripts/profile_with_scalene.py path/to/script.py
uv run python scripts/profile_with_scalene.py tutorials/audio/x.ipynb
```

Coexists with `scripts/profile_imports.py` (cold-start import-time profiler from a previous feature). The two are complementary: `profile_imports.py` measures import time, `profile_with_scalene.py` measures everything after imports.

Spec, plan, and CLI contract live in `specs/20260503-235625-scalene-profiling/` (workspace-only, not part of this commit).

## Test plan

- [ ] `uv sync` (no flag) does not pull scalene
- [ ] `uv sync --group profiling` installs scalene 2.2.x and nbconvert >=7
- [ ] `uv run pytest src/tests/scripts/` passes (5 tests)
- [ ] `uv run pytest -x --ignore=src/tests/scripts` still passes (existing suite unaffected)
- [ ] `uv run python scripts/profile_with_scalene.py --help` prints all CLI options
- [ ] `uv run python scripts/profile_with_scalene.py /tmp/tiny.py` produces HTML under `artifacts/scalene/`
- [ ] `uv run python scripts/profile_with_scalene.py some_tutorial.ipynb` converts and profiles successfully

# Version

Published prerelease version: `1.3.1-alpha.27`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - Add scalene profiling wrapper as opt-in dependency group [#507](https://github.com/sensein/senselab/pull/507) ([@satra](https://github.com/satra))
  
  #### Authors: 1
  
</details>

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,10 @@ dev = [
 docs = [
   "pdoc~=16.0"
 ]
+profiling = [
+  "scalene>=2.2",
+  "nbconvert>=7"
+]
 
 [tool.hatch.build.targets.wheel]
 packages = ["src/senselab"]
--- a/scripts/profile_with_scalene.py
+++ b/scripts/profile_with_scalene.py
@@ -24,6 +24,7 @@ from __future__ import annotations
 import argparse
 import importlib.util
 import os
+import re
 import secrets
 import shutil
 import subprocess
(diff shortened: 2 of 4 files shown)


satra:
Verified valid — reproduced the SyntaxError with a one-line file `import sys ... \nfrom __future__ import annotations`. Fixed in ffee86a1: the stub is now inserted *after* any `from __future__ import` lines via line-walk. Added a regression test (`test_wrapper_handles_notebook_with_future_imports`) using a synthetic notebook that starts with `from __future__ import annotations`. Thanks for catching this.

---

**Re. medium-priority comment on module stubbing:**

Respectfully disagree — verified empirically that `m.__getattr__ = lambda name: ...` does work for module attribute access on Python 3.7+ via PEP 562. Test:

```python
import sys, types
m = types.ModuleType("foo")
m.__getattr__ = lambda name: "fallback"
sys.modules["foo"] = m
from foo import bar  # missing attribute
print(bar)  # -> "fallback"
```

This prints `fallback` correctly, so `from IPython.display import Audio` (or any other unfamiliar name) hits the fallback as designed. PEP 562 looks up `__getattr__` from the module dict, which is what we set here. Leaving the implementation as-is.

PR #508, opened 2026-05-05
Promote scalene profiling wrapper to main (alpha→main)

satra:
Promotes the Scalene profiling wrapper from alpha to main. Already validated end-to-end on `alpha` via PR #507 (squash-merged after CI green and review).

## Summary

- Adds `scripts/profile_with_scalene.py`: a thin wrapper around Scalene 2.2 that profiles `.py` scripts and `.ipynb` notebooks and writes standalone HTML or JSON reports to `artifacts/scalene/`
- Adds `[dependency-groups.profiling]` (scalene + nbconvert) — opt-in only; default `uv sync` is unchanged
- Adds smoke tests under `src/tests/scripts/`, guarded by `@pytest.mark.skipif(not scalene_available)` so existing CI is unaffected when the profiling group is not installed

## Test plan

- [x] CI green on `alpha` (PR #507): pre-commit, build-and-deploy, cpu-tests (3.12) all passed
- [ ] CI green on alpha→main PR (this PR)
- [x] All 6 wrapper smoke tests pass (`uv run pytest src/tests/scripts/`)
- [x] Default `uv sync` does not pull scalene (verified)
- [x] `uv sync --group profiling` installs scalene 2.2.x and nbconvert ≥7

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,6 +108,10 @@ dev = [
 docs = [
   "pdoc~=16.0"
 ]
+profiling = [
+  "scalene>=2.2",
+  "nbconvert>=7"
+]
 
 [tool.hatch.build.targets.wheel]
 packages = ["src/senselab"]
--- a/scripts/profile_with_scalene.py
+++ b/scripts/profile_with_scalene.py
@@ -1,0 +1,391 @@
+#!/usr/bin/env python3
+"""Profile any Python script or Jupyter notebook with Scalene.
+
+A thin wrapper around Scalene 2.2's `run` and `view` subcommands that:
+- Accepts .py or .ipynb targets (notebooks are auto-converted via nbconvert)
+- Produces a standalone HTML report (default) or JSON profile
+- Supports path-substring scoping (--scope, --no-thirdparty, --exclude)
+- Writes reports to artifacts/scalene/ with a timestamp suffix
+
+Install:
+    uv sync --group profiling
+
+Examples:
+    uv run python scripts/profile_with_scalene.py path/to/script.py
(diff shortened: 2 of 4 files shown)


satra:
Addressing both review comments:

**MEDIUM (line 386 — quote paths in "Open with"):**
Valid. Fixed in 6faab494 using `shlex.quote()` (more robust than naive double-quoting; handles spaces, quotes, and any other shell-special characters in paths). Added a regression test `test_wrapper_quotes_paths_with_spaces` that runs the wrapper with `--output-dir "dir with space"` and asserts the printed `Open with:` line contains a shell-quoted path.

**HIGH (line 191 — IPython module `__getattr__` ineffective):**
False positive. This is the same claim the bot made on PR #507 (PR that this one promotes). PEP 562 (Python 3.7+) explicitly supports module-level `__getattr__`, and CPython looks it up from the module dict — which is exactly where `m.__getattr__ = lambda name: ...` puts it. Verified empirically once again before responding:

```python
import sys, types
m = types.ModuleType("IPython.display")
m.__getattr__ = lambda name: (lambda *a, **k: None)
sys.modules["IPython.display"] = m
from IPython.display import HTML, Audio  # both names we never pre-set
print(HTML("x"), Audio("y"))   # prints `None None` — fallback fired correctly
```

The smoke test `test_wrapper_handles_notebook_with_ipython_imports` already exercises this path (notebook with `from IPython.display import display` and `get_ipython().run_line_magic(...)`) and passes. Leaving the implementation unchanged.

References: [PEP 562 — Module __getattr__](https://peps.python.org/pep-0562/).

PR #510, opened 2026-05-08
audio analysis script + multilingual MMS aligner + 4 new ASR backends

satra:
## Summary

- **`scripts/analyze_audio.py`** — single-command multi-task audio analysis (diarization, AST, YAMNet, quality features, ASR, speaker embeddings) with content-addressable caching, full provenance, and a hierarchical Label Studio export bundle. Runs the same task suite twice (raw + speech-enhanced) and produces parallel timeline tracks per analyzer × model.
- **`senselab.audio.tasks.forced_alignment` extension** — multilingual MMS aligner (`facebook/mms-1b-all`, ~1100 languages) selectable via `aligner_model=MMS_MODEL_ID`. Auto-romanizes ja/zh transcripts via `uroman` (added under `[nlp]` extra).
- **Four new ASR backends**, auto-routed by model id:
  - `nvidia/canary-qwen-` → NeMo SALM in isolated `nemo-canary-qwen` venv (text-only)
  - `Qwen/Qwen3-ASR` → Alibaba `qwen-asr` wrapper in isolated `qwen-asr` venv (native word-level chunks via the bundled `Qwen3-ForcedAligner-0.6B` companion)
  - `ibm-granite/granite-speech-` → in-process backend using `GraniteSpeechProcessor` directly (no venv; just adds `peft>=0.13` to core deps for the LoRA adapter)
  - HF pipeline path: now correctly handles `return_timestamps=False` for CTC / generative-seq2seq models that previously raised
- **Auto-align stage in the script** — when an ASR result lacks native timestamps (Granite, Canary-Qwen), the script post-processes the text through MMS to add per-segment timing. Cached separately from the (slow) ASR result, so re-running just alignment doesn't re-run ASR.

## Why some backends use subprocess venvs and others don't

| Backend | Approach | Reason |
|---|---|---|
| Granite Speech 3.3 | in-process | No version conflicts; only adds `peft` (small wheel). bfloat16 on CUDA + MPS, float32 on CPU. |
| Canary-Qwen 2.5B | subprocess venv | NeMo trunk + `nemo_toolkit[asr,tts]` is huge and unstable |
| Qwen3-ASR 1.7B / 3B | subprocess venv | `qwen-asr` 0.0.6 pins `transformers==4.57.6`; senselab core requires `>=5.0` (hard conflict) |

## E2E validation (run 15 on a 38.4s real-speech WAV)

[...]

PR #511, opened 2026-05-08
Fix/ser wav2vec2 regression head

wilke0818:
## Description
Fixes continuous SER, at least for a subset of models by using a custom class that matches the expected class

## Related Issue(s)
None

## Motivation and Context
Use of the continuous SER by collaborators

## How Has This Been Tested?
Manually

## Types of changes

## Checklist:

# Version

Published prerelease version: `1.3.1-alpha.30`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  #### Authors: 3
  
</details>


src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -200,6 +282,67 @@ def _classify_continuous_ser_venv(
...
+        _wav2vec2_emotion_models[key] = (loaded, processor, id2label)
+
+    loaded_model, processor, id2label = _wav2vec2_emotion_models[key]
+
+    results: List[AudioClassificationResult] = []
+    for audio in audios:
+        if audio.waveform.shape[0] != 1:
+            raise ValueError("Only mono audio is supported for continuous SER.")
+        data = audio.waveform.squeeze().numpy()
+        inputs = processor(data, sampling_rate=audio.sampling_rate, return_tensors="pt", padding=True)
```
satra: **Silent sampling-rate hazard.** `Wav2Vec2Processor` normalizes/pads but does not resample — it just trusts the `sampling_rate` you tell it. If a caller passes 44.1 kHz audio, this runs the model on out-of-distribution input and returns plausible-but-wrong scores (arguably worse than the 0.33 bug because it's silent).

The convention used elsewhere in the repo (see `huggingface.py:142`) is to resample inside the loop:

```python
expected_sr = int(getattr(processor.feature_extractor, "sampling_rate", 16000) or 16000)
...
for audio in audios:
    if audio.waveform.shape[0] != 1:
        raise ValueError("Only mono audio is supported for continuous SER.")
    if audio.sampling_rate != expected_sr:
        audio = resample_audios([audio], resample_rate=expected_sr)[0]
    data = audio.waveform.squeeze().numpy()
    inputs = processor(data, sampling_rate=expected_sr, return_tensors="pt", padding=True)
```

---
_Generated by [Claude Code](https://claude.ai/code)_

src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -200,6 +282,67 @@ def _classify_continuous_ser_venv(
...
+        inputs = {k: v.to(device_type.value) for k, v in inputs.items()}
+        with torch.no_grad():
+            _, logits = loaded_model(inputs["input_values"])
+        scores = logits[0].cpu().tolist()
+        labels = [id2label[i] for i in range(len(scores))]
```
satra: **`id2label` indexing can KeyError.** HuggingFace stores `id2label` keys as strings in JSON. `AutoConfig.from_pretrained` normally coerces them to ints, but the **raw-dict fallback** above (`AutoConfig.for_model("wav2vec2", **config_dict)`) doesn't always — the keys may stay as `"0"`, `"1"`, .... Indexing with `id2label[i]` then throws.

Recommend coercing once when caching, with a fallback for missing keys:

```python
raw = getattr(config, "id2label", None) or {}
coerced: dict = {}
for k, v in raw.items():
    try:
        coerced[int(k)] = v
    except (TypeError, ValueError):
        continue
n = int(getattr(config, "num_labels", len(coerced)) or len(coerced))
labels = [coerced.get(i, f"class_{i}") for i in range(n)]
```

(Mirrors the SpeechBrain branch's `class_{i}` fallback at `_load_speechbrain_ser_model`.)

---
_Generated by [Claude Code](https://claude.ai/code)_

src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -200,6 +282,67 @@ def _classify_continuous_ser_venv(
...
+            config = AutoConfig.from_pretrained(model.path_or_uri, revision=model.revision)
+        except Exception:
+            from huggingface_hub import hf_hub_download
+
+            config_path = hf_hub_download(str(model.path_or_uri), "config.json", revision=model.revision)
+            with open(config_path) as f:
+                config_dict = json.load(f)
+            from transformers import AutoConfig as _AC
+
+            config = _AC.for_model("wav2vec2", **config_dict)
```
satra: Nit: `AutoConfig` is already imported at the top of this function (line 296) — the `_AC` alias is redundant.

```suggestion
            config = AutoConfig.for_model("wav2vec2", **config_dict)
```

---
_Generated by [Claude Code](https://claude.ai/code)_

src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -85,6 +95,88 @@ class SERType(Enum):
...
+            super().__init__(config)
+            self.wav2vec2 = Wav2Vec2Model(config)
+            self.classifier = _Wav2Vec2RegressionHead(config)
+            self.init_weights()
+
+        def forward(self, input_values: torch.Tensor) -> Tuple[torch.Tensor, torch.Tensor]:
+            hidden = self.wav2vec2(input_values)[0]
+            pooled = hidden.mean(dim=1)
+            return pooled, self.classifier(pooled)
+
```
satra: Nit: `_make_wav2vec2_emotion_model_class()` rebuilds the class every cache miss, so `isinstance` checks against it are unstable across calls. Easy to memoize:

```python
_wav2vec2_emotion_model_cls: Optional[type] = None

def _make_wav2vec2_emotion_model_class() -> type:
    global _wav2vec2_emotion_model_cls
    if _wav2vec2_emotion_model_cls is not None:
        return _wav2vec2_emotion_model_cls
    from transformers import Wav2Vec2Model, Wav2Vec2PreTrainedModel
    ...
    _wav2vec2_emotion_model_cls = _Wav2Vec2EmotionModel
    return _Wav2Vec2EmotionModel
```

Also — `Tuple` from `typing` can become lowercase `tuple` (the codebase already uses lowercase generics in `_load_speechbrain_ser_model` etc.), which lets you drop `Tuple` from the imports.

---
_Generated by [Claude Code](https://claude.ai/code)_

satra (commented):
Thanks for tracking this down — the head-structure mismatch diagnosis is solid and the new `_Wav2Vec2EmotionModel` reproduces the original audeering class faithfully. A few items worth addressing before merge:

**CI**: `cpu-tests (3.12)` is failing on this commit ([job 75099350753](https://github.com/sensein/senselab/actions/runs/25580902679/job/75099350753)). The PR description ticks "All new and existing tests passed" — please investigate; `test_speech_emotion_recognition_continuous` exercises this exact path now.

**Correctness** (see inline):
1. Silent sampling-rate hazard — `Wav2Vec2Processor` does not resample.
2. `id2label[i]` can `KeyError` in the raw-dict fallback path because HF JSON keys are strings.

**Test coverage**: the existing continuous-SER test only checks label *names*, which would have passed before this fix too (random head still produced correct labels with bogus scores). Worth adding:
- `assert set(labels) == {"arousal", "valence", "dominance"}`
- scores in `[0, 1]`
- regression guard that not all scores ≈ 0.33 (`any(abs(s - 0.33) > 0.05 for s in scores)`)

**Minor cleanup** (see inline): redundant `_AC` alias; `Tuple` import unused if switched to lowercase `tuple`; class factory rebuilds on every cache miss.

Otherwise looks good — once the CI failure is understood and the resample/id2label edges are handled, happy to see this land.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -85,6 +95,88 @@ class SERType(Enum):
...
+    weights are stored under classifier.dense.* / classifier.out_proj.*. The
+    standard pipeline would randomly initialize those layers instead.
+    """
+    try:
+        from transformers import AutoConfig
+
+        config = AutoConfig.from_pretrained(model.path_or_uri, revision=model.revision)
+        auto_map = getattr(config, "auto_map", None)
+        architectures = getattr(config, "architectures", None) or []
+    except Exception:
```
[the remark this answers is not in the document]
satra: The broad catch is intentional. The exception we actually need to recover from on this codebase is `huggingface_hub.dataclasses.StrictDataclassFieldValidationError`, raised when a checkpoint ships `vocab_size: null` (e.g. `audeering/wav2vec2-large-robust-12-ft-emotion-msp-dim`) and `huggingface_hub>=1.0`'s strict-dataclass validators reject it. Its MRO is `StrictDataclassFieldValidationError → StrictDataclassError → Exception` — it's not a `ValueError`, `TypeError`, or `OSError`, so narrowing the catch to those would silently bypass the fallback path and re-introduce the exact "near-0.33 outputs" bug this PR is fixing. Leaving as `except Exception:`.

---
_Generated by [Claude Code](https://claude.ai/code)_

src/senselab/audio/tasks/classification/speech_emotion_recognition/api.py
```diff
@@ -200,6 +282,67 @@ def _classify_continuous_ser_venv(
...
+    device_type, _ = _select_device_and_dtype(
+        user_preference=device, compatible_devices=[DeviceType.CUDA, DeviceType.CPU]
+    )
+
+    key = f"{model.path_or_uri}-{model.revision or 'main'}-{device_type.value}"
+    if key not in _wav2vec2_emotion_models:
+        EmotionModel = _make_wav2vec2_emotion_model_class()
+        try:
+            config = AutoConfig.from_pretrained(model.path_or_uri, revision=model.revision)
+        except Exception:
```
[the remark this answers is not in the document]
satra: Same rationale as the line-180 thread: the broad catch is needed because the actual recoverable exception is `huggingface_hub.dataclasses.StrictDataclassFieldValidationError` (MRO ends at `Exception`), not in the `ValueError`/`TypeError`/`OSError` set. Narrowing breaks the fallback that this fix relies on.

On the refactor suggestion: the two `try`/`except` blocks look similar but extract different fields (`auto_map`/`architectures` vs the full config + sanitization for `model_type` and `vocab_size: null`), and they're called in different phases of the pipeline (capability check vs construction). Folding them into a shared helper would either need a wide return type or a parameterized callback, both of which would obscure rather than clarify what each call site actually needs. Leaving as-is.

---
_Generated by [Claude Code](https://claude.ai/code)_

wilke0818:
seems reasonable. the main issues that were pointed out by claude:
  1. os.chdir is process-global. speechbrain_loading_cwd changes the process working directory, which will silently affect any other thread running concurrently during that
  block. For a single-threaded use case this is fine, but if someone ever calls from_hparams from multiple threads simultaneously (e.g. loading two SpeechBrain models in
  parallel), they'll get CWD races. A comment warning about this would be prudent, or alternatively keeping it to the known-bad CWD paths only.
  2. The _emotion_head_kind call happens twice in the happy path — once in classify_emotions_from_speech and once implicitly when building the model key. Each call fetches the
  config from HuggingFace Hub. For the two known checkpoints this is fast (config-only), but it's still two round-trips on every inference call until the result is cached. The
  function result could be memoized.
  3. The _resolve_apply_softmax function also calls AutoConfig.from_pretrained, meaning a third config fetch per inference call (again until caching kicks in). This is a minor
  inefficiency but worth noting.
  4. The test for the existing discrete model (test_speech_emotion_recognition_discrete) still requires a GPU (gpu_device fixture) and no test covers the new HuBERT/WavLM
  registry entries in practice. The Phase 1 unit tests are good but they mock everything — there's no integration-level confidence that the HuBERT/WavLM entries in
  _BASE_REGISTRY actually work on real checkpoints.
  5. The plan doc (specs/plan.md) being committed to the repo is a somewhat unusual convention. It's useful for posterity, but embedding it in the working tree means it'll
  accumulate and diverge from reality over time. A reference to a GitHub issue or discussion thread would age better.

satra:
Thanks for the careful review — all five points addressed (or, in #5's case, defended) in #515 which stacks on this branch:

1. **CWD race** — `speechbrain_loading_cwd` now serializes via a module-level `threading.Lock`. Docstring documents the trade-off: parallel SpeechBrain init across threads now blocks rather than races; processes are the escape hatch.
2. & 3. **Redundant `AutoConfig.from_pretrained` calls** — introduced `_load_config_cached(model)` returning `(typed_config_or_None, raw_dict_or_None)` memoized on `(path, revision)`. The four call sites (`_get_ser_type`, `_emotion_head_kind`, `_resolve_apply_softmax`, `_classify_wav2vec2_speech_cls_ser`) now share one round-trip. Added `test_load_config_cached_round_trips_once` to lock that in.
4. **No HuBERT/WavLM integration confidence** — partial fix: three parametrised unit tests verify `_resolve_base("wav2vec2"|"hubert"|"wavlm")` resolve to real `transformers` classes (catches `_BASE_REGISTRY` typos at test-collection time). Real-checkpoint integration tests are still future work — happy to file an issue if you want to track.
5. **Plan doc in `specs/`** — skipped for consistency with PR #513's pattern (`specs/20260506-.../tasks.md`). Genuinely open to changing it if maintainers prefer issue-based docs; let me know.

---
_Generated by [Claude Code](https://claude.ai/code)_

satra:
Follow-up #515 just landed on this branch (squash-merged as 834be0c). The diff on this PR now reflects all five fixes from your review:

| # | Resolution |
|---|---|
| 1 (CWD race) | `speechbrain_loading_cwd` serializes via module-level lock; docstring warns to use processes for parallel SpeechBrain init |
| 2 + 3 (redundant `AutoConfig.from_pretrained`) | `_load_config_cached(model)` memoizes typed config + raw dict per `(path, revision)`. Per-key locking (one `threading.Lock` per cached entry) so concurrent loads of *different* models don't serialize — mirrors `dependencies.ensure_hf_model`'s `_HeartbeatLock` strategy. The 4 call sites now share one network round-trip |
| 4 (no HuBERT/WavLM tests) | 3 parametrised `_resolve_base` smoke tests catch `_BASE_REGISTRY` typos at test-collection time without downloading checkpoints. Real-checkpoint integration tests still future work — happy to file a separate issue if you want to track |
| 5 (plan doc convention) | Kept (consistent with #513); open to changing if you'd prefer issue-based docs |

CI on the latest commit is green; ready when you are.

---
_Generated by [Claude Code](https://claude.ai/code)_

wilke0818:
seems good

satra:
adding the gpu tests also to check the venv stuff (i think the image does have cuda 12.8 so this should still pass there).

PR #512, opened 2026-05-09
comparison & uncertainty stage for analyze_audio.py

satra:
## Summary

Adds a fourth stage to `scripts/analyze_audio.py` that consumes the per-task outputs from the existing pipeline and emits comparison artefacts in four classes:

1. **Raw vs enhanced** — per (task, model) timeline of where the same model gave a different answer on the raw input vs the speech-enhanced input.
2. **Within-stream** — model-pair disagreements per pass (pyannote vs Sortformer; six C(4,2) ASR pairs; AST vs YAMNet).
3. **Cross-stream** — ASR-implied speech-presence vs diarization, AST/YAMNet "Speech" subtree vs diarization, ASR-derived ARPAbet phonemes (via g2p_en) vs PPG argmax sequence.
4. **Uncertainty** — per-region confidence harvested from the underlying models (Whisper avg_logprob, AST/YAMNet top-1 score + entropy), aggregated via configurable `--uncertainty-aggregator` (default `min`).

Outputs (all additive — `--skip comparisons` produces bit-identical legacy output):
- `<run_dir>/<pass>/comparisons/<task>/<a>_vs_<b>.parquet` (within-stream + raw_vs_enhanced)
- `<run_dir>/<pass>/comparisons/cross_stream/...parquet`
- `<run_dir>/disagreements.json` — top-N ranked by combined uncertainty
- `<run_dir>/timeline.png` — multi-row aligned plot showing diarization speakers, ASR token spans, scene-window strips, and per-comparator disagreement bars colored by combined uncertainty
- LS bundle: one fixed-enum Labels track per (pass, comparison_kind, task, model_pair); ASR-vs-ASR pairs get a sibling TextArea with WER + both transcripts

New CLI flags: `--skip-comparisons {raw_vs_enhanced,within_stream,cross_stream}`, `--cross-stream-win-length` (0.2 s), `--cross-stream-hop-length` (0.1 s), `--uncertainty-aggregator {min,mean,harmonic_mean,disagreement_weighted}`, `--phoneme-disagreement-threshold` (0.50), `--speech-presence-labels`, `--asr-reference-model`, `--diarization-boundary-shift-ms` (50.0), `--disagreements-top-n` (100).

## Spec workflow

[...]

PR #516, opened 2026-05-13
subprocess_venv: route torch/torchaudio via CUDA-aware PyTorch index

satra:
## Summary

Subprocess-venv backends (`nemo-canary-qwen`, `nemo`, `qwen-asr`) were resolving `torch` and `torchaudio` through PyPI's default index. On hosts with system CUDA newer than the PyTorch default-wheels CUDA (e.g. CUDA 12.9 against `cu128`), the two libraries could end up compiled for different CUDA toolchains and break their ABI at import time.

The shared `ensure_venv` helper in `src/senselab/utils/subprocess_venv.py` now:

- Probes host CUDA once per venv build via `nvidia-smi` (default header `CUDA Version: X.Y`) → `nvcc --version` (`release X.Y`) → none.
- Picks a matching PyTorch wheel index from a static map (`cu128` / `cu126` / `cu124` / `cu121` / `cpu`), preferring the highest entry `<=` host CUDA.
- Routes `uv pip install` through `--index-url <chosen> --extra-index-url https://pypi.org/simple` so torch/torchaudio go through the PyTorch index while non-torch packages still resolve from PyPI.
- Records the resolved index in the venv marker file (`.senselab-installed` gains a `torch_index` field). Markers from the pre-fix code (no `torch_index` key) auto-rebuild on first run after upgrade.
- On wheel-not-found failures, wraps the install error into a `SenselabCudaCompatibilityError` carrying host CUDA + attempted index + failing packages + recommended action.
- Wipes the half-built venv directory on any install failure so the next run starts clean.

Operator override: set `SENSELAB_TORCH_INDEX_URL` to use a custom URL verbatim (internal mirrors, forced CPU fallback, unusual hosts). Empty string is treated as unset.

Reusable across the three current subprocess-venv backends and any future one with no per-backend code change. In-source warnings landed near each backend's `*_REQUIREMENTS` list to prevent future bypass.

## Test plan

- [x] `uv run pytest src/tests/utils/cuda_probe_test.py` — 15 unit tests covering both probe paths, all index-map entries, env override behavior, error message contents.
- [x] `uv run pytest src/tests/utils/subprocess_venv_test.py` — 14 integration tests covering marker-mismatch rebuild paths, install argv routing, env override short-circuit, `SenselabCudaCompatibilityError` wrapping, unrelated-error pass-through, `_classify_uv_failure` unit cases, parameterized regression test across all three real backend requirement lists.
- [x] `uv run pre-commit run --all-files` — ruff + mypy + codespell + TOML/YAML/JSON formatters all clean.
- [ ] **Out-of-band**: validate happy path on a real CUDA 12.9 host per `specs/20260512-204619-fix-canary-cuda-conflict/quickstart.md`. Project CI has no CUDA 12.9 runner — this validation is a release-blocker checklist item, not a per-PR gate.

## Spec docs

Full spec, plan, research, contracts, data-model, and quickstart under `specs/20260512-204619-fix-canary-cuda-conflict/`.

# Version

Published prerelease version: `1.3.1-alpha.31`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  #### Authors: 3
  
</details>


wilke0818:
Claude comment on comparison to #517: One thing worth flagging in review (if you want to): the static map currently tops out at cu128 and the picker falls back to "highest ≤ host" — that means a cu13.x
  host quietly gets cu128, which is forward-compatible but not what the regex result "13.0" might lead a reader to expect. Either document explicitly ("we deliberately don't
  add cu130 until PyTorch publishes a stable cu130 index") or let the picker raise on host > max-mapped. Minor.

Otherwise this seems strictly better than #517 minus one small change that I will PR separately

wilke0818:
The bug isn't just probe-fall-through — it's that even with the right cu128 index explicitly named, uv puts --extra-index-url pypi in higher precedence and PyPI's wheels win
  for both torch and torchaudio. The override doesn't actually fix anything.

  ## PR #516 — `--index-url cuda --extra-index-url pypi` doesn't actually route `torch`/`torchaudio` through the CUDA index

  Hi Satra — testing #516 on ORCD, I hit a failure mode that I think makes the override path non-functional in addition to the probe being too aggressive. Three findings,
  stacked.

  ### 1. Probe falls through to CPU on hosts where `nvidia-smi` is installed but no GPUs are visible

  On the host I tested (interactive shell on a compute node, no SLURM `--gres=gpu` allocation), `/bin/nvidia-smi` is installed and exits 0, but its output is just:

  No devices were found.
  
  No `CUDA Version: X.Y` header (that line is per-device). The regex doesn't match, `nvcc` is not installed cluster-wide, so `pick_torch_index()` lands on `cpu`. Marker
  confirms:

  ```json
  "torch_index": { "tag": "cpu", "url": "https://download.pytorch.org/whl/cpu", "source": "static-map" }
```
  This affects login nodes, GPU nodes without a job allocation, CI runners with nvidia-smi but no GPU, scheduler hosts that submit work elsewhere. Build-once-reuse-many
  semantics means a build host without visible GPUs silently produces a venv that breaks every GPU runtime host that reuses it.

  Suggestion: distinguish source="present-unknown" (nvidia-smi exited 0 but version unparseable) from source="none" (nvidia-smi failed), and have pick_torch_index map the
  former to the highest static-map entry (cu128) instead of CPU. cu128 is forward-compatible with any driver >= 12.8, which covers basically all current HPC. CPU is too
  aggressive a fallback — it silently demotes downstream GPU use with no warning.
  
  2. Setting SENSELAB_TORCH_INDEX_URL=https://download.pytorch.org/whl/cu128 does NOT actually route torch through cu128

  This is the bigger issue. With override set, the install command is:

  uv pip install \
    --index-url https://download.pytorch.org/whl/cu128 \
    --extra-index-url https://pypi.org/simple \
    'nemo_toolkit[asr,tts] @ git+https://github.com/NVIDIA/NeMo.git' \
    'torch>=2.8,<2.9' 'torchaudio>=2.8,<2.9' \
    pyarrow<18 matplotlib soundfile safetensors numpy

  Expected: torch-2.8.0+cu128 from the named primary index. Actual:

  torch-2.8.0+cu129.dist-info       (from PyPI — has +cu129 local tag)
  torchaudio-2.8.0.dist-info        (from PyPI — no local tag, bundled with cu128 internally)

  → cu128/cu129 ABI mismatch:

  different CUDA versions. PyTorch has CUDA version 12.9 whereas
  TorchAudio has CUDA version 12.8.

  The cause: --extra-index-url takes precedence over --index-url in uv (opposite of pip). uv docs note this: "uv treats --extra-index-url as a list of extra indexes that take
  precedence over --index-url." So PyPI is consulted first for every package, and since PyPI has both torch and torchaudio, the CUDA index is effectively never used for these
  packages. Verified empirically — when I dropped --extra-index-url and ran uv pip install --index-url https://download.pytorch.org/whl/cu128 --force-reinstall --no-deps torch 
  torchaudio, uv correctly installed matched +cu128 wheels.

  3. Structural fix that resolves both issues

  Split the install into two steps:

  # Step 1: torch + torchaudio from the chosen CUDA index ONLY (no --extra-index-url)
  uv pip install --python <venv-python> \
    --index-url <chosen_index> \
    'torch<spec>' 'torchaudio<spec>'

  # Step 2: everything else from default PyPI (no CUDA index)
  uv pip install --python <venv-python> \
    <rest_of_requirements>

  Properties:

  - Step 1's --index-url is unopposed by --extra-index-url, so the CUDA index is genuinely primary. Both wheels come from it, matched.
  - Step 2 sees torch and torchaudio already installed and satisfied; uv doesn't reconsider them.
  - The CUDA index governs only the two packages it's designed for. Setuptools, NeMo, pyarrow, etc. resolve from PyPI as normal — no stale-wheel-on-CUDA-index issues.
  - Works regardless of whether the probe is exactly right; even a cu126-on-cu130-host mismatch yields a self-consistent venv that runs via driver forward-compat.

  Workaround that does work today (band-aid)

  Force-reinstall torch+torchaudio in-place after ensure_venv returns:

  uv pip install --python <venv-python> --force-reinstall --no-deps \
    --index-url https://download.pytorch.org/whl/cu128 \
    'torch>=2.8,<2.9' 'torchaudio>=2.8,<2.9'

  Reuses the rest of the venv (NeMo + transitive deps) and swaps just the two affected wheels. ~5 seconds. Not appropriate for the PR — too much wasted bandwidth on first build
   — but unblocks downstream testing.

  Happy to send a follow-up PR with the two-stage split + the present-unknown probe state if useful.

satra:
send a PR (the workaround seems reasonable).

PR #518, opened 2026-05-14
subprocess_venv: two-stage install to fix torch/torchaudio CUDA-tag s…

wilke0818:
## Summary

  Follow-up to #516. The `--index-url <cuda> --extra-index-url pypi` pattern doesn't actually route `torch` / `torchaudio` through the chosen PyTorch wheel index in uv.
  Splitting the install into two stages (CUDA-index for torch+torchaudio, then default PyPI for everything else) makes the routing real, eliminates the cu128/cu129 ABI mismatch
   this PR was meant to fix, and keeps every behavioral guarantee #516 introduced.
  
  ## Why

  While testing #516 on ORCD, the canary venv consistently rebuilt into the same broken state #516 is trying to prevent:

  ```
  different CUDA versions. PyTorch has CUDA version 12.9 whereas
  TorchAudio has CUDA version 12.8.
  ```
  
  Reproduced even with `SENSELAB_TORCH_INDEX_URL=https://download.pytorch.org/whl/cu128` explicitly set:

  ```bash
  $ ls ~/.cache/senselab/venvs/nemo-canary-qwen/lib/python3.12/site-packages/ \
      | grep -E "^(torch|torchaudio)-[0-9].*dist-info$"
  torch-2.8.0+cu129.dist-info       # from PyPI — has +cu129 local tag
  torchaudio-2.8.0.dist-info        # from PyPI — no local tag, internally cu128
  ```

  The marker reported `torch_index.url == https://download.pytorch.org/whl/cu128` (override was honored at the routing-decision layer), but the install argv still landed both
  wheels from PyPI.

  ### Root cause

  uv treats `--extra-index-url` as taking **precedence over** `--index-url` (opposite of pip). Quoting uv's own docs: *"uv treats `--extra-index-url` as a list of extra indexes
   that take precedence over `--index-url`."* So in #516's install command:

  ```
  uv pip install \
    --index-url https://download.pytorch.org/whl/cu128 \
    --extra-index-url https://pypi.org/simple \
    <requirements>
  ```

  PyPI is consulted first for every package. PyPI has both `torch` and `torchaudio`, so the CUDA index is effectively never used for them. The two wheels come from PyPI with
  mismatched local-version tags, and the env override path is non-functional too.

  Verified directly: running `uv pip install --index-url https://download.pytorch.org/whl/cu128 --force-reinstall --no-deps torch torchaudio` (no `--extra-index-url`) correctly
   installed matched `+cu128` wheels into the same venv.

  ## The fix

  Split the install into two stages.

  **Stage 1** — `uv pip install --index-url <chosen> torch torchaudio` with **no** `--extra-index-url`. The chosen CUDA index is unambiguously primary; both wheels and their
  `nvidia-cuda-runtime-cu12` transitives come from the same toolchain. Any version pin in the caller's `requirements` (e.g. `torch>=2.8,<2.9`) flows through verbatim; backends
  that don't pin torch (e.g. qwen) fall back to bare names.
  
  **Stage 2** — `uv pip install <rest of requirements> safetensors numpy` with no index flags. uv resolves against default PyPI, sees `torch` / `torchaudio` already installed
  and satisfying any pin, and leaves them alone. The PyTorch index now governs only the two packages it's designed for, and stale wheels on the CUDA index for utilities like
  setuptools / pyarrow stay out of the picture.
  
  `torchaudio` is no longer re-appended in Stage 2 — it's installed in Stage 1, and re-listing it without an index flag would let uv consider replacing the matched wheel with
  PyPI's tagless one.

  ## Behavioral guarantees preserved from #516

  - Marker schema (`torch_index` field with `tag` / `url` / `source`).
  - Cache-hit fast path (marker compares `requirements` + `torch_index.url`).
  - Env override (`SENSELAB_TORCH_INDEX_URL`) — flows into Stage 1 unchanged.
  - `SenselabCudaCompatibilityError` wrapping on wheel-not-found errors, now scoped to Stage 1 where it semantically belongs (Stage 2 hits default PyPI, so a failure there
  isn't a CUDA-compat surface).
  - Pass-through of unrelated `CalledProcessError`s.
  - Half-built-venv cleanup on either stage's failure.
  - Host-CUDA probe diagnostic surface.
  - Cross-backend routing (canary, nemo, qwen all hit the same `ensure_venv`).

  ## Tests

  The 14 existing `subprocess_venv_test.py` tests are kept; 4 needed call-count or argv-shape updates for the new two-stage form, and the single install-argv test was split
  into three more specific ones:

  - `test_stage_one_pins_torch_and_torchaudio_to_chosen_index` — Stage 1 names only `--index-url`, no `--extra-index-url`, and includes the torch + torchaudio specs from
  `requirements`.
  - `test_stage_one_uses_bare_names_when_requirements_omit_torch` — qwen-style backend (no explicit torch pin) still gets torch + torchaudio installed in Stage 1 so transitives
   don't pull them in later from PyPI.
  - `test_stage_two_uses_default_pypi_and_includes_ipc_deps` — Stage 2 carries neither `--index-url` nor `--extra-index-url`, includes `safetensors` + `numpy`, and does NOT
  re-list `torchaudio`.

  The cross-backend regression test (`test_all_three_subprocess_backends_route_through_same_torch_index`) now asserts both that Stage 1 routes through the chosen index AND that
   Stage 2 does *not* — guarding against any future re-introduction of `--extra-index-url` on the torch install. The `SenselabCudaCompatibilityError` wrapping test and the
  unrelated-error pass-through test were unchanged (they fire on the first `uv pip install` call, which is now Stage 1 — semantically correct).
  
  All 30 unit/integration tests pass:

  ```
  $ uv run pytest src/tests/utils/subprocess_venv_test.py \
                  src/tests/utils/cuda_probe_test.py --noconftest -q
  ..............................                                           [100%]
  30 passed, 2 warnings in 9.28s
  ```

  ## Out-of-band verification

  Reproduced the bug, applied this fix, and verified end-to-end on an ORCD compute node:
  
  - Cleared `~/.cache/senselab/venvs/nemo-canary-qwen` to force a fresh build.
  - Re-ran a canary loader with no env override set, on a non-GPU interactive shell where the probe correctly picks `cpu`.
  - Stage 1 installed matched `torch-2.8.0+cpu` and `torchaudio-2.8.0+cpu` (clean CPU path — no PyPI fallback leaking cu128/cu129).
  - Stage 2 installed NeMo from PyPI without disturbing torch/torchaudio.
  - Canary transcribed audio without `RuntimeError`. CPU path proven functional.

  Inside a `srun --gres=gpu:1` allocation the probe picks `cu128` from the static map and Stage 1 installs matched `+cu128` wheels — same outcome, GPU-capable.

  ## Test plan

  - [x] `uv run pytest src/tests/utils/subprocess_venv_test.py src/tests/utils/cuda_probe_test.py --noconftest` — all 30 tests pass.
  - [x] Fresh canary venv build with `SENSELAB_VENV_CACHE` redirected to a throwaway dir; verified torch + torchaudio wheels carry matching local-version tags after the
  install.
  - [x] `temp/asr.py` runs canary end-to-end against a real audio clip; no `RuntimeError`, transcript returned.
  - [ ] Cross-check on a fresh cu13x-driver host that the matched-tag outcome holds under the probe path (no override).

### DIFF
--- a/src/senselab/audio/tasks/speech_to_text/qwen.py
+++ b/src/senselab/audio/tasks/speech_to_text/qwen.py
@@ -30,15 +30,16 @@ _QWEN_REQUIREMENTS = [
     # Pin to a known-good release; bump intentionally as Alibaba publishes new
     # versions of the wrapper. 0.0.6 is the first version that exposes both
     # Qwen3ASRModel and Qwen3ForcedAligner via from_pretrained on PyPI.
-    #
-    # NOTE: torch / torchaudio are NOT named here directly — they come in
-    # transitively (qwen-asr depends on them, and ``ensure_venv`` always
-    # appends ``torchaudio`` as an IPC dep). The CUDA-aware index routing
-    # in ``ensure_venv`` is what guarantees those transitive installs use
-    # a matching toolchain pair. If you ever add a direct ``torch`` /
-    # ``torchaudio`` pin to this list, keep going through ``ensure_venv``
-    # so the routing stays intact.
     "qwen-asr==0.0.6",
+    # ``qwen-asr`` pulls ``torch`` + ``torchaudio`` transitively, but pin
+    # them explicitly here so ``ensure_venv``'s CUDA-aware Stage 1 routes
+    # both wheels through the matched PyTorch index. Without these pins
+    # the transitives would resolve from default PyPI in Stage 2 and
+    # could split across mismatched ``+cu`` local-version tags (the
+    # ``torch==X+cu129`` / ``torchaudio==X`` ABI mismatch this PR
+    # exists to prevent).
+    "torch>=2.8,<2.9",
+    "torchaudio>=2.8,<2.9",
 ]
 _QWEN_PYTHON = "3.12"
 
--- a/src/senselab/utils/subprocess_venv.py
+++ b/src/senselab/utils/subprocess_venv.py
@@ -198,6 +198,18 @@ def ensure_venv(
 ) -> Path:
     """Create or reuse an isolated virtual environment.
 
+    Whether the CUDA-aware two-stage install fires is decided by the
+    contents of ``requirements`` itself: any ``torch`` or ``torchaudio``
+    spec triggers the probe + Stage-1 install via the chosen PyTorch
+    wheel index. Backends that declare neither — including the genuinely
+    torch-free case (e.g. a venv that only consumes a pure-Python GitHub
+    repo) — skip the probe and the CUDA index entirely and do a single
+    install pass against default PyPI. Backends that need ``torch`` /
+    ``torchaudio`` (including via a transitive dep) MUST pin them in
+    their own ``_REQUIREMENTS`` so the CUDA routing applies — otherwise
+    Stage 2's transitive resolution against PyPI can split them across
+    mismatched local-version tags.
+
     Args:
         name: Unique identifier for this venv (e.g., "coqui", "ppgs").
         requirements: List of pip install specs (e.g., ["coqui-tts~=0.27"]).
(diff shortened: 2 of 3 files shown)


satra:
the only consideration is that if a subprocess venv does not require torchaudio (or relies only on torchcodec), this forces it. especially as this approach is meant not just for legacy environments but also when the environment needs something from simply a github repo.

wilke0818:
Added a `requires_torch: bool = True` parameter to `ensure_venv`. Default keeps the existing behavior so all current callers (canary, nemo, qwen, yamnet, coqui, ppgs, sparc)
  are unchanged. Passing `requires_torch=False` collapses the install back to a single pass:

  - Skips the `nvidia-smi` / `nvcc` probe entirely (so login nodes / no-GPU shells don't spend the few seconds shelling out).
  - Skips Stage 1 — no torch / torchaudio install, no CUDA index in the argv.
  - Skips the IPC `torchaudio` append (the `safetensors` + `numpy` IPC deps still get installed since those are pure-Python and cheap everywhere).
  - Marker omits the `torch_index` field; if the same venv name is later called with `requires_torch=True`, the existing URL-mismatch clause invalidates the cache and rebuilds
  correctly. Covered by `test_requires_torch_false_then_true_invalidates_cache`.

  So a future backend that consumes a pure-Python GitHub repo can declare:

  ```python
  ensure_venv("my-backend", ["some-repo @ git+https://..."], python_version="3.12", requires_torch=False)

  and pay none of the multi-GB CUDA wheel cost. Test coverage in test_requires_torch_false_skips_probe_and_stage_one asserts the probe is never invoked and the single install
  argv has neither --index-url nor torchaudio.

  Worth flagging that the current all_reqs = [..., "safetensors", "numpy", "torchaudio"] line predates this PR — the unconditional torchaudio append goes back to the original
  ensure_venv implementation. This patch is the first place that lets a caller opt out of it. If we want to be more aggressive (e.g. drop torchaudio from the default IPC list
  and only add it when a backend explicitly needs it for FLAC encoding), that's a separate refactor — happy to do it as a follow-up if you'd rather change the default than gate
   it behind a flag.

satra:
but torch doesn't always require torchaudio, does it? i thought it was removed as a requirement. so the only time this comes in is when some other thing requires it.

wilke0818:
Right — torch doesn't pull torchaudio (the dep direction is the other way). The unconditional `torchaudio` append in `ensure_venv` is a senselab IPC convention from [#446 
  (`14ab40c7`)](https://github.com/sensein/senselab/commit/14ab40c7), added "for FLAC audio encoding" because the IPC layer serializes Audio objects as `.flac` via
  `torchaudio.save()` / `torchaudio.load()`.

  So your point is sharper than what `requires_torch=False` actually covers in this PR. That opt-out handles the *no-torch-at-all* case, but you're flagging the orthogonal one:
   torch-yes, torchaudio-no — a backend that uses torch but never moves Audio through FLAC IPC (or uses torchcodec for its own decoding) still pays for torchaudio today.

  Honest fix is to drop the unconditional append from `ensure_venv` and require each backend to declare `torchaudio` in its own `_REQUIREMENTS` only when it actually does Audio
   IPC. Looking at current callers:

  - **canary, nemo** — already pin `torchaudio>=2.8,<2.9` explicitly, so unaffected.
  - **qwen** — relies on `qwen-asr`'s transitive `torchaudio` dep; would need an explicit line.
  - **yamnet, coqui, ppgs, sparc** — currently rely on the implicit append; each would need to add `torchaudio` explicitly *if* its IPC path round-trips Audio via FLAC, or stay
   silent if it doesn't.

  Two ways to land that:
  
  1. **In this PR.** ~20 lines across the four affected backends plus one base-line removal from `ensure_venv`. Keeps the CUDA fix and the IPC cleanup together since they're
  closely related, but mixes two behavioral changes in one diff.
  2. **As a follow-up.** Clean blame separation — this PR stays focused on the CUDA-tag split, the next one cleans up the IPC convention.

  Mild preference for (2) since each PR's effect is then isolated and bisectable, but (1) is straightforward if you'd rather not see two PRs back-to-back. Which way?

PR #523, opened 2026-06-04
20260527 151905 speaker profile embedding

wilke0818:
# Speaker profile embedding for `analyze_audio`

**Base:** `alpha` · **Feature branch:** `20260527-151905-speaker-profile-embedding`

## Summary

Adds a standalone **speaker-profile** stage that runs before `analyze_audio`. For a subject, it pools per-window speaker embeddings across all of that subject's files, clusters them, and stores the **dominant cluster's centroid per model** as a contamination-tolerant profile artifact. `analyze_audio` can then optionally consume that profile to flag windows likely containing **another voice** (or the **wrong/absent subject**), and to emit a **target-speaker recording-quality** indicator — as raw, decision-ready signals integrated into the existing outputs.

The profile is a **signal producer, not a gate**: no PASS/REVIEW verdict and no operating-point policy are embedded. The recall-biased operating point and cross-recording metric calibration belong to a separate downstream triage spec that consumes these signals.

The design **reuses existing workflow code** — `embeddings.extract_per_window_embeddings`, `cluster_pass_speakers` (dominant-cluster selection), `calibrate_cosine_uncertainty` / `_empirical_calibration_band`, and the shared speech-mask + cosine helpers. The genuinely new work is cross-file pooling into a persisted artifact, leave-one-file-out scoring, a HuggingFace WavLM SV backend, and the `analyze_audio` integration.

## Review response (reuse + "move behind the identity signal")

Following maintainer review, a verified pass addressed both points; details in `research.md` and tasks Phase 8.

**Reuse — duplication consolidated.** The leaf primitives were already shared, but the *orchestration* and a few helpers were duplicated. Now one code path each:
- `presence.reference_grid_and_speech_mask` — used by both the identity-axis clustering (`compute.py`) and the profile build gate (`build.py`), so they apply one speech definition (FR-002) and the redundant per-model mask recompute is gone.
- `embeddings.cos_sim` / `cos_dist` — one canonical cosine helper replacing three private copies.
- one canonical `DEFAULT_SPEECH_PRESENCE_LABELS` (was hand-copied between `analyze_audio` and the profile build).

**"Behind the identity signal" — the profile is now a real voter (option C).** Previously the injected `speaker_profile/*` votes were *inert* — `aggregate_identity` never read them. Now the per-bucket profile **other-voice consensus is folded into the identity uncertainty** as a *reference-based* voter ("is this the enrolled subject?") alongside the reference-free who-talks voters, and the identity rows are re-aggregated after injection, so the **identity parquet + disagreements ranking + Label Studio bundle reflect the profile**. Target-quality stays its own claim (hybrid). The profile-unique signals it adds — which the reference-free axis structurally cannot produce:
- a **continuous `subject_dominance`** / wrong-subject measure (catches "the dominant voice isn't the enrolled subject");
- an **independent presence-gated other-voice flag**, scored on a **scene-derived voice-presence gate** (speech/babble/conversation, foreground *or* background) so a non-subject voice in an otherwise quiet stretch is caught while cough/breath/silence are excluded;
- a **diarization-overlap corroborator** that raises other-voice where 2+ speakers are active — exactly the mixed-speaker windows where the profile's own distance is unreliable.

All profile outputs are emitted as **continuous, composable atoms** (paired with `profile_confidence`, calibrated within-profile only); the discrete `flag` is a derived layer.

## Cross-stage cache reuse (FR-015) — delivered

`extract_per_window_embeddings` now takes an opt-in `cache_dir` and caches each model's window list per `(audio signature, model, window/hop)` with a **caller-agnostic** key, so `build_speaker_profile` and `analyze_audio` (and re-runs, and the leave-one-file-out sibling re-extraction) **reuse embeddings instead of re-running the GPU models**. Delivered *without* the high-blast-radius keying swap that originally deferred Phase 6: embeddings was an *uncached* task, so adding a new caller-agnostic cache touches **no existing task's keying** — no forced global cache miss. float32↔JSON round-trips bit-exactly, so a cached result equals a freshly-computed one (verified). `build_speaker_profile --cache-dir` is now active.

## What's included

**New `speaker_profile` workflow package** (`src/senselab/audio/workflows/speaker_profile/`)
- `build.py` — per-file speech-window extraction (shared gate), cross-file dominant-cluster aggregation → L2-normalized per-model centroid + empirical calibration band, confidence policy (`ok`/`low`/`insufficient`), per-file keep/drop records, optional same-session weighting, `build_speaker_profile()` entrypoint, opt-in `cache_dir`.
- `compare.py` — per-window scoring vs the profile on the short detection grid, consensus fusion, voice-presence gate, diarization-overlap corroborator, continuous `subject_dominance`, leave-one-file-out, target-quality indicator.
- `io.py` — JSON artifact load/save (versioned schema, atomic write, forward-compatible reader); `types.py`; `constants.py` (named/documented/configurable thresholds); `cache.py` (caller-agnostic cache helpers); `doc.md`.

**WavLM speaker-embedding backend** — `tasks/speaker_embeddings/wavlm.py` + `api.py` dispatch (HuggingFace `WavLMForXVector`, default `microsoft/wavlm-base-plus-sv`, 16 kHz mono, graceful degradation). Default consensus **ECAPA + ResNet-TDNN + WavLM**.

**`analyze_audio` integration** — `--speaker-profile` / `--profile-other-voice-threshold`; profile folds into the identity axis (option C) + a per-pass `speaker_profile.json` sidecar + `single_speaker`/`quality` summary sub-signals; leave-one-file-out auto-applied (matched by audio content hash); embedding cache wired via `--cache-dir`. Supporting changes in `audio_analysis/{embeddings,presence,compute,aggregate,global_summary}.py`.

**New CLI** — `scripts/build_speaker_profile.py`. **Synthetic fixtures** — `scripts/gen_synthetic_test_audio.py` (SpeechT5, MIT, run-once) + committed 16 kHz FLAC. **Docs** — `scripts/docs/speaker_profile_with_analyze_audio.md`, module `doc.md`, the spec set.

## Non-breaking guarantee (SC-006)

A run with **no `--speaker-profile` is byte-identical to baseline** — the feature is opt-in and zero-impact when off. When a profile *is* supplied, it now intentionally moves the speaker-identity/quality surface (the identity uncertainty + votes, the `single_speaker`/`quality` headlines + `identity_axis_mean`, the disagreements / Label Studio ordering, the sidecar) — option C feeds the identity axis on purpose. Everything else (presence/utterance rows, `raw_vs_enhanced` deltas, AST/YAMNet/features/PII/ASR) stays byte-identical. `regression_test.py` enforces exactly this. *(Earlier this test asserted "purely additive"; it's been narrowed to the real invariant above.)*

## Testing

- Unit/integration: `build_test`, `compare_test` (incl. consensus, voice-presence gate, diar-overlap corroborator), `io_test`, `cache_test`, `conftest` composers, `success_criteria_test` (real ECAPA+ResNet on the fixtures, SC-002/003/004/005), `regression_test` (SC-006), `embeddings_test`, `global_summary_test`, `aggregate_test`, and `analyze_audio_test` (incl. the overlap-mask builder).
- GPU-verified on a compute node: full suite green; **cached == uncached centroids bit-exact**; SC-006 byte-identity holds; quality gates (`ruff`/`mypy`/`pytest`) pass.

## Scope (deferred — see `spec.md` Clarifications)

- **Optional per-window centroid confidence weighting (T028b):** deferred; SQUIM-gated trust identified as the right lever.
- **Non-subject cough/breath attribution & task-aware contextualization:** deferred to the triage/metric spec (speaker embeddings are out-of-distribution on non-speech; emit the raw scene/diar ingredients now).
- **Triage/gating + cross-recording rank-orderable metric calibration:** out of scope by design; this PR emits raw signals only.

## Empirical characterization (synthetic; advisory, in `research.md`)

Thresholds are **provisional and configurable** — synthetic-derived, not validated on real recordings.

- **Robustness ↔ discrimination trade-off:** WavLM is noise-robust but weak at fine discrimination; ECAPA/ResNet are discriminative but noise-fragile (the error-decorrelation rationale for the consensus).
- **Same-gender / similar-timbre blind spot:** a known limitation of profile-similarity with the default models (consensus ~0.16 same-gender vs ~0.67 cross-gender) — a model-capacity floor; the lever is a stronger discriminator (WavLM-Large SV) or the diarization-overlap corroborator, not windowing.
- **Window length is an operating-point knob, not a discriminability knob:** detection stays short (1.0/0.5) to catch brief + same-gender intrusions; build stays at 2.0/1.0.
- **Speech enhancement before scoring does not help** (as a fix or an embedding probe); SQUIM-gated trust is the path for noise-vs-other-voice.
- **WavLM** is the decorrelated noise-robust consensus member; wav2vec-BERT 2.0 / WavLM-Large SV recorded as future candidates (the latter needs a new backend; the former has no off-the-shelf SV head).

## How to use

```bash
# 1. Build one profile per subject (embeddings ready), across all the subject's files
uv run python scripts/build_speaker_profile.py \
    --files-from subject_filelist.txt \
    --subject-id sub-001 \
    --output    derivatives/speaker_profile/sub-001/profile.json \
    --cache-dir derivatives/analyze_audio_cache

# 2. Analyze a recording with the profile attached (same --cache-dir reuses the embeddings)
uv run python scripts/analyze_audio.py recording.wav \
    --speaker-profile derivatives/speaker_profile/sub-001/profile.json \
    --cache-dir derivatives/analyze_audio_cache
```

See `scripts/docs/speaker_profile_with_analyze_audio.md` for the per-subject batch workflow, leave-one-file-out behavior, and the build/detect window rationale.


satra:
why use wavlm for this? and not the top3 speaker embedding models from gasser's work?

satra:
also why not reuse the sliding window embedding extraction and clustering already in the analyze script. it seems to not be reusing a bunch of components already there. may want to ask it to refactor and reuse, or move things behind the identity signal

wilke0818:
> why use wavlm for this? and not the top3 speaker embedding models from gasser's work?

I don't think you ever sent me what those 3 were. this is a wavlm finetuned on sv that I was hoping would be better. It is noise resistant much more so than ecapa and resnet, but it's embeddings are the best for disentangling

wilke0818:
> also why not reuse the sliding window embedding extraction and clustering already in the analyze script. it seems to not be reusing a bunch of components already there. may want to ask it to refactor and reuse, or move things behind the identity signal

I'll have to take a look. I thought it was

wilke0818:
> also why not reuse the sliding window embedding extraction and clustering already in the analyze script. it seems to not be reusing a bunch of components already there. may want to ask it to refactor and reuse, or move things behind the identity signal

Thanks — both points are right, and digging in confirmed them, so I'm reworking accordingly.
  
**Reuse (done).** The leaf primitives were already shared (`extract_per_window_embeddings`, `cluster_pass_speakers`, `calibrate_cosine_uncertainty`, `_empirical_calibration_band`, the speech mask), but the *orchestration* and a few helpers were duplicated. Fixed and GPU-verified:
- one shared `reference_grid_and_speech_mask(...)` now used by both the identity-axis clustering (`compute.py`) and the profile build gate (`build.py`) — same speech definition by one code path, and it removes a redundant per-model mask recompute;
- one canonical cosine helper (`embeddings.cos_sim`/`cos_dist`) replacing three copies;
- one canonical `DEFAULT_SPEECH_PRESENCE_LABELS` (was hand-copied between `analyze_audio` and the profile build).

These are behavior-preserving (regression / compute-axis / embedding tests all pass).

**"Move behind the identity signal" (in progress).** You're right that the profile votes were inert — `aggregate_identity` never read the `speaker_profile/*` keys, so they only rode the identity parquet/ranking, not the decision; the real signal reached the headline via a separate `single_speaker` fold. The wrinkle is semantic: the identity axis is *reference-free* (how many speakers / where do they change), while the profile is *reference-based* (is this the enrolled subject). So rather than a pure "move," the plan is to make the profile a genuine **reference-based voter** that feeds `aggregate_identity` (so it actually moves the identity uncertainty), while keeping the genuinely-distinct pieces separate:
- profile's per-window **other-voice / subject-identity** signal → folded into the identity axis as a voter (this *will* change the identity uncertainty when a profile is supplied — the SC-006 "purely additive" guarantee narrows to its real invariant: a run with **no** profile is byte-identical to baseline);
- a **continuous** "wrong/absent subject" measure (`subject_dominance`) — the profile uniquely catches "the dominant voice isn't the enrolled subject," which the reference-free axis can't;
- an **independent presence-gated flag** retained for the recall case where identity isn't being measured (e.g. a background voice in an otherwise non-speech region) — scoring gated on *voice presence* (scene speech/babble labels) so background/secondary voice is caught while cough/breath aren't mislabeled as other-voice;
- **overlap** leans on diarization's overlap detection, not profile-distance (mixed-window embeddings are unreliable there);
- target-quality stays its own claim.

Embedding recompute (build re-embeds, and leave-one-file-out re-extracts siblings) is the deferred FR-015 cross-stage cache; folding it in with this.

**WavLM.** It's there as the *decorrelated, noise-robust* consensus member, not the discriminator — our sweeps have ECAPA carrying fine discrimination and WavLM near-zero on same-gender but far more noise-robust, which is exactly the error-decorrelation rationale. On the top-3 from Gasser's work / wav2vec-BERT 2.0: happy to swap if you point me at them — w2v-BERT 2.0 looks like a candidate to *replace* WavLM as the robust member (it reportedly beats WavLM but trails ECAPA, so it wouldn't lift discrimination), but it has no off-the-shelf SV head so it's a new backend rather than a checkpoint swap. For the same-gender gap specifically the more direct lever is a stronger SV model (WavLM-Large SV). Noted as candidates; current ECAPA+ResNet+WavLM stands for now. Gasser's work also specifically doesn't try the wavlm-base-plus-sv which actually has some finetuning for weighting which layers to pull from and a more complex SV head, whereas Gasser trains a simpler head. How this compares I am not sure, but I will leave as is for now since we are still relying on ECAPA as the main discriminator.

Proceeding with the integration above — flag anything you'd rather I not change before it lands.

PR #525, opened 2026-06-11
fix(speech-enhancement): pick speechbrain loader by model name

wilke0818:
The default enhancement model (speechbrain/sepformer-wham16k-enhancement) is a SepFormer separation checkpoint, but the loader always tried SpectralMaskEnhancement first — which fails on it (compute_stft/hparams mismatch), logged a misleading error, wasted a load attempt, then fell back to SepformerSeparation. Select the inference class directly from the model name instead (sepformer -> SepformerSeparation, else SpectralMaskEnhancement): one deterministic load, fail loudly on a genuinely unloadable model. Add a unit test for the selection logic.

### DIFF
--- a/src/senselab/audio/tasks/speech_enhancement/speechbrain.py
+++ b/src/senselab/audio/tasks/speech_enhancement/speechbrain.py
@@ -1,7 +1,7 @@
 """This module provides the Speechbrain interface for speech enhancement."""
 
 import time
-from typing import Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import torch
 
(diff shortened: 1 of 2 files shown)


satra:
Adding measured evidence for the case this PR's `_loader_for` docstring explicitly scopes out, since it turns out to be worse than "would garble" — senselab currently returns non-audio for seven model ids, silently.

Measured through `enhance_audios` on a 14 s recording, every SepFormer **separation** checkpoint reachable today:

| checkpoint | output length vs input | xcorr peak with input |
|---|---|---|
| `sepformer-wsj02mix` | exactly 2.000x | \|r\| <= 0.08 |
| `sepformer-wsj03mix` | exactly 3.000x | \|r\| <= 0.08 |
| `sepformer-libri2mix`, `-whamr16k`, `-wham`, `-whamr`, `resepformer-wsj02mix` | 2.000x | \|r\| <= 0.08 |

The `reshape(1, -1)` on `(batch, T, n_src)` interleaves the sources sample-by-sample, so ASR returns an empty transcript for every one of them — independent confirmation the output is not audio. De-interleaving post hoc recovers streams that match direct-call `src0`/`src1` measurements exactly, so nothing is lost, it is purely mis-shaped.

Your note in `_loader_for` — "true multi-speaker separation checkpoints ... are out of scope here, because `enhance_audios_with_speechbrain` reshapes the model output to a single waveform and would garble multiple separated sources" — is exactly right, and the loader-by-name change here is a real improvement independent of it.

**What I am doing with this, so you are not surprised:** your two commits (`677975a9`, `41904baa`) are cherry-picked onto a branch off `triage` with authorship preserved, and I am adding the part you scoped out on top — a shape-derived guard that raises rather than returning interleaved audio, plus `WaveformEnhancement` as a third loader class so `mtl-mimic-voicebank` stops failing at inference (`SpectralMaskEnhancement` rejects it, the `SepformerSeparation` fallback constructs and then dies in `separate_batch`). I will link the PR here when it is up.

The reason for a new branch rather than pushing to yours: this one targets `20260527-151905-speaker-profile-embedding` rather than `triage`, and `speechbrain.py` has moved under three merges today. Happy to close this in favour of the new one, or to keep it open if you would rather land it on its original base — your call.

satra:
Superseded by #569, which is merged into `triage`.

Both commits from this branch are in `triage` by content — `_loader_for` is at `speechbrain.py:36`, called at `:99`, with the off-convention fallback comment at `:110`. The SHAs differ because #569 was rebased and squashed.

#569 also carries a fix this branch does not: `_single_source` (`speechbrain.py:132`) refuses a multi-source separation output instead of flattening it. Seven checkpoints were returning interleaved samples as a single waveform — 2× the input length, |r| ≤ 0.08 against the reference, empty ASR transcripts. That is the failure mode `_loader_for`'s docstring warned about, now enforced rather than described.

Checked before closing: the only lines unique to this branch are *older* forms of code `triage` has since improved — `source=model.path_or_uri` predates the snapshot-path revision pinning, and `run_opts={"device": device.value}` predates `device_run_opt`. Closing loses nothing.

Thanks — the loader-by-name approach here is what shipped.

PR #527, opened 2026-06-30
fix(hf): load HF models offline-first to avoid Hub 429 rate-limiting

wilke0818:
- HF generates 429 errors due to too many model calls especially when running parallelized across many nodes
- Attempts to fix this for ASR models
- Should be applied across all HF code

### DIFF
--- a/src/senselab/audio/tasks/forced_alignment/forced_alignment.py
+++ b/src/senselab/audio/tasks/forced_alignment/forced_alignment.py
@@ -25,6 +25,7 @@ from senselab.audio.tasks.forced_alignment.data_structures import (
 )
 from senselab.audio.tasks.preprocessing import extract_segments, pad_audios
 from senselab.utils.data_structures import DeviceType, HFModel, Language, ScriptLine, _select_device_and_dtype
+from senselab.utils.dependencies import hf_offline_loading
 
 try:
     from nltk.tokenize.punkt import PunktParameters, PunktSentenceTokenizer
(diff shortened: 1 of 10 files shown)


satra:
@wilke0818 - this seems weird. this was implemented before.

wilke0818:
> @wilke0818 - this seems weird. this was implemented before.

Claude quick response:
> The problem this PR fixes: When lots of jobs load the same AI model at once, each one still "phones home" to Hugging Face to check the model version. Too many calls at once → Hugging Face throttles you (error 429) → the model fails to load. That's why Qwen kept coming back "missing."

> Why the old code didn't fix it: There was already code to download each model only once. But (a) the model-loading paths didn't actually use it, and (b) it stopped duplicate downloads, not the version check that was causing the 429. So the fix wasn't aimed at the real culprit.

Honestly I don't love this solution. I think it would probably be better to make the version check and create a cache entry that can include a TTL or something similar. Part of it I think is also maybe a larger rework making sure that all models are going through a consistent API when being used under the senselab umbrella. Hopefully already happening but I'd want to do a more thorough check and rework of that later.

satra:
indeed there should also be a TTL on the cache. somehow this PR is not passing my smell test.

wilke0818:
I agree. I think reconfiguring the current logic will be better. There's
also a major bug caught in code review that might be adding to that. I
might close this and just begin somewhat from scratch. Will be a tomorrow
issue as I am jumping on a flight.

satra:
Superseded by #550, which rewrites this from scratch — the direction you and satra both landed on in this thread.

All six call sites this PR patched are now routed through shared helpers rather than patched individually:

- `load_hf_resilient` and `hf_subprocess_env` in `src/senselab/utils/dependencies.py` (same contract as this PR's versions: all-cached gate, `also=` companions, no `base_env` mutation)
- `forced_alignment.py`, `speech_to_text/huggingface.py`, `granite.py`, `qwen.py`, `canary_qwen.py`, `nemo.py`

The mechanism changed. Instead of toggling `HF_HUB_OFFLINE` / `TRANSFORMERS_OFFLINE` around an in-process load, #550 resolves each ref to an immutable commit SHA and passes `revision=<sha>`, which hits `huggingface_hub`'s commit-hash shortcut: cached files, no network, no HEAD request. That matters because an in-process env toggle is a no-op — `huggingface_hub` freezes offline mode at import — so the in-process half of this PR likely never suppressed the version check at all; only the subprocess half did. SHA pinning works regardless of import order, and resolution itself reads the local cache first (`src/senselab/utils/model_revision.py`), with a run-scoped file-backed manifest so a fleet of jobs agrees on one commit per model.

The `revision` double-keyword collision your third commit fixed is now structural: `huggingface.py` keeps `revision` out of `pipeline_kwargs` by construction. `granite.py` passes the resolved SHA to both `AutoProcessor` and `AutoModelForSpeechSeq2Seq`.

`src/tests/utils/hf_load_coverage_test.py` is an AST guard that fails CI when a new HF load site appears without going through one of these helpers — so the property is enforced rather than re-patched each time.

One thing worth recovering: two regression tests here have no counterpart in #550 — `test_granite_forwards_revision_to_from_pretrained` and `test_hf_pipeline_forwards_revision_without_collision`. Both behaviours are correct on #550, but nothing pins them. Happy to port those over if you want them back; say the word and I'll add them.

PR #528, opened 2026-06-30
feat(canary): split long audio at pauses to avoid the ~40s truncation

wilke0818:
- Canary silently fails when number of tokens > 1024 (typically > 40s of audio)
- chunks audio based on pauses to pass separately and then reform together
- Possibly should consider alternative methods that include overlap + deduplication but currently uses non-overlapping windows for decent results


wilke0818:
Superseded by #534, which is the same Canary pause-chunking fix rebased directly onto `alpha` and fully decoupled from the offline-loading work in #527 (this PR was stacked on #527). Closing in favor of #534.

PR #535, opened 2026-07-22
feat(hf): model cache & version consistency — planning (supersedes #527)

wilke0818:
> **Draft — planning only.** This PR contains the spec-kit design artifacts (`spec.md`, `plan.md`, `research.md`, `data-model.md`, `contracts/`, `tasks.md`) for the HuggingFace model cache & version-consistency rework. No implementation code yet; the 65-task breakdown lands incrementally on this branch. Supersedes #527.

## The need

When a large batch of jobs (100+ across nodes) all load the same HuggingFace model at once, each job's `from_pretrained` still issues a **network revision-check** to the Hub to see whether its cached copy is current. The model is already on disk — nothing needs downloading — but those simultaneous *checks* trip the Hub's rate limit → **429 errors** → models fail to load or silently return empty (the "Qwen missing" symptom).

**Why the earlier attempts didn't hold:**
- **Pre-#527**: had download-once dedup, but it suppressed *duplicate downloads*, not the *version check* that actually 429s — right idea, wrong layer.
- **#527**: flipped `HF_HUB_OFFLINE` / `TRANSFORMERS_OFFLINE` around the load. But in the installed `huggingface_hub` 1.11 / `transformers` 4.55, **offline mode is captured at import time**, so setting the env var mid-process does **not** suppress the check for in-process loaders (its subprocess path worked because a child imports fresh — hence the inconsistency the maintainer flagged). It also had no version verification and no freshness policy.

## The plan

A single shared mechanism every HF-backed backend uses:

1. **Resolve once, pin by SHA.** Turn the requested ref (`"main"`/tag/branch) into an immutable commit SHA a single time, then load `revision=<sha>, local_files_only=True` — a pure filesystem read, **zero network**. This is the real 429 fix, and it's honored at call time (unlike the import-frozen env flag).
2. **Verify the version, or fail loud.** Compare the local snapshot's SHA against the requested one at the cache/hub layer — independent of whether the loader exposes a `revision` param. Wrong/missing version → clear error naming the model; never a silent substitution.
3. **Bounded freshness, coordinated.** A shared sidecar record stores `{resolved_sha, last_verified_epoch}`. Re-check the Hub only after a **7-day window** (configurable), and when due, **exactly one** check runs across all concurrent jobs (file-locked) — not one per job.
4. **Two freeze levers.** A **run-scoped freeze** holds versions steady for one long batch; a **system-level freeze** (`SENSELAB_HF_FREEZE=1`) pins versions across all runs and disables re-checks. Precedence: system ⊇ run ⊇ default.
5. **One mechanism, everywhere.** Migrate all 23 HF-backed loading sites (transformers, SpeechBrain, pyannote, NeMo, sentence-transformers, GLiNER) onto it — incrementally, behind unchanged public APIs, each with a behavior-preserving test — so no backend keeps bespoke caching/version code and new backends inherit the guarantees for free.

The coordination core is factored as a provider-agnostic primitive (`ensure_cached_resource`), so a follow-up can extend download-once/offline reuse to non-HF sources (SPARC, Coqui, s3prl, yamnet) without re-implementation. Those are explicitly out of scope here.

## Status / next steps

- ✅ `/speckit.plan`, `/speckit.tasks`, `/speckit.analyze` complete — coverage 25/25 requirements, 0 critical/high findings, constitution-aligned (TDD, uv, documented thresholds).
- ⏭️ `/speckit.implement` — MVP first: Foundational → US1 (the 429 fix), then US2 (verification) → US3 (freshness/freezes) → US4 (migrate all backends).


wilke0818:
@satra — you were right that a model cache "was implemented before," so this rework isn't rebuilding from zero. But a close review of the **existing (pre-#527)** mechanism (`utils/data_structures/model.py` + the download-once helpers in `utils/dependencies.py`) turned up concrete defects that explain why it doesn't actually solve the 429s — and why the fix is a refactor, not another patch. Mapping each proposed change to the defect it addresses:

**1. The existing cross-process lock can corrupt the cache.** `_HeartbeatLock` "breaks" a stale lock by `unlink()`-ing the lockfile and re-acquiring. On a busy node a *live* holder whose daemon heartbeat thread just didn't get scheduled for >90s is misjudged dead; `filelock` then locks a brand-new inode at the same path, so two processes hold "the same" lock and write the same cache dir concurrently. The heartbeat only proves the thread ran, not that the download is healthy.
→ **Change:** replace it with an owner-liveness **safe lease** — only take over when the owner is provably gone, never unlink a live holder's lock (tasks T010a + test T004a).

**2. The existing mechanism never covers the load-time version check — which is the actual 429 source.** `check_hf_repo_exists → ensure_hf_model` only guarantees *download-once + existence*. The real per-load Hub HEAD/revision check happens later inside each backend's `from_pretrained`/`pipeline`/`from_hparams`, which nothing wraps. So even if every model went through the existing channel, 100 concurrent jobs would still each HEAD the Hub → 429.
→ **Change:** resolve the ref to an immutable SHA once, then load `revision=<sha>, local_files_only=True` so the loader itself makes zero network calls.

**3. #527's offline toggle doesn't work in-process.** It flips `HF_HUB_OFFLINE` at call time, but in the installed `huggingface_hub` 1.11 / `transformers` 4.55 offline mode is captured at *import* time — so the flag is a no-op for in-process loaders (it only worked in the subprocess workers, which import fresh). This is the inconsistency behind your smell-test reaction; it's why #527 is being discarded rather than extended.
→ **Change:** rely on SHA-pinned `local_files_only=True` (honored at call time), not env toggling.

**4. There is no TTL today — and you specifically asked for one.** The result-cache record stores only `{status, commit_hash}`, no timestamp; a cached model is therefore *never* re-checked (silent staleness) and a cached "not found" error persists forever.
→ **Change:** add `last_verified_epoch` + a bounded, coordinated **7-day freshness window** (configurable) on both success and error records, with a single re-check shared across concurrent jobs — plus run-scoped and system-level freezes (`SENSELAB_HF_FREEZE`) for reproducible runs. Precedence system ⊇ run ⊇ default.

**5. `ensure_hf_model`'s "resolved commit hash" can be a branch name.** It falls through to `return revision` (e.g. `"main"`) for repos without `config.json` or in offline mode, and `is_hf_model_cached` returns True under `HF_HUB_OFFLINE=1` even when the model is absent; `config.json` presence is also treated as a complete snapshot (a partial/aborted download reads as "cached" and never self-heals).
→ **Change:** derive the SHA independently (`model_info` / the `refs/<ref>` file), verify the local snapshot's SHA matches the request (loud failure on mismatch), and gate the fast path on an integrity check (`scan_cache_dir` warnings) rather than a single-file probe (T008).

**6. Adoption is genuinely incomplete — but that's not the whole story.** 20 of 31 model-loading sites don't fully use the channel today. Migrating all of them (behind unchanged public APIs, per-backend behavior tests) is part of the plan (US4) — but points 1–5 show that adoption alone wouldn't have fixed the 429s.

Full write-up in `specs/20260706-213755-hf-model-cache/research.md` (§D7 maps every defect → design response). The spec is scoped to HuggingFace-backed backends; the coordination core is factored as a provider-agnostic primitive so non-HF sources (SPARC, Coqui, s3prl, yamnet) can reuse download-once/offline in a follow-up. Would appreciate a sanity check on the freshness-window default (7 days) and the freeze-precedence model before implementation starts.

satra:
1. needs to take into account multiple processes running on the same node, or across nodes, or in subenvs. it also needs to be resistance to crashes, canceled batch jobs, and kills.
2. we don't need to be checking for a new revision on each call. we can debounce across an entire process.

PR #536, opened 2026-07-22
Scene-aware presence/quality axis + ASR transcription overhaul (CrisperWhisper 2.0, Qwen)

satra:
## Summary

Reworks the `audio_analysis` uncertainty workflow's **presence axis into a scene-aware dimension** and overhauls the **ASR / transcription** stack, driven by `SPEECH_PRESENCE_CERTAINTY_ANALYSIS.md`, a spec-kit design (`specs/20260722-175022-scene-quality-utterance/`), and validation against a Label Studio ground-truth annotation of the 4-speaker, noisy `audio_48khz_mono_16bits` clip.

All new presence signals are **additive columns** — the `presence` axis name, parquet paths, LS tracks, and `aggregated_uncertainty` are unchanged, so downstream consumers keep working.

## Scene-aware presence axis

- **Audio quality (per bucket, `[0,1]` degradation):** `quality_snr` (Brouhaha frame SNR ✕ existing DSP SNR), `quality_clip` (existing `proportion_clipped_metric`), `quality_reverb` (Brouhaha C50), `quality_bandwidth` (torch.stft spectral rolloff vs Nyquist), plus `quality_uncertainty` (estimator spread).
- **Background sound sources:** `src_{speech,people,machine,environment}` + `src_dominant`, from a checked-in AudioSet→category map over full AST/YAMNet distributions (`--scene-top-k`).
- **Fine temporal resolution:** continuous frame posteriors from `pyannote/segmentation-3.0` (+ Brouhaha VAD), per-axis grids (presence 0.1 s/0.02 s), and a `presence_confidence` / `presence_uncertainty` split; coarse voters demoted on fine grids. Long recordings handled via chunked sliding-window stitching.
- **`pyannote/brouhaha`** runs in an isolated subprocess venv (it needs the GitHub-only `brouhaha-vad` + old pyannote); loads via `ensure_hf_model`.
- **Timeline PNG** gains scene-quality and sound-source rows.

## ASR / transcription overhaul

- **CrisperWhisper 2.0 turbo** (new default) — verbatim + native word timestamps, via a platform-aware subprocess venv (ct2 on Linux x86_64, transformers elsewhere). **Granite and whisper-large-v3-turbo dropped from defaults** (kept opt-in).
- **Qwen3-ASR-1.7B** retained (native word timestamps).
- **Qwen3-ForcedAligner-0.6B** is the new default aligner for text-only ASR (e.g. canary); MMS still available via `--aligner mms`.
- Native confidence flows via `ScriptLine.score` **when a backend exposes it** (verified: none of the current three do; plumbing is ready).
- **Quality-vs-noise:** speech-appropriate SNR anchors (25→5 dB) so noisy recordings read more degraded than clean.

## Infra fixes (surfaced during validation)

- `numba>=0.60` in all NeMo/Qwen subprocess venvs (qwen-asr, canary-qwen, nemo, nemo-diarization) — fixes the `librosa→numba→llvmlite 0.36` py3.12 backtrack.
- `huggingface_hub<0.24` in the Brouhaha venv; `peft` in the canary venv; `ctranslate2` on the CrisperWhisper transformers path.
- `librosa` reverted from an explicit dep (bandwidth reimplemented on `torch.stft`).

## Validation against ground truth (`audio_48khz_mono_16bits`)

- **Transcription:** CrisperWhisper 2.0 + Qwen3-ASR both **WER 0.05** with native word timestamps; canary works (Qwen-aligned).
- **Presence:** frame acc 0.91, recall 1.0; the brief 0.38 s "Kenny" turn resolved; the ~0.4 s gap correctly high-uncertainty (both VADs detect activity).
- **Quality:** noisy clip reads SNR-degradation 0.65 (clean ≈ 0); bandwidth 0.85→0.95 rolloff fix (no longer flags full-band speech).
- **Identity:** diar models collapse on the 5 s clip (model-side); finer 0.25 s embedding clustering carries the signal.

## Status / scope

- Spec `specs/20260722-175022-scene-quality-utterance/` (spec/plan/research/data-model/contracts/tasks) committed. US1–US3 implemented; US4 (utterance token-entropy) is partial (native-confidence plumbing only, as no backend exposes it); US5 (synthetic calibration harness) deferred.
- Speaker-detection measures deferred per design discussion (transcription-cleanup prioritized).

## Caveats for reviewers

- CrisperWhisper ct2 path is validated only via the transformers backend on macOS arm64; the ct2 (Linux x86_64) path is wired but unrun here.
- Native transcription *confidence* is currently absent (the chosen ASRs don't expose logprobs); re-add a whisper if that signal is needed.
- Subprocess-venv model paths (Brouhaha, CrisperWhisper, Qwen, canary) are skipped by default CI (venv-absent skipif) and validated locally with `HF_TOKEN`.

# Version

Published prerelease version: `1.3.1-alpha.35`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - Scene-aware presence/quality axis + ASR transcription overhaul (CrisperWhisper 2.0, Qwen) [#536](https://github.com/sensein/senselab/pull/536) ([@satra](https://github.com/satra) [@claude](https://github.com/claude))
  - feat(canary): split long audio at pauses to avoid the ~40s truncation [#534](https://github.com/sensein/senselab/pull/534) ([@wilke0818](https://github.com/wilke0818))
  - ScriptLine: accept empty text as "model produced no transcript" signal [#520](https://github.com/sensein/senselab/pull/520) ([@wilke0818](https://github.com/wilke0818))
  - Feat/pii subprocess venv [#519](https://github.com/sensein/senselab/pull/519) ([@wilke0818](https://github.com/wilke0818))
  
  #### Authors: 3
  
</details>


wilke0818:
### `brouhaha` scene-quality venv fails to build on modern-CUDA hosts — `torch>=2.0,<2.3` is unsatisfiable on the selected wheel index

Running `analyze_audio.py` with the default scene-quality path on a CUDA-12.8 GPU host, the `brouhaha` subprocess venv fails to build. Scene-quality signals (SNR / reverb / clip / bandwidth) then come back null, and clips whose only quality estimator is brouhaha (e.g. low-speech breath/cough tasks) hit the `sys.exit(2)` scene-quality gate and fail the whole run.

**Root cause is the PyTorch wheel-index selection — not HF model access.** The venv dies at torch install, before the gated model is ever loaded:

- `scene_quality/brouhaha.py` pins `torch>=2.0,<2.3` / `torchaudio>=2.0,<2.3` (the pyannote-3.x era the brouhaha stack needs). This is correct for the isolated venv.
- `ensure_venv` → `pick_torch_index` selects the wheel index from the **host** CUDA (static map top entry = `cu128` on a 12.8 host) or from the `SENSELAB_TORCH_INDEX_URL` override — both resolve to `cu128` here.
- The `cu128` index has **no `torch<2.3` linux-x86_64 wheels**, so Stage-1 `uv pip install --index-url .../cu128 torch>=2.0,<2.3 torchaudio>=2.0,<2.3` is unsatisfiable and the venv is torn down.

Proof (`uv pip install --dry-run`, py3.11):

```
# cu128  → fails
torch>=2.0.0,<=2.0.1 cannot be used.
And because you require torch>=2.0,<2.3, we can conclude that your requirements are unsatisfiable.
(only wheel available for torch 2.0.1 is manylinux2014_aarch64)

# cu121  → resolves
+ torch==2.2.2+cu121
+ torchaudio==2.2.2+cu121
```

The selector keys the index on the *host* CUDA version but ignores the *venv's* torch-version ceiling. So **any backend pinning old torch (`torch<2.3`, brouhaha today) is unbuildable whenever the host resolves to cu124/cu126/cu128** — i.e. every current GPU host. This is not an override artifact — confirmed on a CUDA-13.1 node with `SENSELAB_TORCH_INDEX_URL` **unset**:

```
detect_host_cuda()            -> HostCuda(version=(13, 1), source='nvidia-smi')
pick_torch_index(env=None)    -> TorchIndex(tag='cu128', ...)   # static-map top ≤ host
uv pip install --index-url .../cu128 'torch>=2.0,<2.3'  -> unsatisfiable
```

It builds only where the host resolves to `cu121` or lower (CUDA 12.1–12.3) — which is presumably why it works in the dev environment. There is also no fallback: `ensure_venv` raises on the wheel-not-found rather than retrying an older index.

**Suggested fix:** choose the index per-venv so it satisfies that venv's torch constraint — e.g. cap the selected index to the newest one that carries a compatible torch (`cu121` → torch 2.2.x for brouhaha), or let a backend declare its own index tag alongside its requirements. cu121 wheels are forward-compatible with 12.x/13.x drivers for inference, so this doesn't regress the modern ASR venvs (which legitimately pin `torch>=2.8` and need cu128).

**Secondary blocker (only reached once the venv builds):** `pyannote/brouhaha` is a gated model — the worker needs `HF_TOKEN` in its environment and access granted on the HF account.

wilke0818:
@satra — recommended course of action on the `brouhaha` scene-quality build failure, after a code review of `utils/cuda_probe.py`, `utils/subprocess_venv.py`, and `audio/tasks/scene_quality/brouhaha.py` at HEAD (`d11f6759`). (Follows up my earlier note above with a concrete fix.)

### The issue (verified on current HEAD)
`brouhaha.py` pins `torch>=2.0,<2.3` (the pyannote-3.x era it needs). `ensure_venv` → `pick_torch_index` chooses the wheel index as the **highest static-map entry ≤ the host's `nvidia-smi` CUDA**, applied uniformly. On any CUDA-12.4+ host that resolves to cu124/cu126/cu128 — **none of which publish `torch<2.3` wheels** — so the Stage-1 `uv pip install` is unsatisfiable, the venv is torn down, scene-quality (SNR/reverb/clip/bandwidth) comes back null, and low-speech clips then hit the `sys.exit(2)` scene-quality gate.

Confirmed on a CUDA-13.1 node with the override unset: `pick_torch_index` → `cu128`; `torch>=2.0,<2.3` → *unsatisfiable*. `cu121` resolves `torch==2.2.2` and runs fine on the 13.1 driver (CUDA is backward-compatible).

### Root cause
`pick_torch_index` keys only on host CUDA and ignores the **venv's own torch-version ceiling**. It's correct for the modern backends (nemo-canary-qwen / qwen-asr pin `torch>=2.8` → cu128 ✓, their venvs build fine) and breaks only for the one backend pinning old torch. It's not a per-instance fluke — any future old-torch venv will hit it on a current GPU host.

### Recommended fix (general, no per-backend config)
In `ensure_venv`'s Stage-1 install-failure path — where a wheel-not-found currently raises `SenselabCudaCompatibilityError` — **retry down the index ladder** (`cu128 → cu126 → cu124 → cu121 → cpu`), installing against the first index that resolves the requested torch spec, and only raise if none do. Because newer drivers run older-CUDA wheels, this is safe and self-correcting for both brouhaha today and any old-torch backend later. (Optional refinement: let a backend declare a `max_torch_index`/explicit index so it skips straight to the right one instead of paying failed attempts.)

Lower-priority adjacent item: `detect_host_cuda` reports the node driver's CUDA and doesn't gate on an actually-usable/allocated GPU (no device-count / `CUDA_VISIBLE_DEVICES` check), so a GPU node with no allocation would still pick a CUDA index. Didn't cause this, but a usable-device guard would harden it.

### One flag outside the torch scope
brouhaha's worker calls raw `pyannote Model.from_pretrained(..., use_auth_token=...)` inside its subprocess venv, so it bypasses the HF-resilience layer (`retry_on_transient_error` + heartbeat filelock in `utils/dependencies.py`) that the in-process loaders use — under parallel runs its gated-model fetches 429/401/403 with no retry. If the hf-model-cache work **doesn't reach subprocess-venv model loads**, this is a second gap worth closing (pre-resolve/cache in the parent before spawning, or wrap the worker load in the same guard). Noting it only in case it's not already covered there.

For now I'm running with a local cu121 cap on brouhaha's `ensure_venv` and an isolated `SENSELAB_VENV_CACHE` (the shared `~/.cache/senselab/venvs` was getting clobbered by a concurrent run on another branch — same-name venvs, different py/reqs).

satra:
Thanks @wilke0818 — the diagnosis is exactly right, and confirmed against the code + reproduced. `pick_torch_index` (`src/senselab/utils/cuda_probe.py`) keyed the wheel index solely on the host CUDA version and had no knowledge of a venv's `torch` ceiling, so on a CUDA-12.4+ host it selected `cu128`, which publishes no `torch<2.3` linux-x86_64 wheel → brouhaha's Stage-1 install was unsatisfiable and the venv never built.

Reproduced directly with `uv pip install --dry-run` (py3.11, `x86_64-manylinux2014`):
- `cu128` → *"you require torch>=2.0,<2.3 … unsatisfiable"* (only an `aarch64` wheel for torch 2.0.1)
- `cu121` → resolves `torch==2.2.2+cu121` + matching `torchaudio`

**Fix (your first suggestion — cap to the newest compatible index):**
- `pick_torch_index` / `ensure_venv` gained an optional `max_cuda_version` ceiling; the picker now selects the newest index `≤ min(host, ceiling)`. `SENSELAB_TORCH_INDEX_URL` still wins outright, CPU hosts and modern-torch venvs (canary/qwen, `torch>=2.8` → `cu128`) are unchanged.
- brouhaha declares `_BROUHAHA_MAX_CUDA_VERSION = (12, 1)` → `cu121` (forward-compatible with 12.x/13.x drivers for inference).

**Validation:** unit coverage for the ceiling in `cuda_probe_test.py`, plus a GPU-gated regression test (`src/tests/audio/tasks/scene_quality/brouhaha_venv_gpu_test.py`) that builds the *real* brouhaha venv on a CUDA host — it caught the bug (red on `cu128`), then went green after the fix. Confirmed on an EC2 GPU runner: the brouhaha venv builds via `cu121` and the full suite passes (786 passed).

Commits: fix `90d43842`, regression test `8c2b84f7`/`82745973`.

On the **secondary blocker** — you're right that `pyannote/brouhaha` is gated. That's orthogonal to the routing bug (the venv build fails before any model load), so the smoke test deliberately isolates the venv/torch build and doesn't require `HF_TOKEN`; the model-access requirement remains for actually running the estimator.

PR #542, opened 2026-07-30
feat(scripts): local PII detection + task compliance pipeline

varun-thvar:
## Description

Adds `scripts/pii_compliance_pipeline.py` — a fully-offline triage screen for folders of `.pt` feature files carrying ASR transcripts. For each recording it answers two questions:

1. **Does this contain PII?** — names, ages over 90, dates, locations, contact details, ID numbers, and subtler quasi-identifiers that can single a participant out.
2. **Did the participant actually do the task?** — did they read the passage, sustain the vowel, answer the question, or is the recording empty / off-script?

Each file comes back **`pass`** or **`review`**, and the run ends with the list of flagged filenames. **Nothing is ever auto-deleted, auto-redacted, or auto-rejected** — it builds a triage queue and a person makes the final call.

Three files:

| File | What it is |
|---|---|
| `scripts/pii_compliance_pipeline.py` | The pipeline. Self-contained. |
| `scripts/task_reference.json` | 797 task definitions (Bridge2AI acoustic tasks) driving Tier A/B/C routing. |
| `scripts/pii_compliance_pipeline.md` | Full documentation — engines, tiers, modes, blind spots, data safety. |

## Motivation and Context

Implements the two text-level checks — **(3) PII Detection** and **(4) Task Compliance** — from the poster *"An AI Pipeline for Ensuring Audio Data Quality"* (Ng, Wilke, Johnson, Ghosh), operationalising the companion proposal *"Unified Framework for PII Detection and Task Compliance Verification in Clinical Voice Recordings."*

**PII** runs four engines in parallel and cross-validates them — a rule cascade (regex, gazetteers, spaCy NER, self-disclosed demographics, rare roles, age > 90, plus a combinatorial re-identification window), GLiNER, Presidio, and an optional local LLM. Agreement boosts confidence; disagreement surfaces the span rather than resolving it silently.

**Task compliance** identifies the task from the BIDS `task-<Name>` field of each filename, looks it up in `task_reference.json`, and routes by modality: Tier A word-error-rate against the reference prompt for scripted reads, Tier B skipped for acoustic tasks (not assessable from text — a near-empty transcript there is *expected*), Tier C LLM judge for open tasks, with an always-on transcript-quality floor underneath.

**Privacy.** Transcript text never leaves the machine — engines run in-process, and the optional LLM talks only to `localhost`. The only network traffic is a one-time download of model weights and gazetteers.

### Notes for reviewers

- **Self-contained by design.** The script imports nothing from `senselab`, and `senselab` does not import it. It needs only its companion `task_reference.json`, resolved relative to the script's own folder so the pair runs from any working directory. Every detection engine is optional and degrades with a printed notice.
- **Dependencies go in the `pii` extra, not core.** `gliner` and `wordfreq` sit alongside `presidio-analyzer` and `spacy`; `nltk` is already in `nlp`. Adding them to core dependencies would have put three heavy packages (and a duplicate `nltk`) into every senselab install for one `scripts/` tool.
- **`task_reference.json` is large (4,725 lines)** but it is *data*, not code — 797 task definitions, mostly the 720 Harvard/IEEE sentences. Tier A cannot work without it: a missing reference means every scripted read silently loses its word-error-rate check and passes as clean, so the script **stops** rather than continuing without it.
- **Configuration is one block at the top of the file** (`MODES` / `SETTINGS` / `NOTES`) — no environment variables, no config file — so it runs from an IDE with no arguments. CLI flags override for one-off terminal use.
- ⚠️ **spaCy has no wheels for CPython 3.14**, so on a 3.14 environment `--extra pii` cannot install spaCy or Presidio and the pipeline runs with two of four PII engines absent (it says so at startup). Pre-existing — `spacy>=3.7` was already in the extra — but worth knowing, and it is called out in the docs. Use 3.12/3.13 for the full engine set.

## How Has This Been Tested?

`python scripts/pii_compliance_pipeline.py --selftest` runs **127 checks** on synthetic data in a temp directory, touching no real files and never calling the LLM. Coverage:

- **Precision guards** — format-validation of structured identifiers, holiday reclassification, GLiNER windowing offsets, the common-word-NAME false-positive lever.
- **Tier A calibration** — WER banding, the absolute-error-floor and reference-coverage guards, and that neither guard can ever hide a genuine off-script read.
- **Task → reference resolution** — exact-match precedence and sibling-variant safety (`diadochokinesis-PA` must not pull in `-Pataka`).
- **Composite scoring** — the pass/review/fail bands, the fail→review collapse, and the no-assessable-check case.
- **The companion reference** — that it loads, has a uniform shape, and that a missing or unparseable file stops the run rather than silently disabling scripted checks.
- **End to end** — `.pt` folder → task identified from filename → flagged filenames out, including that reading the *wrong* script is caught (only the task reference reveals it) and that a near-empty acoustic transcript is *not* a compliance failure.

Also run against real `.pt` data and against human-reviewed ground-truth labels during development.

## Types of changes

## Checklist:
- [x] I have added tests to cover my changes. *(built-in `--selftest`, 127 checks; not wired into pytest — see below)*
- [x] My code follows the code style of this project. *(`scripts/` is excluded from ruff and mypy in this repo)*

### Open questions

1. **Should `--selftest` be wired into the pytest suite?** It is self-contained and hermetic, so a thin `src/tests/` wrapper invoking it as a subprocess would be easy. Left out for now since `scripts/` is otherwise untested here.
2. **Is `scripts/` the right home**, or would this be better as a `senselab` module with a CLI entry point? It is currently a standalone research/ops tool, which is why it lives beside `analyze_audio.py`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -191,3 +191,11 @@ src/senselab/video/tasks/pose_estimation/models/
 tutorials/audio/tutorial_audio_files/
 tutorials/video/tutorial_images/
 tutorials/video/visualize/
+
+# PII compliance pipeline outputs (scripts/pii_compliance_pipeline.py).
+# These contain participant transcripts, real subject/session IDs and raw PII spans.
+# Treat them as sensitive as the recordings themselves -- never commit them.
+pii_compliance_report.json
+pii_compliance_summary.csv
+flagged_files.txt
+scripts/eval_out/
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -55,6 +55,12 @@ repos:
   hooks:
   - id: codespell
     args: [--skip=*.ipynb, --ignore-words-list, AER]
+    # scripts/task_reference.json holds the exact sentences participants are asked to
+    # read aloud (Harvard/IEEE lists). Its wording must match what they were given --
+    # "correcting" it would give every recording of that task a permanent word-error-rate
+    # mismatch. It trips codespell on the corpus's own "busses" and on "wit" (cleverness,
+    # not a typo for "with"), so the data file is excluded rather than spell-corrected.
+    exclude: ^scripts/task_reference\.json$
     additional_dependencies:
     - tomli
 - repo: local
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -71,7 +71,12 @@ pii = [
   # spaCy model itself is not pulled by pip — install it separately:
   # uv run python -m spacy download en_core_web_lg
   "presidio-analyzer>=2.2",
-  "spacy>=3.7"
+  "spacy>=3.7",
+  # Additional engines used by scripts/pii_compliance_pipeline.py. All optional at
+  # run time — the pipeline skips any that are missing and says so — but installing
+  # the extra should give the full detector set.
+  "gliner>=0.2.28",  # HIPAA-identifier detector
+  "wordfreq>=3.1.1"  # word frequencies behind the rare-word / rare-role guards
 ]
 text = [
   "sentence-transformers>=5.1",
--- a/scripts/pii_compliance_pipeline.md
+++ b/scripts/pii_compliance_pipeline.md
@@ -1,0 +1,363 @@
+# PII Detection + Task Compliance Pipeline
+
+`pii_compliance_pipeline.py` screens a folder of `.pt` feature files containing ASR
+transcripts and answers two questions per recording:
+
+1. **Does this recording contain personally identifying information?** — names, ages over 90,
+   dates, locations, contact details, ID numbers, and subtler *quasi-identifiers* that can
+   single a participant out.
+2. **Did the participant actually do the task they were asked to do?** — did they read the
+   passage, sustain the vowel, answer the question, or is the recording empty / off-script?
+
+Each file comes back as **`pass`** (nothing worth a human's time) or **`review`** (a human
+should look at it). **Nothing is ever auto-deleted, auto-redacted, or auto-rejected** — the
+tool builds a triage queue and a person always makes the final call.
+
+**Everything runs locally.** Transcript text is only ever processed in-process (spaCy / GLiNER /
+Presidio) or sent to `localhost` (Ollama). **No transcript text leaves the machine.** The only
+network traffic is a one-time download of model weights and gazetteers — never your data.
+
+This implements the two text-level checks — **(3) PII Detection** and **(4) Task Compliance** —
+from the poster *"An AI Pipeline for Ensuring Audio Data Quality"* (Ng, Wilke, Johnson, Ghosh),
+operationalising the companion proposal *"Unified Framework for PII Detection and Task
+Compliance Verification in Clinical Voice Recordings."*
+
+## Two files, and they travel together
+
+| File | What it is |
+|---|---|
+| `scripts/pii_compliance_pipeline.py` | The pipeline. Self-contained — imports nothing from `senselab`. |
+| `scripts/task_reference.json` | 797 task definitions (Bridge2AI acoustic tasks): task → instructions + expected read-aloud prompts. Drives Tier A/B/C routing. |
+
+The JSON is resolved **relative to the script's own folder**, not your working directory, so the
+pair runs from anywhere. If it is missing the run **stops** rather than quietly skipping every
+scripted-task check — see *Known limitations*.
+
+The script is deliberately standalone: it is a research/ops tool that happens to live in this
+repo, not a `senselab` module, and nothing in the package imports it.
+
+## Install
+
+Only `torch` is strictly required (to read the `.pt` files). Every detection engine is
+optional and degrades gracefully with a printed notice if absent — run with none of them and
+you still get the rule cascade plus task compliance.
+
+```bash
+uv sync --extra pii --extra nlp
+uv run python -m spacy download en_core_web_lg
+```
+
+> **Python version.** spaCy currently publishes wheels for CPython 3.11–3.13 only, so on a
+> **3.14** environment `--extra pii` cannot install spaCy or Presidio. The pipeline still runs
+> — it prints `[spaCy] not importable` / `[Presidio] not importable` and continues on the rule
+> cascade plus GLiNER — but with two of four PII engines silently absent, which will show up as
+> lower name/location recall. **Check those two startup lines before trusting a run's numbers.**
+> Use a 3.12/3.13 environment for the full engine set.
+
+The optional local LLM additionally needs [Ollama](https://ollama.com) running on
+`localhost`, with a model pulled:
+
+```bash
+ollama pull gemma4
+```
+
+## Quickstart
+
+```bash
+uv run python scripts/pii_compliance_pipeline.py --selftest
+```
+
+That runs 127 checks on synthetic data in a temp directory and touches no real files. Then
+point it at some data:
+
+```bash
+uv run python scripts/pii_compliance_pipeline.py /path/to/pt_files
+```
+
+Or set `INPUT_FOLDER` in the CONFIG block and run the file straight from your IDE — see
+*Configuration*. Either way you get a JSON report, a CSV summary, and the flagged file list
+printed at the end.
+
+> **Sanity check before a big run:** set `MAX_FILES = 20` to process just the first 20 files.
+
+## Configuration
+
+All settings live in **one place**: the `MODES` and `SETTINGS` blocks at the top of the script.
+There are no environment variables and no config file. Command-line flags exist for one-off
+overrides, but running the file from an IDE uses exactly what is written in those blocks.
+
+`MODES` is the panel of on/off toggles — engines, LLM, precision vs recall posture, always-flag
+rules. `SETTINGS` holds paths, model names and thresholds. A `NOTES` section below them records
+*why* each default is what it is. Every run prints the modes actually in force, so an IDE run
+without a visible command line still tells you what was on.
+
+The one value you must set is `INPUT_FOLDER`. With no input path the script stops immediately
+and says so — it does not guess, and it does not scan the working directory.
+
+## Modes — pick your operating point first
+
+This is a **triage screen**, and screens have no single "best" setting. Everything below trades
+the same two quantities: **how many real problems you catch** against **how many files a human
+has to open**. Nothing here can auto-reject a recording; only the size and purity of the review
+queue changes.
+
+Numbers are from the study corpus — **894 recordings, 42 genuinely non-compliant**. Treat them
+as the shape of the trade, not as guarantees for your data.
+
+| Mode | How | Catches | Queue | Use when |
+|---|---|---|---|---|
+| **Precision-first** | *(defaults)* | ~12 / 42 | ~13 files, ~92% real | Reviewer time is scarce; you want a short list that's almost all real. |
+| **+ open-task judging** | `ENABLE_LLM = True` | **20 / 42** † | ~100 files, ~20% real † | You need free-speech / story-recall tasks checked at all. Adds ~1–2 s per file. |
+| **+ recall nets** | `HIGH_RECALL = True` | more | larger | Thin open responses and weak PII should surface too. |
+| **+ acoustic with speech** | `RECALL_FLAG_ACOUSTIC = True` | more | much larger | A vowel/breath task that transcribed as words is worth a look. |
+| **Maximum recall** | above, plus `RECALL_ACOUSTIC_MAX_CONTENT_WORDS = -1` | ~41 / 42 | ~620 files (~69%) | A missed problem is unacceptable and you will screen most of the corpus. |
+
+† Measured with a previous model (phi4); the default is now gemma4, so treat those two figures
+as indicative until re-measured. Every other row is unaffected.
+
+**Do not steer by accuracy.** Only ~5% of this corpus is non-compliant, so a screen that flags
+*nothing* scores ~95% and catches nothing.
+
+Two switches are orthogonal to that ladder — combine them with any mode above:
+
+| Switch | Effect |
+|---|---|
+| `COMPLIANCE_ONLY = True` | Skip PII detection entirely. Much faster (no spaCy/GLiNER/Presidio load). PII is reported as **skipped**, not as clean. |
+| `PRECISION_MODE = False` | Drop the PII false-positive guards. Affects PII only, never task compliance. |
+
+## How it works
+
+```mermaid
+flowchart TD
+    A[".pt file<br/>(transcript + optional metadata)"] --> B["Task identified from<br/>filename: task-&lt;Name&gt;"]
+    A --> C[PII DETECTION]
+    B --> D[TASK COMPLIANCE]
+
+    C --> C1["Rule cascade<br/>regex · gazetteers · spaCy NER<br/>honorifics · professions<br/>demographics · rare roles · age&gt;90"]
+    C --> C2["GLiNER<br/>HIPAA identifiers"]
+    C --> C3["Presidio<br/>Microsoft NER"]
+    C --> C4["Local LLM<br/>(optional)"]
+    C1 & C2 & C3 & C4 --> C5["Cross-validate + merge<br/>agreement boosts confidence"]
+    C5 --> C6{"review_worthy?<br/>always-flag rules:<br/>any NAME, any age &gt; 90"}
+
+    D --> D1["Modality → Tier<br/>scripted=A · acoustic=B · open=C"]
+    D1 --> D2["Tier A: word error rate + coverage<br/>Tier B: skipped (needs audio)<br/>Tier C: LLM judge (opt-in)"]
+    D --> D3["Transcript-quality floor<br/>empty / non-speech / filler-only"]
+
+    C6 --> E["Composite score Q, confidence C"]
+    D2 & D3 --> E
+    E --> F{"Decision"}
+    F --> G["pass"]
+    F --> H["review<br/>(human triage queue)"]
+```
+
+### Step 1 — Load the transcript
+
+Each `.pt` is loaded exactly as:
+
+```python
+data = torch.load(path, map_location="cpu", weights_only=False)
+transcription = data["transcription"]
+```
+
+Optional keys are used if present and ignored if absent: `task`/`task_id`/`task_name`,
+`reference_text`/`reference`/`target_text`, `prompt`/`instruction`/`rubric`, `tier`.
+
+### Step 2 — PII detection: several engines in parallel, then cross-validated
+
+| Engine | Role |
+|---|---|
+| **Rule cascade** | regex for structured identifiers (SSN, email, phone, URL, MRN), name/place gazetteers, spaCy NER, honorifics, professions, rare-word MISC, self-disclosed **demographics**, **rare roles**, **age > 90**, and a **combinatorial** sliding window catching re-identification from several weak details together |
+| **GLiNER** | HIPAA-identifier detector (names, dates, phone/fax, email, SSN, MRN, device/vehicle IDs, URLs, IPs, biometrics, geo) |
+| **Presidio** | Microsoft's NER-based PII analyzer |
+| **Local LLM** *(optional)* | Independent judgement on ambiguous spans, plus a "do these details *together* identify someone?" call |
+
+Engines run independently and are then merged. Where they agree, confidence is boosted
+(`cross_validated: true`); where they disagree the span is surfaced rather than silently
+resolved. Categories map onto the annotation-guideline label set: `AGE, NAME, DATE, TIME, ORG,
+IDNUM, LOC, PROFESSION, CONTACT, URL, MISC`.
+
+**The key concept — `review_worthy`.** Every detection is recorded in the JSON report, but a
+file is only *flagged* if it has at least one review-worthy span. This is the main defence
+against drowning reviewers in false alarms. A span is review-worthy when it is a confirmed
+direct identifier, **any** name, **any** age over 90, a cross-validated (≥2 engine)
+`ORG`/`LOC`/`DATE`/ID, a combinatorial cluster, or a quasi-identifier.
+
+#### Always-flagged identifiers
+
+Two types bypass every precision guard and always route to review:
+
+- **Every personal `NAME`** — regardless of whose it is or how weak the signal. A spoken name
+  belongs in the redaction queue, so even a lone common word tagged as a name (`Will`, `May`,
+  `Grant`) flags the file. Expect some NER false positives; that's the intended trade. Names
+  are still **never** auto-failed. Toggle: `FLAG_ALL_NAMES`. This changes how a detected name
+  is *treated*, not whether it is *detected* — a name the NER engines miss can't be flagged.
+- **Any age over 90** — at the top of the age distribution the population thins out enough that
+  age alone can identify someone. (HIPAA Safe Harbor treats *all* ages over 89 as identifiers —
+  set `AGE_REVIEW_OVER_YEARS = 89` for strict alignment.) Catches the phrasings ASR really
+  produces: `95 years old`, `97-year-old`, `aged 102`, spelled-out numbers (`ninety-five`,
+  `a hundred and two`), plus `in her nineties` / `centenarian`. Non-age numbers
+  (`100 percent`, `95 dollars`) are never misread as ages.
+
+#### Quasi-identifiers (the subtle cases)
+
+- **Unusually specific roles.** The motivating real example: a participant mentioning being a
+  *professional figure skater* — ethics judged that "a bit too much info". Distinguishing
+  qualifiers (*professional / Olympic / former …*) and self-described uncommon occupations flag
+  as review-worthy `MISC`. Common jobs ("I'm a teacher", "professional development") are
+  deliberately excluded.
+- **Self-disclosed demographics.** Gender / gender identity (*"I identify as nonbinary"*) and
+  race / ethnicity (*"I'm Black"*) are protected attributes that aid re-identification, so they
+  route to review as `MISC`. Deliberately high-precision: only *self-* or *person-anchored*
+  mentions fire, so *"black coffee"*, *"Chinese food"* and the rainbow passage's *"white light"*
+  all stay clean. Toggle: `DEMOGRAPHIC_PII`.
+
+### Step 3 — Task compliance: identify the task, then route it to the right test
+
+The task is read from the `.pt` **filename's** BIDS-style `task-<Name>` field
+(`sub-…_ses-…_task-Rainbow_features.pt` → `Rainbow`), looked up in `task_reference.json`, and
+classified into a **modality** that determines how — or whether — compliance can be judged from
+text alone:
+
+| Modality | Example tasks | Tier | How it's scored |
+|---|---|---|---|
+| **scripted** | Rainbow passage, Caterpillar, Cape-V, Harvard sentences | **A** | word error rate + reference coverage against the reference prompt |
+| **acoustic** | DDK, max phonation, prolonged vowel, glides, loudness, breath/cough | **B** | **skipped** — non-verbal; a near-empty transcript is *expected* and never failed |
+| **open** | free speech, picture description, story recall, fluency | **C** | LLM judge against the task's instruction (needs `ENABLE_LLM`) |
+
+Matching tolerates case, `-`/`_`/space differences, subtype variants (`diadochokinesis-PA`,
+`cape-V-sentences-3`) and a trailing `-features`. Sibling variants are resolved most-specific
+first, so `Picture-description` never silently pulls in `picture-description-option2`'s prompt.
+
+#### Tier A in detail — "did they read the script?", not "how clean is the ASR?"
+
+Those are very different bars, and conflating them is the single biggest source of false
+positives here. On clinical read-aloud speech a **WER of 0.1–0.4 is the normal cost** of
+dysphonia, accent, room noise and ASR error on a participant who read the script perfectly
+well. Someone who read the *wrong* text lands far higher — 0.6 to over 1.0. So WER is mapped
+onto the compliance score through explicit thresholds rather than used raw:
+
+| Word error rate | Score | Decision |
+|---|---|---|
+| ≤ `TIER_A_WER_REVIEW` (0.50) | 1.0 → 0.9 | **pass** — they read it; the rest is ASR noise |
+| 0.50 → `TIER_A_WER_FAIL` (0.80) | 0.9 → 0.5 | **review** — doubtful read |
+| > 0.80 | 0.5 → 0.0 | hard fail → **review** — wrong text or no read |
+
+Two further guards stop the two ways WER lies about read speech. **Both can only move a file
+towards `pass`** — neither can create a flag, so neither trades away recall:
+
+- **Absolute error floor** (`TIER_A_MIN_WORD_ERRORS`, default 4). WER is a *ratio*, so it
+  punishes short references hardest: three misheard words in a 5-word Cape-V sentence is WER
+  0.6, while the same three errors in the 64-word rainbow passage is WER 0.05. The short
+  sentence is not twelve times less compliant.
+- **Reference coverage** (`TIER_A_COVERAGE_PASS`, **off by default**). WER also counts
+  *insertions*, so someone who reads correctly then adds anything — "sorry, let me start again",
+  an aside to the researcher — can exceed WER 1.0 having said every reference word. Coverage is
+  insertion-blind by construction, so it answers *was the script read?* where WER answers *how
+  cleanly?*. **This guard is an assumption about your annotators, not about ASR**, which is why
+  it ships disabled: in the study corpus reviewers labelled two such recordings `partial`, so
+  enabling it turned two catches into misses while rescuing no false positives. Set it to
+  `0.90` and re-measure to find out which way your own labels go.
+
+Every Tier A result reports `wer`, `coverage`, `word_errors`, `ref_words`, the
+substitution/deletion/insertion split, and which guard (if any) fired — so a reviewer can see
+*why* a read passed or was queued.
+
+> **The reference text has to be right.** Tier A is only as good as `task_reference.json`: if a
+> task's prompt there isn't the text participants were actually given, *every* recording of that
+> task gets a large, near-identical WER and the whole batch lands in the queue. That signature —
+> one task, many subjects, the same WER — means a wrong reference, not non-compliant
+> participants. Check it before retuning thresholds.
+
+**Transcript-quality floor (always on).** Independent of the task, a degenerate transcript means
+the task was never performed — a signal needing no metadata at all:
+
+| Transcript | Outcome |
+|---|---|
+| empty / whitespace only | hard compliance failure → **review** |
+| only non-speech markers (`[noise]`, `(coughing)`) | hard compliance failure → **review** |
+| only filler vocalizations (`uh`, `um`) | hard compliance failure → **review** |
+| 1–2 content words, or mostly filler | **review** (near-degenerate) |
+| real, substantive content | treated as a valid attempt |
+
+This floor is **suppressed for acoustic tasks** (a breath recording has little text by design)
+and wherever a Tier A/C score is authoritative, so those are never falsely failed.
+
+### Step 4 — Score and decide
+
+`Q = Σ wᵢ·cᵢ·sᵢ` (weighted quality) and confidence `C = weighted_mean(c)·(1 − λ·std(c))`, so
+**disagreement between engines widens the review band** rather than being averaged away.
+
+The three-way logic (`pass` / `review` / `fail`) is computed and then **collapsed to two-way:
+`fail` is folded into `review`**, so the final decision is only ever `pass` or `review`. Nothing
+is auto-rejected. The underlying severity is still reported per file:
+
+| Field | Meaning |
+|---|---|
+| `flagged_by_pii` | The file has review-worthy PII. |
+| `pii_confirmed` | Stronger: a confirmed direct identifier (`HARD_GATE_PII_LABELS`, default `{IDNUM, CONTACT, URL}`). |
+| `flagged_by_compliance` | A task / quality / protocol issue. |
+| `compliance_fail` | Stronger: a hard failure (degenerate transcript, missing required step, or score below cutoff). |
+
+A spoken `NAME` is deliberately **not** in `HARD_GATE_PII_LABELS`: it is redactable PII that
+belongs in the review queue, not a reject.
+
+### Optional: interview-style protocol steps
+
+Set `PROTOCOL_SPEC` to a spec JSON and the pipeline additionally checks whether required
+protocol steps happened (consent obtained, identity confirmed, …) via a keyword/phrase
+gazetteer → sequence verification → the LLM on residual steps. This needs curated
+`canonical_phrases`, which is why it lives in an optional file. Most per-recording task
+compliance does not need it.
+
+## Outputs
+
+| Output | Contents |
+|---|---|
+| `OUTPUT_JSON` | Full report: every detection, span offsets, per-tier scores, masked preview, and `summary.flagged_files`. |
+| `OUTPUT_CSV` | One row per recording: decision, Q/C, PII labels, WER/coverage, quality status. |
+| `OUTPUT_FLAGGED_LIST` | Just the flagged file paths, one per line — the plain-text handoff to whatever does review next. |
+
+The flagged list is also printed at the end of every run with a short reason per file
+(`pii:confirmed`, `compliance:fail`, …).
+
+## Known limitations — read before trusting the numbers
+
+**This tool reads text only.** Poster steps needing raw audio — **(1) Audio & Environment
+Quality**, **(2) Unconsented Speaker Detection**, and **Compliance Tier B** — are reported as
+`skipped: requires audio`. That leads to three honest blind spots:
+
+1. **Acoustic-task non-compliance is largely invisible, and this is measurable.** A participant
+   who did too few repetitions, or didn't attempt the task, produces a near-empty transcript —
+   *identical* to a perfectly compliant breath or vowel recording. Meanwhile compliant
+   DDK/counting tasks often *do* transcribe as words. On the study corpus: of ~200 acoustic
+   recordings whose transcript is `too_short`, about 5% are the ones humans marked
+   non-compliant; of ~76 with an `empty` transcript, roughly 1. **No transcript-level feature
+   separates them** — every threshold that catches a miss drags in ~20 compliant files. That is
+   why `RECALL_FLAG_ACOUSTIC` is off by default, and why the fix is audio-level analysis
+   (energy, repetition counting, duration), not better text thresholds.
+2. **"Answered the wrong question" needs the LLM.** Whether a free-speech response actually
+   addresses the prompt is a semantic judgement only the Tier C judge can make. By default open
+   tasks are carried by the transcript-quality floor alone, which sees only *whether* someone
+   spoke. Every open-task miss on the study corpus was a fluent, substantial response that
+   simply didn't do what was asked.
+3. **Subtle contextual PII can be unflaggable.** Some human `context` judgements rest on details
+   no span-level detector can see.
+
+**If the LLM can't be reached** the run prints `[LLM] Ollama not reachable …` or `model … is not
+installed` and continues *without* Tier C — which looks exactly like a fast successful run.
+Check that line. With the LLM on, budget ~1–2 s per file once warm, so tens of minutes for a few
+hundred recordings; pilot with `MAX_FILES` first. The pipeline uses **one** model on purpose:
+Ollama holds a limited set resident, so rotating across several forces a disk reload on nearly
+every call, and that reload tax dominates runtime.
+
+## Data safety
+
+- The pipeline only ever **reads** `.pt` files. Masked previews live in the report and are never
+  written back; nothing is auto-redacted.
+- **Outputs are sensitive.** The JSON and CSV contain real subject/session IDs and PII-derived
+  content — treat them as sensitive as the recordings and keep them out of version control.
+- **No participant paths in source.** `INPUT_FOLDER` ships empty; the script stops rather than
+  guessing.
+- **`--selftest` fabricates its own synthetic data** in a temp directory, so the tool can be
+  verified without touching real recordings.
--- a/scripts/pii_compliance_pipeline.md
+++ b/scripts/pii_compliance_pipeline.md
@@ -67,7 +67,11 @@ ollama pull gemma4
 uv run python scripts/pii_compliance_pipeline.py --selftest
 ```
 
-That runs 127 checks on synthetic data in a temp directory and touches no real files. Then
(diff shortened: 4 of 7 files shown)


wilke0818:
# Review — #542 `feat(scripts): local PII detection + task compliance pipeline`

Reviewed at `b08f4f3a` against the `alpha` merge-base. Every correctness finding in §1 was
reproduced by running the code, not read off.

## Posture

The **task-compliance engine is the valuable, novel contribution** — Tier A (WER against a
reference prompt), Tier B (acoustic/non-verbal), Tier C (LLM judge for open-ended tasks).
senselab has nothing like it, and it answers a question `analyze_audio` cannot:
*did the participant do the task they were asked to do?* The precision work on the PII side
(gazetteers, zipf rarity, rare-role, demographic/age, combinatorial risk, the hard-gate
corroboration tiers) is likewise better than anything currently in the repo.

Two things reshape the structural recommendation, and I want to be upfront that both cut
against the "self-contained by design" note in the PR description:

1. `src/senselab/audio/workflows/audio_analysis/pii.py` **is already merged** (#519) and
   already wired into `analyze_audio.py:2394-2407`. It runs Presidio + GLiNER for the same
   purpose (§2.1).
2. #543 lands `AudioPlus` + a `MetadataProvider` protocol that generalizes task/speaker
   context resolution — which is most of what this pipeline needs and currently derives from
   filenames (§4).

On the dependency-weight argument specifically: reusing `pii.py` **reduces** the dependency
surface rather than growing it, because its detectors live in an isolated subprocess venv —
the host process needs neither `presidio-analyzer` nor `gliner` nor `spacy` installed. That
also sidesteps the CPython 3.14 spaCy-wheel problem flagged in the PR description, since the
venv pins 3.13. Details in §2.1.

**Suggested sequencing:** land §1 on this PR — those are defects in the current behaviour and
several are independent of any restructuring. Treat §2–§4 as the answer to open question 2
("is `scripts/` the right home?") and do them as a follow-up. §1.3 and §1.11 are the answer
to open question 1, and they're less comfortable than the PR text assumes.

---

## 1. Correctness — blocking

### 1.1 `scripts/pii_compliance_pipeline.py:2182` — an LLM verdict suppresses the empty-transcript hard fail

`suppress_quality = (modality == "acoustic") or task_has_score`. For an OPEN task with
`ENABLE_LLM=True`, `_tier_c` always produces a score, so `task_has_score` is True,
`check_quality` returns with `hard_fail=False`, and the deterministic empty-transcript
signal never reaches `_compliance_check_score`.

Reproduced: open task, `transcription == ""`, LLM replies `{"completed": true,
"confidence": 0.9}` → `decision="pass"`, `compliance_fail=False`, file absent from
`flagged_files`. With `--no-llm`: `decision="review"`, `compliance_fail=True`. Even a
low-confidence *"not completed"* (conf 0.3) downgrades the hard fail to plain review.

The NOTES block promises the LLM "can lower a score into review but never rescue a hard
fail". Tier C bypasses that guarantee. The suppression is right for the *soft score* — a
prolonged-vowel task shouldn't be failed for a thin transcript — but a structurally empty
transcript is not a modality artifact, it's a broken recording.

```python
# In check_quality(), keep the hard fail even when suppressed:
    if suppress:
        return dict(q, applies=False, score=None, confidence=None,
                    hard_fail=q["hard_fail"],      # was: False
                    raw_status=q["status"])
```

and in `_compliance_check_score`, OR it in unconditionally rather than only inside the
`score is not None` branch:

```python
    if quality is not None:
        hard_fail = hard_fail or bool(quality.get("hard_fail"))
        if quality.get("score") is not None:
            scores.append(quality["score"])
            confs.append(quality["confidence"])
```

If a genuinely empty acoustic transcript should *not* hard-fail, gate on the emptiness
subtype rather than on modality — but say so in the NOTES, because right now the code and
the doc disagree.

### 1.2 `scripts/pii_compliance_pipeline.py:1536,1606` — separator-stripping normalization binds a file to the wrong task's prompt

`_norm_task` deletes all non-alphanumerics, which destroys the only boundary distinguishing
`list-7-3` from `list-73-1`. `_resolve_reference_keys` then takes the longest key that is a
prefix of the token, and `"harvardsentenceslist731".startswith("harvardsentenceslist73")`
is True.

Reproduced against the shipped `task_reference.json`:

| filename token | matched key | reference used | result |
|---|---|---|---|
| `Harvard-Sentences-List-73-1` | `harvard-sentences-list-7-3` | "He ran half way to the hardware store." | WER 1.125, `compliance_fail=True` |
| `Cape-V-sentences-11` | `cape-V-sentences-1` | wrong sentence | flagged |
| `diadochokinesis-PAT` | `diadochokinesis-PA` | wrong syllable | flagged |

In every case `ambiguous_match=False` and `reference_missing=False`, so nothing warns. The
failure mode is the worst possible one for this tool: **a compliant participant is reported
as having read the wrong script.** A false `compliance:fail` costs more reviewer trust than
a false pass.

This is the sibling-variant contamination `_resolve_reference_keys`' docstring is written to
prevent, arriving from the other direction. The self-test covers it one way — the PR
description lists *"sibling-variant safety (`diadochokinesis-PA` must not pull in
`-Pataka`)"*, and that assertion does pass. The unguarded direction is the token being
*longer* than the key: `diadochokinesis-PAT` **does** pull in `diadochokinesis-PA`.

Fix: normalize separators to a single delimiter instead of deleting them, and require prefix
matches to land on a delimiter boundary.

```python
def _norm_task(s):
    """Lowercase, collapse every run of separators to a single '-'.

    Separators are COLLAPSED, not deleted: deleting them makes 'list-7-3' and
    'list-73-1' both normalize to a common prefix, so the prefix fallback below
    silently scores a file against a sibling task's reference text.
    """
    return re.sub(r"[^a-z0-9]+", "-", (s or "").lower()).strip("-")
```

```python
        # Keys that are a prefix of the token -> token is the more specific name.
        # The trailing '-' is load-bearing: it forces the match to end on a segment
        # boundary, so 'list-7-3' is not a prefix of 'list-73-1'.
        prefixes = [k for k, nk in self._ref_index.items()
                    if nk and nt.startswith(nk + "-")]
```

...and the same boundary condition on the `supersets` branch. Verified this keeps the
intended matches working (`Rainbow-Passage` → `rainbow`) and turns all three rows above into
`reference_missing=True`, which is the honest answer.

I checked the shipped table itself: **797 keys, 0 normalized collisions.** The bug only
fires on filename tokens absent from the snapshot — i.e. on any corpus that has drifted from
it, which is the normal case over a study's lifetime. Worth a regression test that asserts
an unknown sibling variant yields `reference_missing`, not a match.

Fix this here regardless — but note that §4 removes the whole failure class for b2ai data by
reading the reference prompt from each recording's own sidecar instead of matching a table.

### 1.3 `scripts/pii_compliance_pipeline.py:466` — missing `wordfreq` inverts the precision guards and reds the self-test

`_zipf` returns `0.0` on ImportError with the comment *"no wordfreq → don't suppress (favor
recall)"*. But 0.0 means "maximally rare", and both call sites treat rarity as *evidence of
PII*, so the fallback doesn't favor recall — it inverts two guards:

- `_name_hard_gate_eligible`: `common = ... and _zipf(...) >= NAME_COMMON_WORD_ZIPF` is
  always False, so a two-engine common-word NAME false positive becomes hard-gate eligible.
- `_phrase_min_zipf` returns `None`, so `_SOFT_QUALIFIER_RE` fires on everything.

`wordfreq` is only in the new `pii` extra, so this is the state after a default `uv sync`.
Reproduced in the repo venv:

```
$ python scripts/pii_compliance_pipeline.py --selftest --no-spacy --no-gliner --no-presidio
[FAIL] C: common-word NAME -> REVIEW (all names flagged), still not confirmed
[FAIL] C: 'professional development' -> not flagged (common role word)
SELF-TEST FAILED        # exit 1
```

**This is half the answer to open question 1.** A thin `src/tests/` wrapper invoking
`--selftest` as a subprocess would be red on a clean checkout, because CI installs the repo
without the `pii` extra. (The other half is §1.11.)

Two changes: make the sentinel unambiguous, and make the guards fail *safe* (toward
suppression) rather than toward the hard gate.

```python
def _zipf(word):
    """Zipf frequency, or None when wordfreq is unavailable.

    None is NOT 0.0: 0.0 means "measured, maximally rare", which every caller
    reads as evidence FOR a PII hit. Callers must treat None as "unknown" and
    fall back to their precision-safe branch.
    """
    try:
        from wordfreq import zipf_frequency
    except ImportError:
        return None
    return zipf_frequency(word.lower(), "en")
```

Then at the `_name_hard_gate_eligible` call site, an unknown-frequency single token should
not be hard-gate eligible on frequency grounds alone, and the self-test should either skip
the frequency-dependent assertions or declare `wordfreq` a hard dependency of the `pii`
extra it already lives in. My preference: **hard-require it** — it's pure-Python with a
small wheel, and the alternative is two silently different detection behaviours in the field
depending on how someone installed.

### 1.4 `scripts/pii_compliance_pipeline.py:1295` — `--high-recall` manufactures *confirmed* PII from hallucinated spans

`--high-recall` sets `PRECISION_MODE = False` (line 3330), which does two independent things:

1. `postprocess_entities` returns early (line 510), skipping structured-identifier format
   validation — the guard that kills word-only "identifiers";
2. `merge_pii` takes `if not (HARD_GATE_REQUIRE_CORROBORATION and PRECISION_MODE):
   hard_gate_eligible = is_strong` unconditionally.

Together: a single GLiNER `device identifier` hit on the word **"cough"** → `confirmed=True`,
`pii_confirmed=True`, reason `pii:confirmed`. On this corpus that's not a hypothetical span.

The docs promise recall mode only *ADDS review flags*. Here it promotes a known-hallucination
class straight to the hard-gate tier. Two knobs are being driven by one flag; they should be
separate:

```python
# Format validation is a CORRECTNESS check, not a precision/recall tradeoff: a
# "device identifier" with no digits is wrong in either posture. Keep it on always;
# PRECISION_MODE governs only the confidence-threshold guards below it.
def postprocess_entities(entities, source_text):
    out = []
    for e in entities:
        ...
        if cat in ("IDNUM", "CONTACT", "URL") and not _valid_structured_identifier(span_text, cat):
            continue
        if not PRECISION_MODE:
            out.append(e)
            continue
        ...
```

and in `merge_pii`, keep the corroboration requirement independent of `PRECISION_MODE` —
high recall should widen what reaches *review*, never what reaches *confirmed*. If the
intent really is "recall mode auto-fails more", say so explicitly in NOTES [RECALL],
because the current text says the opposite.

### 1.5 `scripts/pii_compliance_pipeline.py:2241` — `torch.load(..., weights_only=False)` on every scanned file

This is a triage tool pointed at folders of `.pt` files produced elsewhere. torch ≥ 2.6
defaults `weights_only=True`; this opts out, so a malicious or corrupted `.pt` gets code
execution during a compliance scan.

The only fields read are `data["transcription"]` plus string metadata, all of which load
fine under the safe default.

```python
    try:
        data = torch.load(path, map_location="cpu", weights_only=True)
    except Exception as e:  # noqa: BLE001
        return None, None, f"failed to load .pt file: {e}"
```

If some real feature files fail under `weights_only=True`, the fix is an explicit
`--trust-input-pickles` opt-in with a printed warning, not a silent default. Note this
finding disappears entirely under the §4 integration path, where transcripts arrive from
`analyze_audio` and no pickle is loaded at all.

### 1.6 `scripts/pii_compliance_pipeline.py:148` — `ENABLE_LLM = True` contradicts every place that documents the default

The module docstring ("The local LLM (off by default)"), `--enable-llm` help ("Off by
default; slow"), `--no-llm` help ("Force the LLM off (the default)"), `print_active_modes`
("Set ENABLE_LLM = True."), and the `.md` mode table (which lists `ENABLE_LLM = True` as the
*non-default* row) all say off.

An out-of-box IDE run therefore probes Ollama and, per the PR's own measurements, takes the
review queue from ~13 files to ~100 at ~20% precision. Set `ENABLE_LLM = False`.

### 1.7 `scripts/pii_compliance_pipeline.py:177` — a developer's personal path ships as the default input folder

`INPUT_FOLDER = "/Users/varunthvar1/Desktop/sample_data"`, while `pii_compliance_pipeline.md`
states *"**No participant paths in source.** `INPUT_FOLDER` ships empty; the script stops
rather than guessing"* and the docstring says *"with no input path the script stops"*.

A bare run dies with `'/Users/varunthvar1/Desktop/sample_data' is not a directory`, and the
`args.folder is None` branch at line 3403 is unreachable. Worse, on any machine where a path
of that name happens to exist, the tool silently scans the wrong corpus.

Set `INPUT_FOLDER = ""` and add a test that the no-argument path exits with the documented
message.

### 1.8 `scripts/pii_compliance_pipeline.py:178` — `MAX_FILES = 10` silently caps every run

The docs treat `MAX_FILES` as an opt-in pilot knob (*"Sanity check before a big run: set
`MAX_FILES = 20`"*), implying the default is "all". Pointed at the 894-recording corpus, the
default run yields a `flagged_files` list drawn from the first 10 files and a summary that
reads as a complete screen.

Set `MAX_FILES = None`, and — independent of the default — record the cap **in the report**,
not just in the console banner:

```python
    report["summary"]["truncated"] = bool(max_files and len(all_files) > max_files)
    report["summary"]["files_available"] = len(all_files)
```

A JSON artifact that gets read six months later has no console scrollback.

### 1.9 `scripts/pii_compliance_pipeline.py:188,2211` — reports of raw transcripts default into the repo working tree

`OUTPUT_JSON = "pii_compliance_report.json"` / `OUTPUT_CSV = "pii_compliance_summary.csv"` are
bare CWD filenames. `analyze()` sets `masked_preview` to the **unmasked** transcript whenever
no PII was found (`build_masked_preview(transcript, pii) if pii else transcript`), and every
merged entity carries the raw `text[start:end]`.

Running the documented `uv run python scripts/pii_compliance_pipeline.py /path/to/pt_files`
from the repo root drops both files into the working tree. Neither name is in `.gitignore`,
so a `git add -A` commits participant transcripts and raw PII spans. The `.md` warns "keep
them out of version control" while the defaults steer the other way.

Three cheap mitigations, all worth taking:
- default the outputs into the input folder (or require `--output-dir`), not CWD;
- add both filenames to `.gitignore` regardless;
- keep `masked_preview` masked in both branches — when `pii` is empty the mask is a no-op, so
  returning `transcript` is only an optimization, and it's the branch that makes the field
  unsafe to treat uniformly. A `--include-transcripts` opt-in for the full text would be
  better still.

### 1.10 `scripts/pii_compliance_pipeline.py:2020,2296` — `summary["fail"]` is structurally always 0

`composite_score` folds `decision == "fail"` into `"review"` before returning, so no record
can ever carry `decision == "fail"` — but `print_summary` prints `fail: 0 (hard gate)` on
every run, including ones with confirmed SSNs and empty recordings, and `write_csv`/the JSON
expose the same dead field.

The two-way collapse is a defensible product decision; printing a permanently-zero counter
next to it is not. Either drop the `fail` line and print the real signals:

```
    review: 13   (of which hard-gate: 4 — pii_confirmed 3, compliance_fail 1)
```

or keep `decision` three-way in the data and collapse only at the presentation layer. I'd
prefer the latter — downstream consumers keying off `summary.fail` currently read "no hard
failures" on a corpus full of them, and `pii_confirmed`/`compliance_fail` are easy to miss.

### 1.11 `scripts/pii_compliance_pipeline.py:2520` — `--selftest` is neither hermetic nor engine-independent

The PR describes the 127 checks as *"self-contained and hermetic"*, *"touching no real files
and never calling the LLM"*. The LLM part holds; the rest doesn't. Parts A/B/J construct a
real `Pipeline`, so `_load_engines()` runs: `load_name_gazetteer()` calls
`nltk.download("names")` (network) and, where installed, `GLiNER.from_pretrained(...)` pulls
weights from HF.

Separately, several Part A assertions are engine-dependent — `A: clean.pt -> pass, no PII`
asserts `num_entities == 0`, which only holds when spaCy/GLiNER/Presidio are *absent* — so
the pass/fail outcome varies with what happens to be installed. Only Part J is annotated as
engine-independent. Combined with the CPython 3.14 spaCy-wheel gap already noted in the PR,
that means the same command legitimately produces different verdicts on 3.12 and 3.14.

So: **yes to open question 1, but the self-test needs splitting first.** `--selftest`
deterministic, hermetic, all engines explicitly off, gazetteer stubbed — that half wraps
cleanly in pytest. `--selftest-full` keeps the engine-dependent checks as an opt-in. §3.4
makes this substantially easier by putting the core somewhere pytest already looks, which is
also why I'd answer question 1 and question 2 together rather than separately.

### 1.12 `scripts/pii_compliance_pipeline.py:3335` — `--flag-all-pii` is silently inert

The MODES comment (line 155) says of all three recall knobs: *"setting any of them switches
it on for you, with a printed note"*. `RECALL_FLAG_ALL_PII` is only read inside
`if HIGH_RECALL:` in `merge_pii` (line 1339), but the `stray` interlock checks only
`--flag-acoustic` and `--flag-all-open`. So `--flag-all-pii` alone changes nothing and prints
nothing — precisely the "I turned that on and it caught nothing" failure the note block
exists to prevent.

```python
        stray = [f for f, on in (("--flag-acoustic", RECALL_FLAG_ACOUSTIC),
                                 ("--flag-all-open", RECALL_FLAG_ALL_OPEN),
                                 ("--flag-all-pii", RECALL_FLAG_ALL_PII)) if on]
```

---

## 2. Structural — this duplicates PII code already merged in senselab

### 2.1 The extraction layer already exists

`src/senselab/audio/workflows/audio_analysis/pii.py` landed in PR #519 and is live:
`analyze_audio.py:2394-2407` imports `detect_pii_in_pass`, runs it per pass, and writes
`pii.json`. It provides:

- Presidio + GLiNER, **isolated in a subprocess venv** (`pii_subprocess.py`) so the host
  process needs neither installed;
- `PiiSpan` / `PiiPassReport` dataclasses with `detector_used`, `failures`, and
  `detection_confidence`;
- an explicit "the check didn't run" state distinct from "ran and found nothing".

#542 rebuilds all of that in-process, including a bespoke `_importable()` that shells out to
a subprocess to survive C-level `SIGABRT` on GLiNER import — a workaround that exists
*because* the engines load in-process, and which the merged venv isolation makes unnecessary.

**This is where I'd push back on the dependency-weight reasoning in the PR description.** The
argument for staying self-contained is that reusing senselab machinery would drag heavy
packages into every install. `pii.py` inverts that: because its detectors live in a
subprocess venv, the host environment needs **neither** `presidio-analyzer`, `gliner`, nor
`spacy`. Consuming it makes the `pii` extra unnecessary for this tool rather than growing it
— and it resolves the CPython 3.14 caveat in the PR description outright, since the venv
pins 3.13 and the host interpreter no longer needs a spaCy wheel at all.

**Recommendation:** keep #542's adjudication layer, drop its extraction layer. The precision
work in `postprocess_entities` / `merge_pii` / `_name_hard_gate_eligible` /
`combinatorial_scan` is the part worth keeping; it should consume `list[PiiSpan]` rather than
run its own engines. That change alone resolves finding 1.5, deletes `_importable`, drops the
`pii`-extra dependency, and removes a second GLiNER weight download.

### 2.2 Required (small) change on the senselab side: `PiiSpan` needs offsets

`PiiSpan` carries `text, category, source, asr_model, pass_label, score` — no character
offsets. The subprocess worker *has* them (`pii_subprocess.py:342` does `text[r.start:r.end]`)
and throws them away.

#542's precision layer needs them in three places: `_name_hard_gate_eligible` reads the 30
characters preceding a span for context cues; `combinatorial_scan` windows by token distance;
`build_masked_preview` splices by offset. Adding them is additive and backward-compatible:

```python
@dataclass
class PiiSpan:
    ...
    # Character offsets into the transcript this span was found in. Needed by
    # downstream adjudication that reads surrounding context (name-context cues,
    # combinatorial proximity) or splices a masked preview. None when a detector
    # reports a match without offsets.
    start: int | None = None
    end: int | None = None
```

Worth doing in this PR or immediately alongside it — it's the seam everything else in §3
depends on.

### 2.3 Two *different* corroborations — don't let them collapse into one word

`pii.py` corroborates **across ASR models**: same detector, N transcripts of the same audio,
`require_cross_model_corroboration` gates `contains_pii`. It filters ASR hallucinations.

#542 corroborates **across detection engines**: N engines, one transcript, `len(engines) >= 2`
gates hard-gate eligibility. It filters detector hallucinations.

These are complementary and both worth keeping — but they're different axes, and a merged
report that calls both "corroborated" will be misread. Name them distinctly
(`cross_model_corroborated` / `cross_engine_corroborated`) and let the hard gate require
whichever the deployment actually has available: `analyze_audio` has 4 ASR models, a
single-transcript CLI run has 1.

### 2.4 Task modality is now defined in three places

- `analyze_audio.py:631` `_NONVOICED_TASK_HINTS` → `_task_expects_speech`
- `analyze_audio.py:643` `_NONLEXICAL_TASK_HINTS` → `_task_is_lexical`
  (the comments say both already mirror the annotate-notebook and ranking hint sets)
- #542 `_ACOUSTIC_TASK_PATTERNS` / `_SCRIPTED_TASK_PATTERNS` → `classify_modality`

Three copies of the same taxonomy with different membership. The concrete failure: a task
that is `acoustic` to #542 but lexical to `analyze_audio` gets "compliance skipped, requires
audio" from one and a transcript-trust claim from the other, on the same recording, in the
same run.

The mapping isn't 1:1 — #542 has three classes (acoustic / scripted / open) where
`analyze_audio` has two independent booleans — so the shared artifact should be the
**finer-grained enum**, with the booleans derived:

```python
class TaskModality(StrEnum):
    NONVOICED = "nonvoiced"  # breath / cough / respiration — no phonation at all
    ACOUSTIC  = "acoustic"   # vowel / DDK / glides — voiced, but no words
    SCRIPTED  = "scripted"   # read a known reference text
    OPEN      = "open"       # free speech against a prompt

def expects_speech(m): return m is not TaskModality.NONVOICED
def is_lexical(m):     return m in (TaskModality.SCRIPTED, TaskModality.OPEN)
```

Four classes, not three: `expects_speech` and `is_lexical` are *not* the same predicate today
— a sustained vowel is voiced but non-lexical, breath is neither — so a single `ACOUSTIC`
class would silently collapse a distinction `analyze_audio` currently makes. #542's Tier B
then maps to `NONVOICED | ACOUSTIC`. One shared module
(`workflows/audio_analysis/task_modality.py`), all three current call sites re-pointed.

---

## 3. Generalizing the interface

### 3.1 Transcript-in, not audio-in

Worth stating explicitly since it came up: this module should **not** take an `Audio`. Every
detector in it is text-only, and accepting audio would mean embedding ASR — duplicating what
`analyze_audio` does across four models with consensus. Keep it transcript-in and let the
caller decide provenance.

The one real gap is Tier B: acoustic tasks currently return
`"skipped: acoustic/non-verbal task (requires audio)"` (line 1710), so a prolonged-vowel task
is *never actually verified*. That's a genuine future feature — and the right home for it is
`analyze_audio`, which already has the acoustic features (openSMILE, phonation) to answer it.
It is not an argument for an audio-shaped interface here.

### 3.2 Task context should be caller-supplied, with the filename as fallback

`resolve_task_context` privileges `task_key_from_filename` and only falls back to `task_meta`
(lines 1647-1661). Inverting that precedence *is* the structural fix for 1.2 — the fuzzy
prefix lookup becomes the b2ai CLI's last resort rather than the path every caller routes
through. It's also what makes the module usable for a non-b2ai protocol, where filenames
carry no `task-<Name>` field at all.

Precedence becomes: caller-supplied task context → `.pt` metadata → filename lookup →
unknown. Each step records itself in `ctx["source"]`, which already exists and is already
reported.

**Don't invent the caller-supplied type here — PR #543 already defines it.** See §4.

### 3.3 A config object instead of ~12 mutated globals

`_apply_runtime_config` (line 3304) mutates twelve module-level globals that every scoring
function reads at call time. Three consequences: two callers in one process cannot use
different settings; the `HIGH_RECALL → PRECISION_MODE = False` coupling (line 3330) is an
invisible side effect (and is half of finding 1.4); and nothing is testable in parallel.

A frozen dataclass threaded through `Screener` fixes all three, and makes the
precision/recall coupling explicit as a `__post_init__` or a named constructor
(`ScreenConfig.high_recall()`) rather than a mutation.

This also matters for the project convention of surfacing thresholds: `TIER_A_WER_REVIEW`,
`TIER_A_WER_FAIL`, `GLINER_THRESHOLD`, `NAME_COMMON_WORD_ZIPF`,
`STRONG_PII_HARD_GATE_CONFIDENCE` are all already well-named and well-commented — they just
need to live on an object that can be serialized into the report, so a run's thresholds are
recoverable from its artifact.

### 3.4 Open question 2: where this should live

Direct answer — **a `senselab` module with a thin CLI**, and not for style reasons.

`pyproject.toml:176` excludes `scripts/` from ruff (and the PR checklist notes mypy too). As
written, this adds **3,429 lines of unlinted, untyped, CI-unchecked logic**. That's also why
open question 1 feels awkward: the code isn't anywhere pytest is pointed.

Suggested split, mirroring how `pii.py` and `analyze_audio.py` already divide:

```
src/senselab/audio/workflows/audio_analysis/
    task_modality.py     # §2.4 — the one shared taxonomy
    compliance.py        # ScreenConfig, Screener, Tier A/B/C — the library
    pii.py               # extended: PiiSpan gains offsets (§2.2)
scripts/
    pii_compliance_pipeline.py   # thin CLI + .pt adapter over the above
```

That gets ruff + mypy + `src/tests/` for free, makes the deterministic half of `--selftest` a
normal test module, and turns `SystemExit` (lines 2500, 2510, 2512, 3184) into a raised
exception the CLI catches — a library that kills the host process on a malformed config JSON
can't be embedded anywhere.

Worth being clear that this doesn't cost the property the PR is protecting: the CLI stays
runnable from an IDE with no arguments, and per §2.1 the *installed* dependency footprint
goes down, not up.

`task_reference.json` (797 tasks) is corpus data, not code. Given it will drift from any real
study over time (see 1.2, and §4.2 for the alternative), worth deciding deliberately whether
it ships inside the package or is loaded from a configured path.

---

## 4. PR #543's `AudioPlus` is the substrate this should be built on

PR #543 (speaker-profile) also lands `src/senselab/audio/data_structures/audio_plus.py` and
`src/senselab/audio/metadata/b2ai.py`. It solves, generically, most of what §3.2 asks for —
worth aligning before #542 hardens a parallel answer.

### 4.1 `TaskInfo` is the `TaskSpec` — don't define a second one

```python
class TaskInfo(BaseModel):
    name: Optional[str] = None       # the acquisition task
    content: Optional[str] = None    # the prompt text
    metadata: dict[str, Any] = Field(default_factory=dict)
```

That's exactly the caller-supplied task context §3.2 asks for — already on `AudioPlus.task`,
already pydantic, matching the `Audio` convention. `compliance.py` should consume `TaskInfo`
and extend it where it falls short (§4.3) rather than defining a parallel type.

### 4.2 `MetadataProvider` is the generalization answer — and it dissolves finding 1.2

The pluggable-provider boundary is exactly the "beyond b2ai tasks in the future" question:

```python
@runtime_checkable
class MetadataProvider(Protocol):
    def lookup(self, ref: str) -> AudioPlusMetadata: ...
```

#542's filename regex + `task_reference.json` becomes **one provider implementation** rather
than the pipeline's only way to identify a task. A different study writes its own `lookup()`;
`NullMetadataProvider` keeps the no-metadata path working; tests inject a stub instead of
shipping a 797-entry table.

More important for correctness: `B2AIMetadataProvider.lookup` reads the recording's own
`_recording-metadata.json` sidecar and takes `task_name` and **`prompts`** from it
(`b2ai.py:170-174`). That is the *authoritative* reference text for that exact recording. No
normalization, no prefix matching, no sibling-variant ambiguity — **finding 1.2's entire
failure class does not exist on this path.** `task_reference.json` demotes from "the task
table" to "a fallback provider for corpora without sidecars", which is the right status for
a snapshot that will drift.

This is the strongest argument for doing §3.4 (move the core under `src/`): the provider
protocol lives there, and a `scripts/`-only module can't participate in it cleanly.

### 4.3 Two gaps in `TaskInfo` that compliance needs — both small and worth raising on #543

**(a) `content` conflates the Tier A reference with the Tier C instruction.** "Read this
sentence: *The rainbow is a division of white light*" and "Describe what you see in this
picture" are both `content`, but Tier A word-error-rates against the first and the LLM judge
scores against the second. Compliance can't tell them apart without the modality (§2.4), so
either `TaskInfo` gains a modality field or `compliance.py` derives it from `name` via the
shared taxonomy — the latter is probably cleaner, keeping `TaskInfo` a pure metadata carrier.

**(b) `_join_prompts` flattens the prompts list and drops the structure.** The sidecar's
`prompts` is a list; `_join_prompts` space-joins it into one string and `TaskInfo.metadata`
is left empty, so the list is unrecoverable. #542's `_tier_a` scores against the
**best-matching** of several `reference_texts` (lowest WER, ties broken by coverage) — which
is the right behaviour for a Harvard-sentence-list task where the participant reads one of N.
Against a joined string, a participant who read one sentence correctly scores a WER near 1.

Minimal fix on #543's side, preserving the current `content` contract:

```python
        task = TaskInfo(
            name=sidecar.get("task_name") or task_entity,
            content=self._join_prompts(sidecar.get("prompts")),
            # Keep the unjoined list: consumers that score against the BEST-matching
            # prompt (per-sentence read tasks) cannot recover it from the joined string.
            metadata={"prompts": [str(p) for p in (sidecar.get("prompts") or []) if p]},
        )
```

### 4.4 `AudioPlus` unlocks two capabilities #542 currently can't have

**Cross-recording PII corroboration — a third axis.** `related_audio_refs` gives the same
speaker's other recordings. A name detected in one recording and again in another from the
same speaker is far more likely real than an ASR hallucination, which is uncorrelated across
sessions. That stacks cleanly on the two axes in §2.3 (cross-ASR-model, cross-engine) and is
the axis best suited to the NAME category, which is exactly where #542's precision guards
work hardest. Refs are unloaded by design, and this needs only the *transcripts*, not the
waveforms.

**Age-grounded demographic scanning.** `SpeakerInfo.age` is a float from the demographics
table. #542's `age_scan` has an `over_years` parameter but no ground truth, so a
self-disclosed age is flagged on pattern alone. With the speaker's actual age available, the
check becomes "did the participant disclose an age *consistent with the record*" — which both
raises precision and turns a pure PII flag into a usable data-quality signal.

**One caution.** `SpeakerInfo.gsd` (gold-standard diagnosis) and `metadata["gsd_details"]` are
themselves clinical PHI. If compliance results carry an `AudioPlus`, GSD can leak into a
report whose whole purpose is to be circulated to reviewers. `report_to_dict`-style
serialization for compliance should **allow-list** the fields it copies out of `AudioPlus`
(`recording_id`, `task.name`, `speaker.speaker_id`) rather than `model_dump()`-ing the bundle.
Worth an explicit comment at the serialization boundary.

### 4.5 It's also the clean entry point for running both tools together

`build_audio_plus` resolves metadata **before** loading the waveform (deliberately — the
docstring says so), which means a compliance-only pass needs no audio decode at all. One
`AudioPlus` per recording is a natural shared input: `analyze_audio` takes the waveform and
produces transcripts; the compliance screen takes `task` + those transcripts. That's the
"run in conjunction" shape without a `.pt` round-trip, and without either tool having to know
how the other resolves a recording.

---

## 5. The `analyze_audio` seam, concretely

### 5.1 There is no first-class consensus transcript to read yet

`analyze_audio` keeps per-model transcripts in `model_votes[m]["text"]`; plurality consensus
is computed inside `_utterance_text_payload` (`labelstudio.py:59-72`) purely to render a
Label Studio card. Any compliance pass wanting "the" transcript needs that lifted into a
real artifact. Worth doing independently — it's useful to more than this feature.

### 5.2 Tier A thresholds are ASR-dependent, and that needs recording

This is the one I'd most want in the code as a comment. Tier A WER measures **reading error +
ASR error jointly**. `TIER_A_WER_REVIEW` / `TIER_A_WER_FAIL` were tuned against whatever ASR
produced the `.pt` transcripts; feeding a different transcript source shifts every score and
therefore every threshold. On this corpus, swapping senselab's chunked Whisper pipeline for the
openai-whisper CLI moved WER 0.224 → 0.187 on identical audio against human-verbatim ground
truth — a shift larger than the gap between the review and fail bands.

So: `ScreenConfig` should carry a transcript-provenance field, the report should record it,
and the thresholds need one calibration pass per source rather than being assumed portable.
A comment at the threshold definitions naming the source they were calibrated against, and
saying to re-check when it changes, is the minimum.

### 5.3 Run it as a second pass, not an inline stage

Even after §2/§3/§4, I'd keep this out of `analyze_audio`'s main loop. The PII engines are a
heavy, separately-versioned dependency set living in their own subprocess venv, and
`analyze_audio` is typically run on GPU nodes where you don't want GLiNER weights competing
for the allocation. Two processes over one shared transcript artifact get the consensus benefit
without the coupling — and the library refactor is exactly what turns that second process
into a clean call instead of a `.pt` round-trip.

### 5.4 Screen per-bucket, not per-file

`analyze_audio` already has diarized buckets with per-model transcripts. Screening at that
granularity is strictly more useful than a file-level verdict: it localizes *which utterance*
carries the PII, and it lets the compliance check distinguish "the participant didn't do the
task" from "an interviewer's speech dominates the transcript". `PiiSpan` already carries
`pass_label`; `TaskInfo` should be resolvable per bucket for the same reason.

---

## 6. Not blocking, worth noting

- `pyproject.toml` (`gliner>=0.2.28`, `wordfreq>=3.1.1` into the existing `pii` extra) and the
  `.pre-commit-config.yaml` codespell exclusion for `task_reference.json` both look right —
  though see 1.3 on whether `wordfreq` should be optional at all, and note `gliner` is now
  declared in two places once §2.1 lands.
- Verified-correct areas I probed and found nothing wrong in: `_align_counts` Levenshtein
  breakdown, `tier_a_score` / `llm_judgement_score` banding and monotonicity, `_gliner_chunks`
  offset mapping, `merge_pii`'s greedy clustering (sorting by `start` does yield true
  connected components, so `build_masked_preview` cannot corrupt spans), `_age_word_to_int`,
  and `write_csv` column arity on error rows.
- The comment density in this file is genuinely good — the *why* is written down almost
  everywhere. Several findings above (1.1, 1.4, 1.6, 1.8, 1.12) are cases where the comment is
  right and the code drifted from it, which is a much better failure mode than the reverse.

---

Summary: §1.1 and §1.2 are the ones I'd want fixed before this is pointed at a real corpus —
both corrupt the verdict the tool exists to produce, and §1.2 does it in the direction that
costs reviewer trust fastest. The rest of §1 is mostly defaults and docs disagreeing with
code. §2–§4 are a follow-up, not a blocker, but worth deciding on now while the interfaces
are still soft.

🤖 Review conducted with [Claude Code](https://claude.com/claude-code); every §1 finding was
reproduced by execution before being reported.

varun-thvar:
Thanks — this was an unusually useful review, and the fact that every §1 finding came with a reproduction made them all quick to confirm and quick to verify fixed.

All twelve §1 findings are addressed in `812246a2`. §2–§4 deferred as follow-up, per your sequencing suggestion. Details below, including two places I deviated from the suggested patch and why.

## §1 — all fixed

| # | Fix |
|---|---|
| **1.1** | `check_quality` gained `keep_hard_fail`. The *score* is still suppressed in both cases; the *hard fail* now survives when suppression came from a Tier A/C score, and is dropped only for acoustic. `_compliance_check_score` ORs the hard fail unconditionally rather than inside the `score is not None` branch. |
| **1.2** | `_norm_task` collapses separator runs to `-` instead of deleting them; both prefix branches of `_resolve_reference_keys` require the match to end on a segment boundary. |
| **1.3** | `_zipf` returns `None`, not `0.0`. Both call sites take their precision-safe branch on `None`. Added `_wordfreq_available()` so the soft rare-role qualifier — which *is* the frequency lookup — doesn't fire at all without it. |
| **1.4** | Format validation and the cross-engine corroboration requirement are both now independent of `PRECISION_MODE`. `PRECISION_MODE` governs only the confidence-threshold guard. |
| **1.5** | `weights_only=True`, with `TRUST_INPUT_PICKLES` as the explicit opt-out; the load error names it so a legitimate failure is self-explaining. |
| **1.6** | `ENABLE_LLM = False`. |
| **1.7** | `INPUT_FOLDER = ""` — the `args.folder is None` branch is reachable again. |
| **1.8** | `MAX_FILES = None`, plus `summary.truncated` / `summary.files_available` in the report and a warning line in the console summary. |
| **1.9** | Bare output filenames resolve beside the **input data**, not the CWD; `masked_preview` is uniformly the masked form; all three default output names added to `.gitignore`. |
| **1.10** | Dropped `summary["fail"]`; added `summary["hard_gate"]`. Summary now prints `review: N (of which hard-gate: M)`. |
| **1.11** | `--selftest` is hermetic — engines forced off, gazetteer stubbed, no `nltk.download`, no weight pull. `--selftest-full` keeps the engine-dependent path as an opt-in. |
| **1.12** | `--flag-all-pii` added to the stray-flag interlock. |

Added **Part K** to the self-test: 11 regressions covering 1.1, 1.2, 1.3, 1.4, 1.10, 1.12. Now 138 checks.

## Two deviations from the suggested patch

**§1.1 — gated on the suppression *reason*, not applied unconditionally.** Your patch keeps `hard_fail` in every suppress case. Applied literally that also hard-fails acoustic tasks, where an empty transcript is the *compliant* outcome — Tier B would start flagging every breath and prolonged-vowel recording. So `keep_hard_fail=False` is passed only for acoustic, and the NOTES block now states that explicitly, since you were right that code and doc disagreed. Both directions are pinned by regressions:

```
[PASS] K1.1: an LLM 'completed' verdict cannot rescue an EMPTY transcript
[PASS] K1.1: ...but an ACOUSTIC empty transcript still never hard-fails
```

If you'd rather gate on the emptiness subtype instead, that's a small follow-up — but it needs a decision about whether `empty` and `non_speech_only` should differ for acoustic, which I didn't want to make unilaterally.

**§1.9 — `masked_preview` is withheld, not just uniformly masked.** Uniform masking alone doesn't close it: when no PII is detected there is nothing to mask, so the "masked" preview *is* the raw transcript. It is now `None` for clean files unless `INCLUDE_TRANSCRIPTS = True`. Detected spans and the genuinely-masked preview are unaffected.

## §1.3 — took the fail-safe route rather than hard-requiring wordfreq

You preferred hard-requiring it; I went with the unambiguous sentinel plus precision-safe call sites, and skipped the one genuinely frequency-dependent assertion when wordfreq is absent. Reasoning: the concrete problem was the self-test going red on a clean checkout, and this fixes that without making a `pii`-extra package mandatory for a `scripts/` tool. Verified both ways:

```
with wordfreq:     138 checks, SELF-TEST PASSED
without wordfreq:  137 checks + 1 skip, SELF-TEST PASSED (exit 0)
```

Happy to switch to a hard requirement if you still prefer it — one line, and it removes the "two silently different detection behaviours in the field" concern you raised, which I agree is the stronger argument long-term.

## §2–§4 — agreed, and deferred

No disagreement on the substance. §2.1 in particular reframes the dependency argument in the PR description, and I'd rather not rewrite the engine layer twice — doing §1 here and the restructuring as a follow-up means the extraction layer that §2.1 deletes doesn't get reworked first.

Specifically noted for the follow-up:

- **§2.1** — consuming `list[PiiSpan]` from the merged `pii.py` deletes `_importable`, drops the `pii` extra for this tool, resolves §1.5 structurally, and sidesteps the CPython 3.14 spaCy-wheel gap. That's a strictly better answer than what's here.
- **§2.2** — `PiiSpan` needs `start`/`end`; agreed it's the seam everything else depends on.
- **§2.3** — cross-model vs cross-engine corroboration being different axes is a good catch. The current code calls the latter `cross_validated`; renaming to `cross_engine_corroborated` before anything downstream keys off it is cheap.
- **§2.4** — three copies of the task taxonomy is the one I'd want to fix soonest, since the failure is two tools disagreeing about the same recording in the same run. Your four-class enum looks right; three would collapse the voiced/non-lexical distinction `analyze_audio` currently makes.
- **§4.2** — reading the reference prompt from each recording's own sidecar removes §1.2's entire failure class rather than guarding it. The boundary fix here is the right patch for the table-lookup path, but that path should become the fallback.
- **§5.2** — the ASR-dependence of the Tier A thresholds is the one I'd most want written down; a comment at the threshold definitions naming the source they were calibrated against is going in with the follow-up.

Also fixed the pre-commit failures from CI in `b08f4f3a`: JSON formatting and `unparseable`→`unparsable`. codespell is excluded for `task_reference.json` specifically — it flags the Harvard corpus's own "busses" and "wit" (cleverness, not a typo for "with"), and accepting either would put the reference text out of sync with what participants were actually given, giving every recording of those two tasks a permanent WER mismatch.

🤖 Response drafted with Claude Code; every fix above was verified by execution.

wilke0818:
Reviewed `812246a2`. All twelve §1 findings are genuinely fixed and `--selftest` is green (138 with `wordfreq`, 137 + 1 skip without — confirmed both ways). Your two deviations are both better than what I suggested: gating §1.1 on the suppression *reason* avoids the Tier B false-positive flood my patch would have caused, and you're right that uniform masking doesn't close §1.9 — with no detected PII the "masked" preview *is* the raw transcript, so withholding it is the actual fix.

**One regression, and it's mine.**

### `scripts/pii_compliance_pipeline.py:1576` — the `_norm_task` change breaks `classify_modality`

`_ACOUSTIC_TASK_PATTERNS` / `_SCRIPTED_TASK_PATTERNS` are separator-free substrings written for the old delete-everything normalization — `"prolongedvowel"`, `"hightolow"`, `"longsounds"`. With separators collapsed, `"prolongedvowel" in "prolonged-vowel"` is False. Nine patterns are now unmatchable, and five keys in the shipped table flip `acoustic → open`:

```
prolonged-Vowel, high-to-low, long-sounds, noisy-sounds, silly-sounds
```

An `open` task gets no Tier C score without the LLM, so transcript quality applies and a prolonged vowel transcribed as `"ahhh"` — or empty — now hard-fails. Verified end-to-end on the same input at both commits:

```
b08f4f3a: task-Prolonged-Vowel  PASS    (modality=acoustic)
812246a2: task-Prolonged-Vowel  REVIEW  (modality=open)
812246a2: task-Prolonged-Vowel  REVIEW → compliance:fail   # empty transcript
```

`Prolonged-Vowel` is high-volume in this corpus, so a real run flags a large fraction of compliant recordings. My review text claimed the fix "keeps the intended matches working" — I verified that only for `Rainbow-Passage → rainbow`, the reference-lookup path, and never checked the modality consumer. Sorry for the bad patch.

`K1.1: ...an ACOUSTIC empty transcript still never hard-fails` passes for the wrong reason — those tasks aren't acoustic anymore. Part C misses it because `Maximum-Phonation-Time` survives via the shorter `phonation` pattern.

Fix: the two consumers want different normalizations. Keep the collapsed form for reference-key lookup; match modality patterns against a separator-stripped copy (or hyphenate the patterns). Worth a Part K case over the acoustic keys actually present in `task_reference.json` — that's what would have caught it.

### Two places the restored hard fail still doesn't survive

**`:1552` `_recall_compliance_override`** builds a fresh dict with `hard_fail=False` hardcoded and returns it *in place of* `quality`, so `--high-recall` erases what `keep_hard_fail=True` just preserved. Open task + Tier C score + empty transcript + `HIGH_RECALL` → `compliance_fail` goes True → False. The comment at `:2270` says the override "can only add a flag"; it replaces the dict and can remove one. Carrying `hard_fail=quality.get("hard_fail", False)` through both branches fixes it.

**`:2016`** `if not scores: return None, None, False, False` discards the hard fail that §1.1 deliberately moved outside the score branch. Latent — the only caller reaching it is the acoustic path where `keep_hard_fail=False` already zeroed it — but it contradicts the new invariant. Return `hard_fail`.

### Minor

- **`:3653`** `_out()`: `Path("./x.json").parent` is also `Path(".")`, so no relative path can target the CWD — only an absolute one escapes. And since the writes happen *after* the full scan, a read-only corpus mount crashes the run at the end with all work lost. The `--output-dir` half of the §1.9 suggestion is the escape hatch that's missing.
- **`:3599`** the modes banner builds `outs` before `_out()` applies, so it prints paths the run won't use — in the banner that exists so an IDE run can see where output went.
- **`:3618`** the gazetteer stub lives in `main()`, so hermeticity belongs to the CLI, not the function. A pytest wrapper that imports and calls `run_selftest(args)` — the natural way — still hits `nltk.download("names")`. Moving the stub into the non-`full` branch makes it a property of the test.
- **`:3383`** K1.12 asserts `"--flag-all-pii" in inspect.getsource(...)`, which passes on a comment and survives a refactor that drops the behaviour. Parse the flag alone and assert `HIGH_RECALL` became True.
- **`:142`** `COMPLIANCE_ONLY`'s trailing comment ended up on `INCLUDE_TRANSCRIPTS`, which now documents a different feature — in the block users are told to read first.
- **`:138`** `TRUST_INPUT_PICKLES` / `INCLUDE_TRANSCRIPTS` have no CLI flags, unlike every other posture knob. §1.5 specifically suggested `--trust-input-pickles`; as shipped the load error points at a constant the user must edit in source.

On §1.3 — your call is fine for the stated problem. I'd still take the hard requirement eventually, for the "two silently different detection behaviours in the field" reason, but that's not a blocker.

🤖 Follow-up review with [Claude Code](https://claude.com/claude-code); regressions above reproduced by execution at both commits.

satra:
Closing: the PII half of this PR has landed in #550, and the rest was scoped out deliberately rather than overlooked.

**What landed**, in `src/senselab/text/tasks/pii_detection/`:

- The full rule cascade, function for function, in `rules.py` — `rigidity_tier`, `postprocess_entities`, the gazetteer/regex/NER/honorific/profession/rareword/rare-role/demographic/age/combinatorial scans, `merge_pii`, `build_masked_preview`. The file header credits this PR as its source.
- Both review fixes from this thread, with the reasoning kept: `_zipf` returns `None` rather than `0.0` (the old sentinel inverted two guards, since "maximally rare" reads as evidence *for* a hit), and `precision_mode` governs only the soft-tier threshold while structured-identifier format validation always runs.
- The rules cascade as a third detector alongside Presidio and GLiNER, with cross-detector corroboration.
- The optional local-LLM detector — `local_llm.py`, added in ad4fffa2. Off by default, loopback-only as a checked invariant, and an unreachable server is a recorded failure rather than a clean pass.
- Real pytest coverage: `pii_rules_test.py`, `pii_detection_test.py`, `pii_detection_api_test.py`, `pii_detection_subprocess_test.py`, `pii_llm_test.py`, `pii_adapter_test.py`.

**What was not carried over, by decision:** task-compliance verification, Tier A/B/C modality routing, the Tier C LLM *judge* (`llm_judgement_score`, which scores task completion — distinct from the detector above), the `.pt`-folder batch driver, and `scripts/task_reference.json`. The 797 Bridge2AI task definitions are available elsewhere and shouldn't live in this repo.

Two things worth preserving from this thread rather than losing with it:

1. @wilke0818's second review is still unactioned — the `_norm_task` separator-collapsing regression that flips five acoustic keys (including the high-volume `prolonged-Vowel`) to `open`, so compliant recordings hard-fail, plus the interaction with `_recall_compliance_override`. All compliance-side, so it travels with that half wherever it lands next.
2. GLiNER windowing. `_gliner_chunks`/`gliner_scan` here window long text with overlap and correct entity offsets; #550 calls `predict_entities` on the whole string, so a long transcript is silently truncated at the model's token limit. That's a real gap in #550 and I'll track it separately — it isn't a reason to keep this open, but it is a fix this PR had and #550 doesn't.

Thanks — the cascade and its precision guards are the substance of the PII work now on `alpha`'s path. Reopen if the compliance half should come back as its own PR.

PR #540, opened 2026-07-28
fix(asr): CrisperWhisper ct2 venv needs torch+transformers for HF→CT2 conversion

wilke0818:
## Problem
On Linux x86_64, the CrisperWhisper backend builds its subprocess venv from `crisperwhisper[ct2]` only. CT2 *inference* is torch-free, but the first-run HF→CT2 **conversion** (`model._init_v2` → `converter.ensure_ct2_model` → `ctranslate2.converters.transformers._load`) needs **torch and transformers**: that module does `try: import huggingface_hub, torch, transformers` and, when they're absent, leaves the names unbound; `_load()` then calls `torch.no_grad()` and the transformers loader → `NameError: name 'torch' / 'transformers' is not defined`.

**Impact:** the venv "builds" but every clip's CrisperWhisper ASR fails in ~0.7 s with that `NameError`, silently dropping the `whisper` model from multi-ASR runs (found on a CUDA-13.1 GPU batch; `canary-qwen` + `Qwen3-ASR` still succeed). macOS is unaffected — it uses the transformers backend (torch already declared) and never invokes the CT2 converter.

## Fix
Add the `[transformers]` extra (`== [all]` alongside ct2: transformers + torch + accelerate) to the Linux-x86 ct2 requirements, and pin `torch`/`torchaudio` so `ensure_venv` routes them through the CUDA-aware index. The conversion result is cached, so torch/transformers are installed but never loaded at CT2 inference time.

```python
# before
_CRISPER_REQUIREMENTS = ["crisperwhisper[ct2]==2.0.1"]
# after
_CRISPER_REQUIREMENTS = ["crisperwhisper[ct2,transformers]==2.0.1", "torch>=2.4", "torchaudio>=2.4"]
```

## Why CI didn't catch it
`crisperwhisper_test.py`:
- the assembly test mocks `ensure_venv` / `venv_python` / `subprocess` → the ct2 venv is never built;
- `test_crisperwhisper_transcribes_when_venv_present` is `@skipif(not CRISPER_VENV.exists())` → skipped in CI, and where a dev *has* the venv (macOS) it runs the torch-having transformers backend.

So no test builds the `[ct2]` venv on Linux-x86 and runs a conversion — the only path that hits this. A GPU-gated regression test mirroring `brouhaha_venv_gpu_test.py` (build ct2 venv + transcribe) would cover it; happy to add as a follow-up.

## Validation
[...]

PR #550, opened 2026-08-09
feat: model integrations — diarization backends, capabilities, DriftSE, unasdiff, PII, Qwen3-TTS, plus shared-cache locking and HF commit pinning

satra:
## Description

One PR carrying the model-integration work and the two infrastructure changes it forced.

| Workstream | State |
|---|---|
| **A** — four diarization backends (from #537) | **complete**, GPU-verified |
| **Capabilities** — declared, harmonised output contract | **complete** |
| **D** — DriftSE one-step speech enhancement | **complete** — one gap, see below |
| **C** — unasdiff source separation | **scaffolding + class map**; worker/API/chunking to come |
| **B** — PII detection as a standalone text task | **relocated + entry point in progress** |
| **Shared-cache locking** (emerged) | **complete** |
| **HF commit-SHA pinning** (emerged) | **complete** |

Two of these were not planned. They came out of running the work on a cluster, and both fix defects that were already there.

### A — four diarization backends, task layer only

**VibeVoice-ASR-HF** (in-process), **USC-SAIL child-adult** (subprocess venv, CUDA-only, role labels), **MOSS-Transcribe-Diarize** and **DiariZen** (subprocess venvs; DiariZen's weights are **CC BY-NC 4.0, non-commercial only**).

Reachable through `diarize_audios(model=...)` and deliberately **not** through `audio_analysis`. The five workflow files from #537 are excluded on purpose: the guards keeping child-adult's `CHILD`/`ADULT`/`OVERLAP` role labels out of embedding clustering live there, and without them a role-label backend would build a `CHILD` centroid blending two different children.

Also raises the core `transformers` floor `>=5.0` → `>=5.3`, inert in a current environment (already resolves to 5.5.4).

### Capabilities — a declared output contract

Six backends share a return type and disagree about almost everything else. Measured on an H100: `text` is populated by two of six; `speaker` denotes an identity for five and a **role** for child-adult; DiariZen's VBx clustering numbers per audio, so one run gave `['1','2']` for one file and `['0','0','1','0']` for another.

A frozen `DiarizationCapabilities` record declares it — `populates_text`, `speaker_label_kind`, `labels_stable_across_files`, `max_speakers`, `honors_speaker_hints`. **`max_speakers: —` means unmeasured, not unlimited.**

### Shared-cache locking

A group sharing one `HF_HOME` shares the weights but not the coordination. Both file locks assumed a single user, in ways that failed silently: lock files created `0644` meant a second user's heartbeat write failed into an `except: pass`, so a *third* user read a stale heartbeat and broke a live lock.

One `SharedFileLock` now serves both call sites — group-writable `0664`/`2775`, a JSON holder payload (user, host, pid, taken_at), staleness takeover, and a `TimeoutError` that names the holder so a human can decide whether to kill it. `_HeartbeatLock` and the decorative `_FileLockWithHeartbeat` (which wrote a heartbeat nothing ever read) are both gone.

Three defects found on the way, each caught by review or measurement:

- **Breaking a stale lock on the timeout path displaces a *live* holder.** A crashed process's flock is kernel-released, so reaching a timeout *proves* someone is alive. Unlinking there hands the new claimant a fresh inode while the original holds the orphan, and both proceed.
- **`with_suffix` collided two locks into one.** It replaces everything after the last dot, so `org--model--v1.5--main` and `org--model--v1.6--main` both derived `org--model--v1.lock`. Derivation now appends, which is injective.
- **`ensure_venv` chmod'd the venv *after* writing `.senselab-installed`.** A kill in between left a venv advertising readiness with permissions half-applied, and since the chmod pass only runs on the fresh-build path, no later call ever repaired it.

### HF commit-SHA pinning

Every production call site loaded at `revision="main"`, no output recorded which commit ran, and **cache keys contained only the model id** — so an upstream push loaded new weights under an unchanged key and served a stale result as current. That last one is a correctness bug, not a provenance gap.

The rule: **resolve the ref to a commit, then load again passing that commit.** Resolving alone binds nothing — `snapshot_download(revision="main")` writes `refs/main` and the caller stays ref-addressed. The second call is free: a full SHA triggers `huggingface_hub`'s commit-hash shortcut, returning cached files with no network at all.

- `HFModel` gains `commit_sha` beside `revision` — what resolved vs what was asked for, so provenance can tell a deliberate pin from a tracked ref that happened to land there.
- `cache_key` and `align_cache_key` take a **required** keyword-only SHA. `CACHE_SCHEMA_VERSION` 22 → 23.
- **`SENSELAB_RUN_ID`** names a run; each `(repo_id, ref)` resolves once into `$SENSELAB_CACHE/runs/<run_id>/resolutions.json` and every participant binds to that answer. A sweep across nodes and hours can otherwise straddle an upstream push — each task recording its own commit correctly while the run as a whole is quietly inhomogeneous.
- An AST guard fails when a new subprocess worker payload carries an unreviewed revision-shaped key.

Two defects worth naming, both found after the feature "worked":

- **The manifest was consulted by the cache key and ignored by every loader.** `resolve_model` passed the ref to `ensure_hf_model`, so loads resolved against the local `refs/main` while artifacts recorded the manifest's answer. On one warm node they agree, which is why the suite was green.
- **Fixing that broke bare loaders.** `snapshot_download(revision=<sha>)` writes **no** `refs/` entry, and NeMo's and DiariZen's loaders accept no revision — so they failed offline. Measured, not reasoned: a bare `from_pretrained` under `HF_HUB_OFFLINE=1` after SHA-only staging fails outright. `resolve_model` now points `refs/<ref>` at the pinned commit, which also upgrades those backends from staged-only to genuinely pinned.

### D — DriftSE

[DriftSE](https://github.com/LiangXu123/DriftSE) (Interspeech 2026, oral) reaches the clean-speech distribution in a **single** network evaluation, against 30 for SGMSE+. Its SSL encoders are a *training* signal, so inference needs none — which is why the venv omits upstream's training and metrics dependencies.

Reachable only by explicit model id — not in a default list, not in `audio_analysis`.

**Licence: unresolved upstream.** No LICENSE file, no README statement; a request was opened 2026-08-08 and is unanswered (LiangXu123/DriftSE#2). The checkpoint mirror stays private until it is.

**Known gap:** upstream JIT-compiles a CUDA `upfirdn2d` extension with a pure-PyTorch fallback whose *selection* is unconfirmed. The end-to-end test is written and skip-gated. It could not be verified because the mirror is private pending the licence answer — so this is blocked on licensing, not effort.

### C — unasdiff (in progress)

Scaffolding, pinned upstream commit, venv constants, and the FSD-41 class map generated from upstream's `label.json`. Worker, the three separation modes, long-form chunking with cross-window permutation alignment, and the registry are still to come.

### B — PII detection (in progress)

Moved out of the audio-analysis workflow to `src/senselab/text/tasks/pii_detection/`, as a standalone task rather than a script — detect PII given text or a `ScriptLine`, **not** wired into `analyze_audio`. The LLM judge stays optional and off by default. The `detect_pii` entry point is in progress.

## How Has This Been Tested?

- **2319 passed, 33 skipped** across `src/tests/audio/`, `src/tests/utils/`, `src/tests/text/`
- `pre-commit run --all-files` clean; `ruff`, `mypy --extra-checks` clean on 245 source files
- Previous CI on this branch: `pre-commit`, `cpu-tests (3.12)`, `build-and-deploy` all **pass**
- **On an H100 (MIT ORCD)**: all four diarization backends on three real recordings; `scripts/run_diarization_backends.py` makes it reproducible and emits the field-occupancy matrix the capabilities design rests on

Defects that only real audio or a real cluster surfaced:

- **VibeVoice turned a missing `Speaker` key into the literal string `"None"`** — labels `['None', '0']`. A phantom label is indistinguishable from a real one to any consumer grouping by speaker.
- **child-adult's short-clip guard fired** on a 4.92 s file. Without it that call returns `[]` — for a backend answering "was an adult present?", a fabricated *no*.
- **A declared capability was wrong**: `max_speakers=2` was commented "it can only ever emit CHILD and ADULT", but the worker also emits `OVERLAP`.

## Notes for reviewers

- **`transformers>=5.3` is the widest-blast-radius change.**
- **Merging this recomputes every `analyze_audio_cache/` entry once.** Schema 23 keys on the commit; pre-23 entries cannot be attributed to one, so none is reused. Intended, and not a regression.
- **Tests must not construct an `HFModel` unmocked** — that triggers a real `snapshot_download`. An earlier revision pulled 20 GB and would have on every cold CI run. Every such test monkeypatches `check_hf_repo_exists` **and** `resolve_revision` independently.
- **A pre-existing test-pollution bug is fixed here**: `uvloop_test.py` popped `"senselab"` from `sys.modules` without restoring it, breaking every later string-target `monkeypatch.setattr` in the session.
- `HF_HUB_OFFLINE=1` with an *empty* cache is not a CI-like environment — `is_hf_model_cached` returns `True` unconditionally under that flag, so failures there are an artifact of an impossible setup, not regressions.
- History is not squashed — each commit is one reviewed step.

## Known follow-ups (recorded in the specs)

1. DriftSE's `upfirdn2d` fallback selection — blocked on the upstream licence answer.
2. Four subprocess backends are pinned via the staged commit plus the `refs/<ref>` pointer, because their upstream loaders accept no revision parameter. DiariZen is genuinely unpinnable at the loader.
3. `max_speakers` is unmeasured for four backends; a NeMo synthetic-speaker probe is specced to replace model-card claims with measurements.
4. child-adult's short-clip guard sits after `ensure_venv`, so a CUDA host handed a too-short clip pays a 438 MB venv build before rejection.
5. `api.py` dispatches on `"nvidia/diar"` while `model_for_task` matches `"nvidia/diar_sortformer"` — a routing-contract decision, deliberately filed rather than patched inside a declaration change.

## Types of changes
- [x] Bug fix (cache-key staleness, three lock defects, two pinning defects)

## Checklist:

Co-authored with Evan Ng, whose #537 the diarization backends are cherry-picked from.

# Version

Published prerelease version: `1.3.1-alpha.41`

<details>
  <summary>Changelog</summary>

  #### 🐛 Bug Fix
  
  - feat: model integrations — diarization backends, capabilities, DriftSE, unasdiff, PII, Qwen3-TTS, plus shared-cache locking and HF commit pinning [#550](https://github.com/sensein/senselab/pull/550) ([@satra](https://github.com/satra))
  - Per-speaker identity and scene: one grid, one fold, and the decisions in the config [#547](https://github.com/sensein/senselab/pull/547) ([@satra](https://github.com/satra))
  - fix(asr): CrisperWhisper ct2 venv needs torch+transformers for HF→CT2 conversion [#540](https://github.com/sensein/senselab/pull/540) ([@wilke0818](https://github.com/wilke0818))
  - hf-cache: stop HF Hub 429s under parallel batch (resolve-once + SHA-pin) [#539](https://github.com/sensein/senselab/pull/539) ([@wilke0818](https://github.com/wilke0818))
  - fix(adaptive): guard None history in convergence improvement delta [#541](https://github.com/sensein/senselab/pull/541) ([@wilke0818](https://github.com/wilke0818))
  - refactor(cache): cached-inference layer out of analyze_audio.py (T051, parts 1-2) [#538](https://github.com/sensein/senselab/pull/538) ([@satra](https://github.com/satra))
  - Scene-aware presence/quality axis + ASR transcription overhaul (CrisperWhisper 2.0, Qwen) [#536](https://github.com/sensein/senselab/pull/536) ([@satra](https://github.com/satra) [@claude](https://github.com/claude))
  - feat(canary): split long audio at pauses to avoid the ~40s truncation [#534](https://github.com/sensein/senselab/pull/534) ([@wilke0818](https://github.com/wilke0818))
  - ScriptLine: accept empty text as "model produced no transcript" signal [#520](https://github.com/sensein/senselab/pull/520) ([@wilke0818](https://github.com/wilke0818))
  - Feat/pii subprocess venv [#519](https://github.com/sensein/senselab/pull/519) ([@wilke0818](https://github.com/wilke0818))
  
  #### ⚠️ Pushed to `alpha`
  
  - feat: restructure outputs as L1/L2/final, add the run summary and L1 evidence plot ([@satra](https://github.com/satra))
  - refactor: separate L1 from L2, and name each statistic for its own mathematics ([@satra](https://github.com/satra))
  - fix(mask): uncertainty is distance from the decision, not spread between questions ([@satra](https://github.com/satra))
  - fix(fuse): stop shadowing a dict-of-lists with a dict-of-floats ([@satra](https://github.com/satra))
  - feat(invariance): opt-in probes for perturbations a correct model must survive ([@satra](https://github.com/satra))
  - fix(mask): mark regions of non-task interest, and stop committing while unsure ([@satra](https://github.com/satra))
  - feat(fuse): split level 1 (signals) from level 2 (final uncertainty maps) ([@satra](https://github.com/satra))
  - refactor(posterior): weight speaker-count claims by measurement, not declared kind ([@satra](https://github.com/satra))
  - fix(support): admit only evidence that can withhold support ([@satra](https://github.com/satra))
  - refactor(identity): measure signal weight instead of declaring it ([@satra](https://github.com/satra))
  - style: import order and docstring formatting ([@satra](https://github.com/satra))
  - fix(identity): apply the derivation gate where the axis is aggregated ([@satra](https://github.com/satra))
  - style: docstring formatting in the reliability tests ([@satra](https://github.com/satra))
  - feat(identity): weight each signal's vote by its own measured reliability ([@satra](https://github.com/satra))
  - fix(identity): calibrate on the comparison the harvester actually makes ([@satra](https://github.com/satra))
  - fix(identity): report the empirical calibration band on every clusterer path ([@satra](https://github.com/satra))
  - docs: calibration script, LS tracks, and the three id namespaces ([@satra](https://github.com/satra))
  - fix(labelstudio): build regions before mutating, and read parquet list columns ([@satra](https://github.com/satra))
  - fix(types): give the test helpers explicit keyword parameters ([@satra](https://github.com/satra))
  - fix(types): unwrap optional measurements at their assertion sites ([@satra](https://github.com/satra))
  - test: assert the recorded cache schema version against the constant ([@satra](https://github.com/satra))
  - chore: drop scripts/profile_imports.py from this branch ([@satra](https://github.com/satra))
  - style: docstring formatting in the per-speaker convergence test ([@satra](https://github.com/satra))
  - fix(identity): pass the policy so the derived-source gate is actually in force ([@satra](https://github.com/satra))
  - feat(calibration): emit detection-margin profiles from measured verdicts ([@satra](https://github.com/satra))
  - feat(sources): separate presence from extent, and add modulation depth ([@satra](https://github.com/satra))
  - feat(labelstudio): surface the background mask and per-speaker presence for review ([@satra](https://github.com/satra))
  - fix(identity): attribute speakers to their actual backers and separate the id namespaces ([@satra](https://github.com/satra))
  - test(identity): guard the axes against the per-speaker change and pin reproducibility ([@satra](https://github.com/satra))
  - feat(identity): derive per-speaker structure from the per-bucket evidence ([@satra](https://github.com/satra))
  - test(scene): cover recovery delta, mask discounting, and cross-level ranking ([@satra](https://github.com/satra))
  - feat: excision routing; two fixes found by running the file as a cough task ([@satra](https://github.com/satra))
  - feat: wire per-speaker identity end to end; discard bracketed non-lexical tokens ([@satra](https://github.com/satra))
  - feat(plot): background mask on the main timeline, per-speaker presence on the adaptive one ([@satra](https://github.com/satra))
  - docs: background scene characterization and per-speaker identity (T107, T108) ([@satra](https://github.com/satra))
  - feat(adaptive): final per-speaker outputs; bump cache schema 3 -> 4 ([@satra](https://github.com/satra))
  - feat(audio_analysis): wire background sources into run_pass; fix band resolvability ([@satra](https://github.com/satra))
  - feat(identity): derive source reliability from perturbation evidence, not assignment ([@satra](https://github.com/satra))
  - refactor(identity): source kind is a policy declaration, not a hardcoded constant ([@satra](https://github.com/satra))
  - feat(audio_analysis): per-speaker identity uncertainty framework (T086-T097) ([@satra](https://github.com/satra))
  - feat(audio_analysis): foreground suppression, depth measurement, writers (T050-T052, T059, T064-T065, T069) ([@satra](https://github.com/satra))
  - feat(noise_floor): foreground-referenced SNR + cross-recording baseline ([@satra](https://github.com/satra))
  - feat(noise_floor): detect stationary sources present throughout the clip (T068, FR-021i) ([@satra](https://github.com/satra))
  - feat(audio_analysis): margin ladder + fabrication guards; mask restricted to unmodified pass ([@satra](https://github.com/satra))
  - feat(audio_analysis): per-band noise-floor estimator (T039-T045, T056-T058, T111) ([@satra](https://github.com/satra))
  - fix(level): peak-limited gain policy, found on a real recording ([@satra](https://github.com/satra))
  - feat(adaptive): mutual-influence guards, ahead of the paths they protect (T073-T084) ([@satra](https://github.com/satra))
  - feat(audio_analysis): wire background mask into run_pass + CLI (T033, T035, T037, T038) ([@satra](https://github.com/satra))
  - feat(audio_analysis): background mask with per-region uncertainty (T024-T032, T034, T034a, T036) ([@satra](https://github.com/satra))
  - feat(classification): lossless YAMNet input path + probe script + measured guard (T015-T016, T018, T020-T023) ([@satra](https://github.com/satra))
  - feat(classification): amplitude-invariance probe + AudioSet score fix (T012-T014, T017, T019) ([@satra](https://github.com/satra))
  - feat(audio_analysis): StageContext carries audio variant + gain (T009) ([@satra](https://github.com/satra))
  - feat(audio_analysis): detection-margin profile + level provenance (T005-T008, T010-T011) ([@satra](https://github.com/satra))
  - feat(deps): declare librosa + pyloudnorm; bump cache schema (T001-T004) ([@satra](https://github.com/satra))
  
  #### Authors: 3
  
</details>


satra:
### Code review

Found 3 issues:

1. unasdiff's cross-window permutation alignment correlates two non-corresponding audio regions on the final window. `_window_starts` appends a flush-to-end window whose overlap with its predecessor is larger than `hop_samples` (its own docstring says so: "that last window overlaps its predecessor by more than `hop_samples`"), but the alignment loop always slices a fixed `prev_tail = t[-overlap_samples:]` / `next_head = t[:overlap_samples]`. For a 4.92 s clip (`starts=[0, 14720]`), the two slices are offset by 17280 samples, so they are not the same audio — which breaks the invariant the comment three lines above asserts ("correlation on the overlap region ... so speech/sound content outside the shared region cannot influence the match"). The `first_real_measurements` margins recorded in `data/permutation_alignment.json` (0.0109, 0.0579) come from exactly this clip and sit three orders of magnitude below the synthetic known-correct margins, which is consistent with correlating misaligned regions.

https://github.com/sensein/senselab/blob/b650d2baf04155de36abe9789b2ed63d2467b1dd/src/senselab/audio/tasks/source_separation/unasdiff.py#L764-L776

2. `language` is silently dropped for Qwen3-TTS. It is bound as a named parameter (`language: Optional[Language] = None`, line 22), but the Qwen branch forwards it by testing `if key in kwargs` — where it can never appear. A caller passing `language=...` gets the worker default `"Auto"` with no error. The CoquiTTS branch on line 77 forwards the named parameter correctly, so this is specific to the new Qwen route; no test covers `synthesize_texts(..., language=..., model=Qwen/Qwen3-TTS...)`.

https://github.com/sensein/senselab/blob/b650d2baf04155de36abe9789b2ed63d2467b1dd/src/senselab/audio/tasks/text_to_speech/api.py#L60-L67

3. The PII module documents two detectors where three now run (CLAUDE.md says "Explain *why* in comments and docstrings, not *what*" — a docstring stating a false fact about current behaviour is the drift that rule exists to prevent). `subprocess_backend.py` defines `_KNOWN_DETECTORS = frozenset({DETECTOR_PRESIDIO, DETECTOR_GLINER, DETECTOR_RULES})` and resolves `detectors=None` to all three, but `api.py` line 14 says "Two detectors run in parallel inside the venv" and line 348 documents `detectors` as `Subset of {"presidio", "gliner"}` running "both" — so the valid value `"rules"` is undocumented. The same two-detector framing also appears at api.py:96-98, :156-157, :364-366 and :434.

https://github.com/sensein/senselab/blob/b650d2baf04155de36abe9789b2ed63d2467b1dd/src/senselab/text/tasks/pii_detection/api.py#L346-L350

🤖 Generated with [Claude Code](https://claude.ai/code)

<sub>- If this code review was useful, please react with 👍. Otherwise, react with 👎.</sub>

PR #553, opened 2026-08-14
fix(provenance): degrade an unresolvable commit to None instead of aborting the stage

wilke0818:
## Problem

`StageContext._commit_sha_for` computes an inference cache key **before** any cache
lookup, and it called `resolve_revision` without catching `RevisionResolutionError`.
Any Hub resolution failure there aborts the whole `analyze_audio` stage. Two real cases
hit this:

- **A non-Hub backend whose alias contains a `/`** — YAMNet's `"google/yamnet"`, loaded
  via a TensorFlow subprocess venv. The Hub 404s it, so a full `analyze_audio` run with
  scene classification enabled dies at `stage_scene` with `RepositoryNotFoundError`
  before producing any output. (An absolute local model path like `/scratch/models/foo`
  tripped the same `"/" in model_id` heuristic.)
- **A transient rate-limit / network blip** mid-run — a 429 during a batch burst took
  down the stage before a cache lookup could even run.

## Fix

Reuse `signal.resolved_commit_sha`, which already exists as the "one place that decides
how a resolution failure degrades": any `RevisionResolutionError` leaves the cache key
un-pinned (`None`) rather than crashing. This is safe because the cache key is not the
load — reproducibility strictness (refusing a mutable ref) lives in the loaders, so a
genuine typo still fails at load, not here. A gated repo we lack access to likewise
degrades here and fails loudly at load, where the missing licence actually matters.

## Verification

Full `analyze_audio` run (3 ASR models + alignment + pyannote diarization + AST + YAMNet
+ features), GPU, end-to-end `rc=0`:

- `Could not resolve google/yamnet@main … 404` → warning → degrades to `None` →
  `[yamnet] ok`, and on a later pass a cache HIT — no crash.
- Complete Label Studio bundle produced (`labelstudio_tasks.json`, `labelstudio_config.xml`,
  `disagreements.json`).

Unit tests: `stage_context_test.py` 27/27 (non-Hub id, Hub resolves, and transient / gated /
not-found all degrade to `None`). `ruff` + `mypy` clean.

## Note

This closes **finding #6** of a local review of #550 (both halves: the transient-stage-crash
and the `"/"`-in-id heuristic). The other findings from that review are being handled in
follow-up PRs.

### DIFF
--- a/src/senselab/audio/workflows/audio_analysis/stage_context.py
+++ b/src/senselab/audio/workflows/audio_analysis/stage_context.py
@@ -215,11 +215,18 @@ class StageContext:
 
         - **Not a Hub id at all** — a bare ``None`` (a model-less stage, e.g. ``features``) or a
           name with no ``/`` (a local backend). Short-circuits with no Hub round-trip.
-        - **A definitive not-found** (``RepositoryNotFoundError``) — the Hub has answered, and the
-          answer is that no commit exists. ``None`` is then the *correct* value rather than a
-          degradation, and it is stable across runs, so no two commits can collide behind it. This
-          is the crash being fixed: ``default.yaml`` ships ``yamnet: google/yamnet``, a TensorFlow
-          backend whose id happens to contain a ``/`` and so trips the Hub-id heuristic above.
+        - **A definitive "there is no commit"** (``RepositoryNotFoundError`` or
+          ``HFValidationError``) — ``None`` is then the *correct* value rather than a degradation,
+          and it is the same answer every run, so no two commits can collide behind it. Two shapes
+          reach it. ``RepositoryNotFoundError`` is the Hub having answered that no such repo
+          exists: ``default.yaml`` ships ``yamnet: google/yamnet``, a TensorFlow backend whose id
+          happens to contain a ``/`` and so trips the Hub-id heuristic above — the crash being
+          fixed. ``HFValidationError`` is the *client* refusing to ask, because the string is not a
+          well-formed repo id at all: a local filesystem path (``/scratch/models/foo``) contains
+          ``/`` and so trips the same heuristic, but ``model_info`` rejects it before any request.
+          That verdict is offline, deterministic and independent of Hub availability, which is what
+          makes it definitive rather than a "could not tell" — it cannot be a transient failure
+          wearing a not-found's clothes, because no network was involved in reaching it.
         - **Anything else** — a 429, a network error, a ``GatedRepoError`` — propagates. Those all
           mean "we could not tell", which is unsound for a key: the load may well succeed, and its
           result would be stored under a commit-blind key that a later run cannot distinguish from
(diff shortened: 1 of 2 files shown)


satra:
### Review

I merged #550, so I asked an independent agent to verify these claims against `alpha` rather than judge my own code. **The defect is real and was reproduced.** Two things need changing before this lands.

#### Confirmed

`ctx.cache_key_for("yamnet", "google/yamnet", ...)` raises `RevisionResolutionError` on `alpha` — reproduced directly. The root cause is sharper than the PR states: `stage_context.py:210` uses `"/" in model_id` as its "is this a Hub model?" test, and the docstring right above it says *"a local backend like `yamnet`, which has no `/`"* — while `default.yaml:145` ships `yamnet: google/yamnet`. The code's mental model is the short alias; the shipped config uses the long one. That divergence is the whole bug. The existing test `test_commit_sha_for_a_non_hub_id_skips_resolution` asserts on `"yamnet"`, which is why it never caught this.

The before-cache-lookup ordering is correct: `stages.py:244` evaluates `cache_key_for` as an *argument*, so it runs in the caller's frame and `run_task`'s catch-all can never see it. No upstream `try` exists.

#### 1. The `except` is broader than the defect, and narrowly reopens what #550 established

`commit_sha=None` is a value in the key payload (`cached_inference.py:393`), so two upstream commits of the same `model_id` collide on one key whenever resolution degrades. Concretely:

- **A local absolute path as a model id** fails resolution *deterministically, every run* — permanently commit-blind key, and the load succeeds. Editing that checkpoint in place then silently serves the stale result. On `alpha` this is a crash; this PR converts it into silent staleness.
- **The alignment path** (`stages.py:445-447`, raw string, no prior `HFModel` construction): a transient failure stores a result from commit X under a commit-blind key. Serving it stale needs a second transient failure in a later run spanning an upstream push — narrow, but it is exactly the window #550 existed to close.

The strongest argument against the current form is the docstring of the exception being swallowed (`model_revision.py:38-44`):

> *"a result whose provenance says 'unknown commit' is worth less than no result, **and it would still be cached under a key that cannot distinguish it from any other commit**."*

The PR's framing — reusing `signal.resolved_commit_sha` as "the one place that decides how a resolution failure degrades" — treats a provenance record and a cache key as the same decision. Degrading a *record* to "unknown" costs nothing. Degrading a *key* collapses every commit onto one bucket.

**Requested change:** narrow the `except` to a definitive not-found, excluding `GatedRepoError`. A `RepositoryNotFoundError` means there is no commit, so `None` is the *correct* answer rather than a degradation; a 429 or a gated repo means "we could not tell", which is unsound for a key. There is precedent for exactly this split at `dependencies.py:525-531`, and the cause is already on `exc.__cause__`. The current tests assert all three causes degrade identically, which throws away the distinction that matters.

#### 2. Import-weight invariant

`stage_context.py:12-16` states *"Deliberately light: no torch, no transformers … That keeps 'I want a cache key' from dragging in the ML stack"*, and `stage_context_test.py:279-294` guards it. Importing `signal` pulls `utils.data_structures.__init__` → torch + transformers. Measured: `after stage_context: torch=False` → `after signal: torch=True`. The guard test does **not** catch this because the import is deferred inside the function.

Keeping the direct `resolve_revision` import and wrapping it in a local `try/except` avoids this entirely — and given point 1, the "one shared place" framing is worth dropping anyway.

#### Smaller notes

- **The transient-429 case is misattributed.** For AST, ASR and diarization the abort already happens one line *earlier*, at `HFModel` construction, because Python evaluates the `model=` argument before `cache_key_for` (`stages.py:215` before `222`; `158` before `162`; `351` before `354`). This PR does not address that. Only the yamnet and alignment paths have `_commit_sha_for` as the first Hub touch. The Verification section evidences only the 404.
- **Worth listing under alternatives considered:** `_YAMNET_ALIASES` already accepts bare `"yamnet"` (`classification/api.py:21`), so changing `default.yaml:145` to `yamnet: yamnet` removes this crash with no cache-key semantics touched. It doesn't cover the general class, but it's the minimal fix.
- The gated-repo claim holds, but "fails loudly" is `run_task` recording `status: failed` and the run continuing — not fatal.
- `_DEFAULT_REVISION_REF`'s docstring still refers to "`_commit_sha_for`'s call into `resolve_revision`", no longer literally true after this change.

I'm pushing the narrowing and the import fix to this branch so it isn't blocked on a round trip — please review those commits rather than treating them as final.

PR #554, opened 2026-08-14
fix(hf-cache): keep resolution failures survivable and correctly typed

wilke0818:
## Summary

Three findings from a local review of #550, each a resolution/lock failure that
turns into the *wrong kind of crash* on the parallel-batch path #550/#539 target.
Sibling PR #553 handles finding #6; the offline/cold-cache trio (findings #1, #2, #4)
and lock hygiene (#8, #9) are separate follow-ups.

## Findings fixed

**#3 — `model_revision.py`: 429 is fatal at construction.**
`_resolve_uncached` makes a mandatory `HfApi.model_info` round-trip on the cold-cache
path, and it was the one Hub call without `retry_on_transient_error`. Under a SLURM
array hitting a cold model, a 429 raised `RevisionResolutionError` immediately — at
`HFModel` construction *and* at cache-key computation — undoing the transient
survivability #539 bought. Now wrapped in the same exponential-backoff retry every
other Hub call uses; a persistent or non-transient (4xx) error still surfaces as
`RevisionResolutionError`.

**#5 — `data_structures/model.py`: wrong exception type escapes the validator.**
`_resolve_commit_sha` runs `resolve_revision` in a `model_validator(mode="after")`
and let `RevisionResolutionError` (a `RuntimeError`) escape. Pydantic only converts
`ValueError`/`AssertionError` raised in a validator into `ValidationError`, so every
caller that catches `ValidationError`/`ValueError` around `HFModel(...)` (the pattern
the `revision` field validator establishes) saw an unhandled `RuntimeError`. Re-raised
as `ValueError` so it wraps consistently. Affects `HFModel`, `SpeechBrainModel`,
`PyannoteAudioModel`, `SentenceTransformersModel`.

**#7 — `file_lock.py`: bare subscript raises the wrong exception on an old holder.**
The Timeout branch computed `time.time() - previous_holder["taken_at"]` as a bare
subscript while every neighbouring field used `.get()`. `lock_holder` returns whatever
JSON is on disk, including a holder written by an older senselab that predates
`taken_at` (or a partial file that still parses) — so `__enter__` raised `KeyError`
(or `TypeError`) instead of the intended `TimeoutError`. The retry loops in
`ensure_hf_model` / `ensure_venv` / `record_resolution` only catch `TimeoutError`, so
it propagated as an unrelated crash. Now tolerates a missing/non-numeric value.

## Tests

- `model_revision_test.py`: a transient error is retried and resolution succeeds; a
  persistent/non-transient error still raises `RevisionResolutionError`.
- `model_test.py`: a resolution failure surfaces as `ValidationError`, not a bare
  `RuntimeError`.
- `file_lock_test.py`: a `taken_at`-less holder still times out cleanly (`TimeoutError`).

`pytest` (3 files) 43 passed / 1 skipped, `ruff` + `mypy` clean, pre-commit green.

### DIFF
--- a/src/senselab/utils/data_structures/model.py
+++ b/src/senselab/utils/data_structures/model.py
@@ -86,9 +86,30 @@ class HFModel(SenselabModel[PROVIDER_T]):
         path_or_uri = info.data["path_or_uri"]
         if not isinstance(path_or_uri, Path):
             if (str(path_or_uri), value) not in cls._hf_cache:
-                cls._hf_cache[(str(path_or_uri), value)] = check_hf_repo_exists(
-                    repo_id=str(path_or_uri), revision=value, repo_type="model"
-                )
+                try:
+                    cls._hf_cache[(str(path_or_uri), value)] = check_hf_repo_exists(
+                        repo_id=str(path_or_uri), revision=value, repo_type="model"
+                    )
+                except Exception as exc:
+                    # check_hf_repo_exists answers False only for "genuinely absent"; it
+                    # deliberately re-raises GatedRepoError and transient Hub errors rather than
+                    # letting either masquerade as "missing". Those are OSError / httpx.HTTPError
+                    # subclasses, and pydantic converts only ValueError and AssertionError raised
+                    # inside a validator into ValidationError -- so they escaped this validator
+                    # unwrapped, the same defect fixed in _resolve_commit_sha below. A caller
+                    # guarding HFModel(...) with `except ValidationError` saw an unrelated crash
+                    # from whichever of the two validators failed first.
+                    #
+                    # The cost, measured on pydantic 2.13.3: wrapping erases the failure's type.
+                    # The resulting ValidationError has __cause__ is None, and
+                    # errors()[0]["ctx"]["error"] is this ValueError, so a caller can no longer
+                    # tell "Hub outage" from "no such model" by exception type -- only by message,
+                    # which is why the original type's name is put in it. The original exception
+                    # itself survives one level deeper, at
+                    # errors()[0]["ctx"]["error"].__cause__, solely because of the `from exc`.
+                    raise ValueError(
+                        f"Could not verify the huggingface model {path_or_uri}@{value}: {type(exc).__name__}: {exc}"
+                    ) from exc
 
             if not cls._hf_cache[(str(path_or_uri), value)]:
                 raise ValueError(
@@ -123,7 +144,9 @@ class HFModel(SenselabModel[PROVIDER_T]):
         # ValueError/AssertionError raised inside a validator into ValidationError, so a RuntimeError
         # would escape unhandled — and every caller that catches ValidationError/ValueError around
         # HFModel(...) (the pattern the `revision` field validator above establishes) would see an
-        # unrelated crash instead. Re-raise as ValueError so it wraps consistently.
+        # unrelated crash instead. Re-raise as ValueError so it wraps consistently -- at the cost
+        # of the failure's type, which pydantic erases from the ValidationError; see the same
+        # trade-off spelled out in validate_hf_model_id above.
         try:
             self.commit_sha = resolve_revision(str(self.path_or_uri), self.revision)
         except RevisionResolutionError as exc:
--- a/src/senselab/utils/data_structures/model.py
+++ b/src/senselab/utils/data_structures/model.py
@@ -117,9 +117,17 @@ class HFModel(SenselabModel[PROVIDER_T]):
         """
         if isinstance(self.path_or_uri, Path) or self.commit_sha is not None:
             return self
-        from senselab.utils.model_revision import resolve_revision
+        from senselab.utils.model_revision import RevisionResolutionError, resolve_revision
 
-        self.commit_sha = resolve_revision(str(self.path_or_uri), self.revision)
+        # resolve_revision raises RevisionResolutionError, a RuntimeError. Pydantic only converts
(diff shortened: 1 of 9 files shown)


satra:
### Review

I merged #550, so an independent agent verified these against `alpha` rather than me judging my own code. **All three findings are real.** Several supporting claims are not, and one of them points at a bug bigger than this PR.

| Finding | Verdict | Bites in practice? |
|---|---|---|
| #3 — 429 unretried in `_resolve_uncached` | Real | **Yes** — cold-cache cluster runs |
| #5 — `RuntimeError` escapes a pydantic validator | Real, verified on pydantic 2.13.3 | No in-repo caller affected |
| #7 — bare `["taken_at"]` subscript | Real as code | No — hand-edited file only |

#### The strongest version of #3 isn't the one the PR makes

`RevisionResolutionError` is a `RuntimeError`, and `_is_transient` returns `False` for it — so wrapping the failure **defeats `load_hf_resilient`'s existing outer retry** at `dependencies.py:681`. That, plus `resolve_model` calling `resolve_revision` *before* `ensure_hf_model` (`dependencies.py:581`) so it is the first network call on a cold cache, is the real mechanism. Worth putting in the PR body — it's more convincing than what's there.

Three claims that don't hold:

- **"the one Hub call in the codebase without `retry_on_transient_error`"** — not accurate. Also unretried: `model.py:131`, `model.py:290-291`, `speech_emotion_recognition/api.py:124,370,379,383`, `speaker_diarization/child_adult.py:160`, `text_to_speech/qwen_tts.py:386`. Several swallow into `except Exception: return None` so they're benign, but the sentence as written is wrong.
- **"fatal at `HFModel` construction"** — essentially wrong. Pydantic runs field validators before `model_validator(mode="after")`, so `validate_hf_model_id` → `check_hf_repo_exists` → `snapshot_download` (already retried) runs first; under a 429 construction dies there with an `HfHubHTTPError`. The only construction-path route to line 196 is the narrow `_read_result_cache` fast path.
- **"a persistent or non-transient (4xx) error still surfaces"** — backwards, and this is the important one. `_is_transient` (`dependencies.py:80-88`) reads `.code` and `.status_code`, but HF errors carry the status on `.response.status_code`. `RepositoryNotFoundError` subclasses `httpx.HTTPError` → `OSError` and has neither attribute, so **`_is_transient(RepositoryNotFoundError(404)) == True`** — a genuine 404 is now retried 3× with backoff before failing.

That last one is a pre-existing bug in my code, not introduced here, and it affects the `snapshot_download` retry too. **Requested change:** fix `_is_transient` to read `exc.response.status_code` as part of this PR, since it is the actual bug behind the mis-scoping.

Consequently `test_resolution_still_fails_on_a_persistent_hub_error` uses `ValueError("not found")` as its "e.g. 404" stand-in — which is non-transient for reasons a real 404 does not share. **The test does not test what its docstring claims.** Please use a real `RepositoryNotFoundError`. The inline comment "exponential backoff on 429 / 5xx / connection errors" is likewise wrong today: it backs off on any HTTP error.

#### #5 — correct, and worth saying it's low impact

Verified empirically: a `RuntimeError` in `model_validator(mode="after")` escapes unwrapped; a `ValueError` becomes `ValidationError`. The four affected classes are right.

Two additions:

- **Incomplete.** The same defect sits two lines up: `validate_hf_model_id` (`model.py:83-89`) calls `check_hf_repo_exists`, which deliberately re-raises `GatedRepoError` and transient errors (`model.py:270-284`). Those are `OSError`/`httpx.HTTPError` subclasses and escape the *field* validator unwrapped too. By the finding's own stated principle, that path needs the same treatment.
- **Unmentioned cost.** Pydantic erases the chain — the resulting `ValidationError` has `__cause__ is None` and `ctx["error"]` is the new `ValueError`. The failure's *type* is destroyed; only the message survives, so a caller can no longer distinguish "Hub outage" from "no such model" by type. Worth a sentence in the comment.
- An AST sweep of every construction of those four classes under `src/senselab/` found three sites, all catching bare `Exception`, and zero occurrences of `ValidationError` anywhere. So this is an API-contract fix for external callers, not a live crash — fine, but the PR should say so.

#### #7 — real bug, invented justification

Confirmed: `file_lock.py:276` is a bare subscript while its neighbours use `.get()`, and all three named retry loops catch only `TimeoutError`, so a `KeyError` escapes an unbounded `while True`.

But the stated reachability does not exist. `git log -S'"taken_at"'` shows `taken_at` was in `_write_holder` in the *introducing* commit (`61840d7b`), and the implementation it replaced (`_HeartbeatLock`, removed in `51589e6f`) wrote **no content at all** — an empty file, which `lock_holder` maps to `None`. So there has never been a senselab that wrote a `taken_at`-less holder. "A partial file that still parses" is also unreachable: `taken_at` is emitted last, and truncated JSON has no closing `}`, so `json.loads` fails and `lock_holder` returns `None`.

That leaves a hand-edited file, or a foreign `<resource>.lock` — the latter only conceivable on the `FileRef` path. **Requested change:** rewrite the justification as "defensive against a hand-edited or foreign lock file" rather than claiming a version history that isn't there.

Also: there is a **fourth** `SharedFileLock` call site with no handler at all (`subprocess_venv.py:946-948`), which the PR doesn't mention — that slightly understates the blast radius.

Nit: `isinstance(True, int)` is `True`, so `"taken_at": true` renders as `held_for ≈ 1.7e9s`. Irrelevant at this severity.

I'm pushing the `_is_transient` fix, the test correction, and the `validate_hf_model_id` wrapping to this branch so it isn't blocked on a round trip — please review those commits rather than treating them as final.

PR #555, opened 2026-08-14
fix(hf-cache): don't chmod caller dirs; share one run manifest per Slurm job

wilke0818:
## Summary

Two findings from a local review of #550, both about senselab writing where it
shouldn't. Sibling PRs #553 (finding #6) and #554 (findings #3, #5, #7) cover the
resolution path; the offline/cold-cache trio (findings #1, #2, #4) is a separate
follow-up that needs a real offline validation run.

## Findings fixed

**#8 — locking a caller's file chmods their directory.**
Holding a `SharedFileLock` runs `_ensure_dir`, which chmods the lock file's parent
to `2775` (setgid + group-write). That is correct for senselab's own cache dirs,
but `subprocess_venv` also locks caller-supplied `FileRef` paths — an input file
that can live under a shared or read-restricted dataset tree. So senselab silently
granted setgid + group-write on a directory it does not own. Added a
`manage_dir_mode` flag (default `True`); the `FileRef` lock in `subprocess_venv`
now passes `False`, dropping a `.lock` beside the file without touching the
directory's mode. The lock/heartbeat *files* stay group-writable as before — only
the caller's *directory* is left alone.

**#9 — a UUID-per-process run id leaks manifest dirs and splits the run.**
`run_id()` minted a fresh `UUID4` per process when `SENSELAB_RUN_ID` was unset, and
`manifest_path` creates `runs/<id>/` per run with no GC. A Slurm array that forgets
to export `SENSELAB_RUN_ID` therefore left one directory (×3 files) per task in an
inode-quota'd cache root — and worse, split the run's resolution manifest, which
exists precisely to keep every task on one commit. `run_id()` now falls back to
`SLURM_ARRAY_JOB_ID` (shared by the whole array) then `SLURM_JOB_ID` before a UUID,
so a submission is self-consistent and reuses one manifest dir with no
configuration. An explicit `SENSELAB_RUN_ID` still wins.

*(The review also floated a TTL sweep as an alternative for #9; the Slurm-id
default is the lower-risk fix and also improves correctness by sharing the manifest
across an array. A GC sweep could still be layered on later if bare-launch UUID dirs
prove to accumulate.)*

## Tests

- `file_lock_test.py`: an unmanaged lock leaves a caller directory's mode (and
  setgid bit) untouched.
- `model_revision_test.py`: `run_id()` prefers the array job id, then the job id,
  with an explicit `SENSELAB_RUN_ID` still taking precedence; the shared fixture now
  drops Slurm vars so the run-id tests stay hermetic inside an allocation.

`pytest` (file_lock + model_revision + subprocess_venv) all green, `ruff` + `mypy`
clean, pre-commit green.

### DIFF
--- a/src/senselab/audio/workflows/audio_analysis/doc.md
+++ b/src/senselab/audio/workflows/audio_analysis/doc.md
@@ -160,13 +160,25 @@ split; it does not prevent it, and a split run is usually worthless rather than 
 So a run resolves each `(repo_id, ref)` **exactly once**, and every participant binds to that answer:
 
 ```bash
-# One submission, one run: every node, task and subprocess venv shares the id.
-export SENSELAB_RUN_ID="$SLURM_JOB_ID"
+# One submission, one run: leave the variable unset and senselab derives the id itself.
 uv run python scripts/analyze_audio.py audio.wav
 ```
 
-Leave it unset and senselab generates a UUID4 at first use and exports it to every subprocess it
-spawns, so a bare launch is its own self-consistent run with no configuration required.
+**Leave it unset.** Senselab derives a run id at first use — `SLURM_ARRAY_JOB_ID` if set, else
+`SLURM_JOB_ID`, else a UUID4 — and exports it to every subprocess it spawns. The array job id comes
+first because it is the one identifier every task of an array shares: a bare launch, a single job
+and an N-task array each end up as one self-consistent run with no configuration required.
+
+Set it explicitly only to force a grouping senselab cannot infer — several submissions that must
+count as one run, or one submission you want split into several. If you do set it for an array,
+set it to `$SLURM_ARRAY_JOB_ID`:
+
+```bash
+# Correct for an array. $SLURM_JOB_ID is *per-task*, so exporting that gives each task its own
+# run id and reintroduces exactly the split this section exists to prevent -- and an explicit
+# SENSELAB_RUN_ID outranks the fallback above, so it is not rescued.
+export SENSELAB_RUN_ID="$SLURM_ARRAY_JOB_ID"
+```
 
 The bindings live in `$SENSELAB_CACHE/runs/<run_id>/resolutions.json`, mapping
 `"<repo_id>@<ref>" -> "<40-hex commit>"`. Resolution consults it before the local `refs/` read and
--- a/src/senselab/utils/file_lock.py
+++ b/src/senselab/utils/file_lock.py
@@ -44,8 +44,23 @@ def _ensure_dir(path: Path, *, manage_mode: bool = True) -> None:
 
     ``manage_mode`` is True for senselab's own cache dirs, where the shared-tree modes are the
     whole point. It is False when the lock guards a caller-supplied path — e.g. a ``FileRef`` over
-    an input file living in a read-restricted dataset tree: senselab must not silently grant setgid
-    and group-write on someone else's directory just to drop a ``.lock`` file in it.
+    an input file sitting in the caller's *own* private directory.
+
+    The harm this guards against is self-inflicted, not inflicted on a stranger. ``chmod(2)``
+    returns ``EPERM`` unless the effective UID owns the directory, and the ``except OSError``
+    below swallows that — so a directory belonging to another user is never modified and never
+    raises. The chmod only lands on directories the invoking user owns, and there
+    ``LOCK_DIR_MODE`` is a widening: measured, a ``0o700`` directory comes back ``0o2775``, which
+    is not merely setgid + group-write but also **other-read and other-execute** — world traversal
+    of a directory its owner deliberately made private, as a side effect of dropping a ``.lock``
+    file in it.
+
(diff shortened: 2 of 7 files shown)


satra:
### Review

I merged #550, so an independent agent verified these against `alpha` rather than me judging my own code. **Both findings are real and both fixes are mechanically correct** — every behavioural claim about the fixes was confirmed by executing the patched `file_lock.py` directly. The diagnoses need correcting, and there's one gap that matters more than either finding.

#### #8 — real, but the framing is backwards in a way worth fixing

Reproduced on `alpha`: a `0700` directory containing the locked file comes back `0775` after one lock/unlock.

- **"a directory it does not own" is inverted.** POSIX `chmod(2)` returns `EPERM` unless the effective UID owns the file, and `_ensure_dir` swallows the `OSError` (`file_lock.py:52-53`). So senselab *cannot* modify a directory it doesn't own — nothing changes, nothing crashes. The chmod only succeeds on a directory the **invoking user owns**. The real harm is "we silently widen the user's own private directory", which is a different and more defensible bug report.
- **The escalation is understated.** `0o2775` from `0o700` also grants **other-read and other-execute** — world traversal of a directory the user deliberately set to `0700`. That's the more serious half and the PR doesn't mention it.
- **The motivating example is the wrong way round.** A `0500` directory: the *old* code chmods it to `0775` first and then writes, so it **defeats** a read-only directory rather than being blocked by it. Under the fix that case now raises `PermissionError`. That's a behaviour change worth calling out — I think failing loudly is right, but it should be stated, and the docstring's "read-restricted dataset tree" example should become "a caller's private directory".
- **Reachability:** `FileRef` is constructed nowhere outside `subprocess_venv.py` — no caller, no test, no script. `lock=True` appears nowhere in the repo. So "silently granted" (past tense) overstates it; nothing has ever taken this path. Still worth fixing as a public-API landmine.

**Requested change:** add a test asserting `subprocess_venv.py:946` passes `manage_dir_mode=False`. The diff tests the primitive only, and since the primitive's default stays `True`, a future edit could drop the argument with the whole suite still green.

Minor inconsistency worth a sentence: the fix argues senselab must not force group-write on a directory it doesn't own, then still forces `0664` on the `.lock`/`.heartbeat` files it drops in that same directory (`_touch_shared`, unchanged). Defensible — cross-user heartbeat needs it — but the boundary is drawn on an argument that applies to both.

#### #9 — the split-run half is real; the leak half is overstated and only partly fixed

**Split-run: real, but narrower than stated.** The chain is as claimed. But `_resolve_uncached` hits the local HF cache first via `_get_cached_commit_hash` → `refs/<revision>`, so on a cluster with one shared `HF_HUB_CACHE` every task reads the same pointer and gets the same SHA — no split. The split requires a genuinely cold pair: per-node scratch caches, or a repo not yet downloaded, straddling an upstream push. That is a real ORCD configuration, so the finding stands — but "splits the run" as an unconditional consequence of an unset variable is stronger than the code supports.

**Leak: overstated.** "×3 files" is one file in the normal case — `__exit__` unlinks the heartbeat and `filelock` unlinks the `.lock`; three files is the killed/preempted case only. And the fix shrinks rather than closes it: an array goes N→1, a non-array job is still 1 dir per job forever, and a bare launch is completely unchanged. The PR's parenthetical concedes this, which I credit, but the finding is titled "leaks manifest dirs" and that half isn't closed. Worth noting `audio_analysis/doc.md` already documents run dirs as "small and safe to delete once a run is finished" — manual cleanup is the stated design.

**Two new risks the PR should document:**

1. **Same-allocation merging.** Every senselab invocation inside one allocation now shares a manifest. Because entries are immutable for the run's life (`model_revision.py:150-151`), a second experiment in the same batch script — or unrelated work in a long-lived `salloc` — is silently pinned to SHAs the first resolved, with no way to re-resolve short of setting `SENSELAB_RUN_ID`. Previously each process got a fresh identity. That's a genuine behaviour change and belongs in the docstring beside the override.
2. **Job-id reuse.** Slurm ids wrap at `MaxJobId` and reset on a controller reinstall. With no GC, a recycled id lands on a months-old `runs/<jobid>/resolutions.json` and adopts its SHAs permanently — "confidently wrong provenance", which this module's own docstring names as the worst outcome. Low probability; cheap mitigation is folding `SLURM_CLUSTER_NAME` into the id, or ageing a manifest out by mtime.

#### The gap that matters most, and it's mine

`audio_analysis/doc.md:164` ships:

```bash
export SENSELAB_RUN_ID="$SLURM_JOB_ID"
```

For a **job array** that is per-task — so the documented recipe produces exactly the split this PR fixes, and because an explicit `SENSELAB_RUN_ID` outranks the new fallback, the fix does not rescue anyone following the docs. `doc.md:169` ("Leave it unset and senselab generates a UUID4 at first use") is stale too. Neither line is touched here. I wrote both in #550.

I'm pushing the doc fix, the call-site test, and the risk documentation to this branch so it isn't blocked on a round trip — please review those commits rather than treating them as final.

PR #560, opened 2026-08-17
Phase 2a: two carried defects fixed, plus a source distribution that ships the source

satra:
Toward Phase 2 of the triage graph. The design (`specs/20260816-143540-triage-graph/design.md`) lists defects that must be **fixed before lifting** rather than carried, because every one was found inside a refiner where rounds and multiple diarizers dilute a single bad vote — and a single-pass graph has no dilution. This PR does the two that are unambiguously safe, plus one repair that is my fault.

## First, the repair

**I merged #557 while its `pre-commit` check was failing.** It had been a draft, so I never watched its CI; when I marked it ready I merged immediately, having verified only #558. That put a red gate on `triage`.

The failures are 30 ruff and 3 mypy errors in the audit's reproduction scripts under `specs/`. Those scripts are **evidence** — several deliberately contain the defective pattern they demonstrate, so linting them to library standards asks them to stop being reproductions, and auto-formatting them edits the record of what was measured. Both tools already exclude `scripts/` for exactly this reason, so `specs/` joins it. The sixteen reproduction scripts the formatter had rewritten are restored to the bytes the audit produced.

## F-162 — the linking policy never reached the fold

`harvest_pass` called `fuse_consensus_words(asr_resolved)` with no `policy=`, so `linking.asr_slot_overlap` / `asr_slot_mid_tol_s` never reached the word-slot join on the only production path the graph lifts. The policy was already a parameter of `harvest_pass` and already passed to another call thirty lines earlier — the fix is one keyword.

The interesting part is that **the defect is inert today**: the packaged values (`0.3`, `0.15`) are identical to the fold's own fallbacks, so no output changes and no test could have caught it. It surfaces the first time someone edits the config and nothing happens — and the provenance then records the value that ran while the config records a different one.

A behavioural test cannot see this while the values coincide, and making them differ in a fixture would test the fixture rather than the wiring. So the guard is over the call site, the same shape as `revision_pinning_guard_test.py`.

**The guard found a second instance the register does not name**: `adaptive/interventions.py:595` calls `harvest_asr_votes` with no `fused=`, so an escalation round re-folds the words with the policy dropped, and `harvest_asr_votes` has no `policy` parameter to forward. Refiner-only — the graph never imports `adaptive/` — so it is allowlisted by name with that reasoning rather than fixed silently.

## F-150 — a run that harvested nothing reported a perfect rate

`high_uncertainty_rate` returned `0.0` when `total_rows` was 0, making a total harvest failure indistinguishable from a clean recording where nothing crossed the threshold — and `0.0` is the *better* reading, so the broken run presented as the best possible outcome. It is `None` now.

This matters beyond one summary field: the graph's review flag is meant to fire when evidence is too thin to adjudicate, and this is the field that told it the opposite.

## Deferred, with reasons

**F-144 — I do not think the design's "delete it" is right, and I measured why.** `speaker_identity.py`'s `multimodal_threshold=0.15` is unfitted and not scale-free, confirmed:

| agreeing sources | + 1 dissenter | `p(dissent)` | `is_multimodal` |
|---|---|---|---|
| 5 | 1 | 0.167 | True |
| 6 | 1 | 0.143 | **False** |

A diarizer that *agrees with the majority* flips the verdict, with no change to the audio and no change to the dissenting evidence. But `is_multimodal` is **consumed** — `speaker_identity.py:585` gates `converged` on it — so deleting it changes convergence behaviour. The arbitrary part is not the direction but the *location* of that flip: an unfitted constant decides how many corroborating sources it takes to silence a dissenter. That needs an evidence-carrying replacement, which is exactly what Phase 3 builds when speaker count becomes an `Estimate`. Deleting a consumed decision with no replacement is a bigger call than the design allowed for.

**F-165 — real, and the mechanism is more specific than the register states.** `speaker.py:527` clears the **entire** votes dict when a bucket has no words, including `target_activity`, which is mask-derived and does not depend on words at all. So non-lexical vocalization loses the one voter that could have registered it. The code comment defends this with adult inter-turn silence — the adult-speech assumption, in situ. Fixing it changes axis semantics and deserves its own PR.

## Verification

- `src/tests/audio/workflows/audio_analysis/` — 1420 passed, 3 skipped
- `pre-commit run --all-files` — 17 hooks green, verified directly, no files rewritten
- Both new tests confirmed to **fail against the pre-fix code** and pass after

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -12,7 +12,7 @@ repos:
   hooks:
   - id: mypy
     args: [--ignore-missing-imports, --extra-checks]
-    exclude: ^scripts/
+    exclude: ^(scripts|specs)/
     additional_dependencies:
     - pydantic>=1.10.4
     - types-requests
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,8 +4,16 @@ repos:
 - repo: https://github.com/astral-sh/ruff-pre-commit
   rev: v0.14.5
   hooks:
+  # Spec trees hold design notes plus the one-off scripts that reproduce a finding. Those scripts are
+  # evidence -- several deliberately contain the defective pattern they demonstrate -- so linting
+  # them to library standards asks them to stop being reproductions, and formatting them edits the
+  # record of what was measured. Excluded here rather than in pyproject's ruff config, because
+  # pyproject describes the package and `specs/` is not part of it (it is not in the wheel, and as
+  # of this change not in the sdist either). The mypy hook below excludes the same trees.
   - id: ruff
+    exclude: ^specs/
   - id: ruff-format
+    exclude: ^specs/
 
 - repo: https://github.com/pre-commit/mirrors-mypy
   rev: v1.18.2
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -133,6 +133,18 @@ profiling = [
 [tool.hatch.build.targets.wheel]
 packages = ["src/senselab"]
 
+# The wheel is scoped by `packages` above, but the sdist had no target config at all, so hatchling's
+# default swept in everything git tracks: `.claude/` (511 entries, including a whole abandoned
+# worktree), `.superpowers/`, `.specify/`, `.uv-cache/`, and 223 files of `specs/` design notes.
+# An 81 MB source distribution of a package whose importable code is a few MB, and every one of
+# those trees is working state for this repository rather than anything a consumer builds from.
+#
+# `only-include` rather than `exclude`: a new agent-tooling directory should be absent by default
+# and added deliberately, which is the opposite of what an exclude list does. `src` carries both
+# `senselab` and its tests, so a downstream rebuild can still run the suite.
+[tool.hatch.build.targets.sdist]
+only-include = ["src", "pyproject.toml", "README.md", "LICENSE.md", "SECURITY.md", "uv.lock"]
+
 [tool.hatch.version]
 source = "vcs"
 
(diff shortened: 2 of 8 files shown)


satra:
### Correction: the lint exclude was the wrong fix

Adding `specs` to pyproject's ruff `exclude` was answering the right complaint the wrong way. `pyproject.toml` describes the **package**, and `specs/` is not part of it — so the lint exclusion belongs beside the mypy hook's, in `.pre-commit-config.yaml`. It now lives there for `ruff` and `ruff-format` too, and the pyproject change is reverted.

Chasing that down found a real defect that had nothing to do with linting.

**The sdist had no target config at all.** The wheel is scoped by `packages = ["src/senselab"]`, but hatchling's sdist default sweeps in everything git tracks:

| tree | entries in sdist |
|---|---|
| `.claude/` (incl. an entire abandoned worktree) | 511 |
| `src/` | 510 |
| `specs/` | 223 |
| `.superpowers/` | 123 |
| `.specify/`, `.uv-cache/`, `tutorials/`, … | 60+ |

An **81 MB source distribution** for a package whose wheel is 1.2 MB, made almost entirely of working state for this repository rather than anything a consumer builds from.

Fixed with `only-include` rather than `exclude`, deliberately: the failure mode here was a default that said yes to everything, and an exclude list reproduces it — the next agent-tooling directory would ship until someone noticed. An allowlist makes a new top-level directory absent by default and present only on purpose.

**Measured:** sdist 81 MB → 18 MB, top-level entries 20 → 8 (`src` plus the metadata files the build references). Wheel unchanged at 1.2 MB with identical contents. The remaining 18 MB is `src/tests/data_for_testing/`, kept so a downstream rebuild can run the suite.

PR #567, opened 2026-08-19
The triage workflow designed from goals: 19 decisions, the measurements behind them, and the first verified labels

satra:
Spec-only. No code, no behaviour change. This is the design record for the triage workflow, built from goals rather than recovered from the current call graph, plus the measurements it rests on and the first human-verified labels in this project.

## What is here

| file | contents |
|---|---|
| `decisions.md` | D1–D19, in the order taken, including what each supersedes |
| `ground-truth-2026-08-18.md` | six human-verified events, and every instrument scored against them |
| `measurements-2026-08-17-span-probe.md` | what real recordings resolve, across ~20 models |
| `branch-1-airway.md` | the airway branch's DAG, its open items, and what it will not report |
| `model-to-branch.md` | tool capability table (not a routing decision — see D19) |
| `flowchart.md` | 11 mermaid diagrams |
| `ports.md`, `workflow.nf` | port tables and Nextflow-shaped pseudocode |
| `design.md` | the argument |
| `test-examples.md` | labelled material that would exercise the flow |

## The decisions that shaped it

**ADMIT discards only a file with no signal or a flat signal** (D1). Everything else — including whether there is speech — is a TAXONOMY measurement carrying uncertainty, not a gate returning a boolean. A speech gate at the front would discard exactly the recordings this workflow exists to catch: an off-target speaker is quiet by construction, and a non-lexical vocalization carries no words at all.

**TAXONOMY is file-level and is the real gate** (D4), with three verdicts rather than two. Discard requires *confident absence*, which is not a low presence score; uncertain always flags. Thresholds stay liberal and underived because no labelled verdicts exist to fit them, and span detection adjudicates downstream.

**Branching is by physiological mechanism, not by task** (D10). The b2ai registry's 45 tasks elicit eight kinds of vocal production, which collapse to four measurement stacks — and a task is a *loading vector* over them, not a member of one. Maximum phonation time loads respiratory and phonatory together, which is what makes it diagnostic. Only one stack needs ASR, which retires the word-gate problem structurally rather than patching it.

**Enhancement is an operation a node may invoke, not a variant or pathway** (D19, superseding D2). Nothing carries a `variant` label and no product exists in two versions, so there is no route to prohibit — only tool properties: an enhancer that takes breath to −40 dB is unusable by any node that needs breath.

**Measure per source, assign the target later or not at all** (D14). Target selection without an anchor is a prior that is most wrong exactly where the branch is not speech, and making it a prerequisite means a wrong prior silently discards the participant's measurements.

## The measurements that overturned things

**Only DSP resolves an onset.** Cough onset bounded to ±5 ms on a 1 ms envelope; YAMNet's response to the same cough is 480 ms wide, AST's 960 ms, and shrinking AST's hop tenfold made its leading edge *worse* — 1.06 s early at 0.48 hop, 1.58 s early at 0.05 — because response width is set by the window, not the hop.

**HeAR is a gate, not a locator.** Its detector window is hard-fixed at 32000 samples (0.5/1.0/1.5/3.0/4.0 s all raise `InvalidArgumentError`), so an event of duration D elevates every window centre within D + 2 s, merging any two events closer than 2 s. Sliding at 20 ms steps, 40 ms of cough — 2% of the window — crosses p > 0.5. So a firing window carries one bit, and inter-cough intervals of 0.3–0.5 s within a bout can never be counted.

**YAMNet resolves what HeAR merges**, at 0.96 s, with an explicit `Silence` class and `Gasp`/`Sigh`/`Throat clearing` in its vocabulary. Which reverses the branch's proposer and labeller: YAMNet proposes, HeAR confirms, DSP bounds.

**Coverage is a property of the element, not the instrument** — 87–98% for speech, 64–98% for cough, **10–52% for breath** across every detector. The breaths are *longer* than the coughs and covered far worse, so it is the gradual energy profile of turbulent flow, not duration. Breath duration is therefore not reportable, and that is structural rather than provisional.

**The configured enhancer is net-harmful above ~5 dB SNR.** `sepformer-wham16k-enhancement`'s output SI-SDR is pinned near 4.8 dB however clean the input, so it degrades any recording already cleaner than that — which is every quiet close-miked capture in the workflow's real input class. The ceiling is checkpoint-specific, not architectural: `dns4` reaches 13.3 dB with the same architecture and harness.

## Three claims of my own that measurement refuted

**unasdiff cannot name a source.** In `speech_sound` a matched control (`Computer_keyboard`) routed 96.7% of cough energy against `Cough`'s 97.1% — indistinguishable, so the label is inert and the separation comes from the speech/sound prior asymmetry. In `sound_sound` the apparent 53 dB label effect proved to be **slot ordering**: swapping `[Electric_piano, Cough]` for `[Cough, Electric_piano]` left both coughs in slot 0 regardless.

**DriftSE's plain checkpoint is not a hallucinating model.** Its training objective is exactly scale-invariant, so its absolute gain is undefined by construction; our worker dropped upstream's rescale and a PCM_16 write then clipped 98.5% of samples into a square wave. Restoring one line takes correlation 0.204 → 0.944.

**A CrisperWhisper discrepancy was an upstream retrain, reached through a stale ref.** Not preprocessing — four input routes give identical output. `831f87e1` (2026-08-03) emits five non-speech tokens; `de0369c8` (2026-08-17 retrain) emits one. A locally cached `refs/main` served the older weights a day after the push, which is the hazard CLAUDE.md records, occurring in our own measurements.

## What is deliberately unresolved

`group_events` — grouping a burst, a voiced phase and an aspirate tail into one cough — has no cue detector behind it, and one class of solution is now ruled out rather than merely untried: events closer than 2 s are merged inside HeAR's posterior before any threshold sees them.

Breath extent is bracketed between an over-covering posterior and an under-covering event list, with nothing that bounds it. `cfg.transient_isolation_s` has no value this recording could supply. Three branches are undesigned. The DAG core is not built.

**And every number rests on one 14-second recording**, one speaker, close-miked, quiet, with background music. That is the binding constraint on all of it.


satra:
**Not merging until this covers every branch and the downstream outputs.** Marked draft so it cannot land early.

What is here today is one branch of several: `branch-1-airway.md` drafts the airway path (inhalation, exhalation, cough, throat clear, mouth non-speech sound). The speech branch is referenced in `model-to-branch.md` — its enhancement question is explicitly left open — but is not designed. `design.md` frames Q0–Q10 and the DAG, and §7 maps to today's code, but the per-branch node graphs and the downstream output contracts are not written.

The branch-1 draft should also be treated as unsettled rather than as a baseline to extend. Two things about the verification recording constrain what any measurement on it can say, and they are properties of the data, not of a design:

- The longest verified-empty stretch is **1.80 s**, against HeAR's hard-fixed 2 s window — so on that file HeAR has no negative available and cannot be scored as a confirmer.
- The verified-empty audio totals ~8.8 s and supports **5 independent window decisions**; one false positive is worth 7.13 FP/min. The event-level false-positive count is structurally zero at every threshold, because merged 0.96 s detections are wider than every gap — which reads as perfect precision.

A recording with gaps longer than the widest model window is a precondition for evaluating any propose/confirm split, whatever the branches end up being.

PR #576, opened 2026-08-19
measure: branch 1's propose-confirm-bound chain against the only verified labels

satra:
Runs the measurement `specs/20260817-triage-workflow-dag/branch-1-airway.md` names as the one that
would settle its proposer choice: recall over the six verified events and false positives per minute
over the verified-empty stretches, on the original and on degraded copies.

Standalone. No `VoteStore`, no pass summaries, no `rounds.py`, no L1/L2 artifacts — the three stages
are called directly and everything is scored in `specs/20260819-branch1-ground-truth-eval/`.

## What it settles

| claim | verdict | number |
| --- | --- | --- |
| YAMNet as proposer recovers the events | partly refuted | 5/6 at every threshold to 0.70; event 1 is never seen — YAMNet asserts `Silence` at 0.72–1.00 over it |
| the proposer's FP/min is defensible | **not answerable on this file** | 8.4 s of verified-empty audio supports 5 independent window decisions; one FP is worth 7.13 FP/min; exact 95% CI on the clean rate is [0.005, 0.716] |
| the proposer is robust | refuted for the `Silence` route | 1−P(`Silence`) fires in 29 of 29 windows at 20 dB SNR white noise (FP 5/5, duty 100%) |
| | partly upheld for the vocabulary route | FP stays 0/5, recall falls 5/6 → 3/6 by 0 dB |
| HeAR as confirmer adds precision | **refuted** | 0 of 50 HeAR windows on this file overlap no event, so it has no negative available; it corroborates 4/7 planted verified-empty proposals above 0.89 |
| HeAR as confirmer costs little recall | refuted | at θ=0.5 it rejects the verified speech event (0.339) and the verified mouth sound (0.036), taking the chain 5/6 → 4/6 for zero precision gain |
| DSP envelope bounds a cough onset to ±5 ms | **upheld, n=1 → n=2** | worst onset error over both coughs and Δ ∈ [3, 20] dB is 5 ms; 0 ms at Δ=6 dB |
| the same bounder gives a usable offset | refuted | cough offsets run −449 to +20 ms and move 180–200 ms across the same Δ sweep |

**The chain end to end: proposed 5/6, confirmed 4/6, correct label 4/6, onsets ±3 ms on the two
coughs, no offset bounded.** The only number that survives comes from the stage that uses neither
model.

## Three findings the design should act on

1. **The event-level FP/min metric is structurally zero and reads as perfect precision.** At 0.96 s /
   0.48 s the merged detections are wider than every gap in this recording, so no detection is
   event-free. The duty cycle is the measure that shows what happens: 50–66% of verified-empty time
   is covered at usable thresholds.
2. **HeAR is disqualified as confirmer by the same geometry that disqualifies it as proposer.** Every
   gap in this file is under 2 s, so no placement of its 32000-sample window avoids a verified event.
   The draft drew half that conclusion.
3. **A merged proposal destroys the bound even when the bounder is good** — event 3 picks up a
   +2619 ms onset error at the one operating point that proposes all six events, because the seed
   contains two events. `span_refine` reading "only inside a proposal" is a dependency on the
   proposer having separated the events, not a safety property.

## Discipline

No timing or memory figures: other work was running on this host, and a contended cost number is
worse than none. Every probe states what result would count as the effect being present before the
result. Degradation parameters are stated rather than defaulted (white and pink noise at stated SNRs
on full-file RMS, synthetic exponential-decay RIR at T60 0.3/0.7 s, DRR 0 dB); no variant clipped.
The `13.79–14.04` unlabelled stretch is excluded from scoring in both directions.

Full numbers, method and limits in
[`specs/20260819-branch1-ground-truth-eval/measurement.md`](specs/20260819-branch1-ground-truth-eval/measurement.md);
raw output in `raw/report.txt`. The degraded WAVs are not committed (24 MB, exactly regenerable from
the recorded seed and manifest).


satra:
Closing — not asked for. I launched this on my own reading of a design-doc line, and scoring branch 1 is triage design, which is the user's call rather than mine.

Docs and scripts only, no source changes, so nothing is lost by dropping it.

